PR #99, opened 2017-06-23
virtual_machine - template disk resize

alekssaul:
disclaimer: first contribution to Terraform and a golang project.

Using vCenter, one can resize the boot disk with the clone wizard, however Terraform provider restricts resizing the disk if it's generated from a template. 

Change Allows resizing of Virtual Machine Disk from template
- removes the logic to check for "size" if disk is a template
- creates a new buildVirtualDeviceConfigSpecResize function to create a VirtualDeviceConfigSpec to be passed on during VM Clone event

```
resource "vsphere_virtual_machine" "test" {
  name   = "terraform-test"
  vcpu   = 2
  memory = 4096

  network_interface {
    label = "private"
  }

  disk {
    template = "/nyc/vm/templates/coreos_1409.2.0"
    datastore = "dscluster/datastore2"
    size = "50"
    type = "thin"
  }

  disk {
    name = "disk2"
    size = "20"
    type = "thin"
  }
}
```
As an example above will resize the template disk to 50 GB during cloning, and attach a second disk during customization process.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -832,16 +832,12 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 				}
 
 				if v, ok := disk["size"].(int); ok && v != 0 {
-					if v, ok := disk["template"].(string); ok && v != "" {
-						return fmt.Errorf("Cannot specify size of a template")
-					}
 
 					if v, ok := disk["name"].(string); ok && v != "" {
 						newDisk.name = v
-					} else {
+					} else if v, ok := disk["template"].(string); !ok && v != "" {
 						return fmt.Errorf("[ERROR] Disk name must be provided when creating a new disk")
 					}
-
 					newDisk.size = int64(v)
 				}
 
@@ -1664,6 +1660,54 @@ func buildStoragePlacementSpecClone(c *govmomi.Client, f *object.DatacenterFolde
 	return sps
 }
 
+// buildVirtualDeviceConfigSpecResize builds VirtualDeviceConfigSpec for Resize action.
+func buildVirtualDeviceConfigSpecResize(c *govmomi.Client, vm *object.VirtualMachine, diskSize int64) types.VirtualDeviceConfigSpec {
+	vmr := vm.Reference()
+
+	var o mo.VirtualMachine
+	err := vm.Properties(context.TODO(), vmr, []string{"datastore"}, &o)
+	if err != nil {
+		return types.VirtualDeviceConfigSpec{}
+	}
+	ds := object.NewDatastore(c.Client, o.Datastore[0])
+	log.Printf("[DEBUG] findDatastore: datastore: %#v\n", ds)
+
+	devices, err := vm.Device(context.TODO())
+	if err != nil {
+		return types.VirtualDeviceConfigSpec{}
+	}
+
+	var disks []*types.VirtualDisk
+	for _, device := range devices {
+		switch md := device.(type) {
+		case *types.VirtualDisk:
+			disks = append(disks, md)
+		default:
+			continue
+		}
+	}
+
+	var editdisk *types.VirtualDisk
+	if len(disks) == 1 {
+		editdisk = disks[0]
+	} else {
+		return types.VirtualDeviceConfigSpec{}
+	}
+
+	if int64(diskSize) != 0 {
+		editdisk.CapacityInKB = int64(diskSize) * 1024 * 1024
+	}
+
+	vdcs := types.VirtualDeviceConfigSpec{
+		Device:    editdisk,
+		Operation: types.VirtualDeviceConfigSpecOperationEdit,
+	}
+
+	vdcs.FileOperation = ""
+
+	return vdcs
+}
+
 // findDatastore finds Datastore object.
 func findDatastore(c *govmomi.Client, sps types.StoragePlacementSpec) (*object.Datastore, error) {
 	var datastore *object.Datastore
@@ -1779,6 +1823,9 @@ func (vm *virtualMachine) setupVirtualMachine(c *govmomi.Client) error {
 	}
 	if vm.template == "" {
 		configSpec.GuestId = "otherLinux64Guest"
+	} else {
+		vdcs := buildVirtualDeviceConfigSpecResize(c, template, vm.hardDisks[0].size)
+		configSpec.DeviceChange = append(configSpec.DeviceChange, &vdcs)
 	}
 	log.Printf("[DEBUG] virtual machine config spec: %v", configSpec)
 
--- a/vsphere/resource_vsphere_virtual_machine_test.go
+++ b/vsphere/resource_vsphere_virtual_machine_test.go
@@ -1471,6 +1471,72 @@ func TestAccVSphereVirtualMachine_DetachUnknownDisks(t *testing.T) {
 	})
 }
 
+const testAccCheckVSphereVirtualMachineConfig_templateResize = `
+resource "vsphere_virtual_machine" "template_resize" {
+%s
+    name = "terraform-test"
+    vcpu = 2
+    memory = 1024
+    network_interface {
+        label = "%s"
+    }
+     disk {
+%s
+        size = 10
+		template = "%s"
+	}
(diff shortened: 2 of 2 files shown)


s-urbaniak:
Nice!!! 👍

verdel:
This is really a necessary thing

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -832,16 +832,12 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
...
 				if v, ok := disk["size"].(int); ok && v != 0 {
-					if v, ok := disk["template"].(string); ok && v != "" {
-						return fmt.Errorf("Cannot specify size of a template")
-					}
 
 					if v, ok := disk["name"].(string); ok && v != "" {
 						newDisk.name = v
-					} else {
+					} else if v, ok := disk["template"].(string); !ok && v != "" {
 						return fmt.Errorf("[ERROR] Disk name must be provided when creating a new disk")
```
grubernaut: Is this error message still correct?

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1779,6 +1823,9 @@ func (vm *virtualMachine) setupVirtualMachine(c *govmomi.Client) error {
 	}
 	if vm.template == "" {
 		configSpec.GuestId = "otherLinux64Guest"
+	} else {
+		vdcs := buildVirtualDeviceConfigSpecResize(c, template, vm.hardDisks[0].size)
```
grubernaut: can only have a single disk in this case?

alekssaul:
Hi @grubernaut ,

Thank you for the review. I'm away from the office for the week. Hoping to work on your comments next week.

vancluever:
Hey @alekssaul - sorry this one has not gotten any response on this end - but I'm pinging you now to see if you are still working on this :) let us know!

alekssaul:
@vancluever : Yes, sorry.. work has been a bit busy :(

vancluever:
@alekssaul - fair enough!

@grubernaut seemed to think that this looked good save a few things - do you think you will have time to work on this in the next couple of weeks? If not, we can finish the work for you and keep your commits :)

alekssaul:
Yes I do. 

Was hoping to write test for it as well if I can squeeze it in.

vancluever:
Awesome! Will keep an eye out then :+1:

vsphere/resource_vsphere_virtual_machine.go
alekssaul: Prior to change, we check to make sure `name` for the disk is provided and fail if not. With the change, we check for the `name` but only fail where disk is not created from a template. Prior to the change every condition where "disk/size" exist required "disk/name" to be present. Now "disk/size" can be used without "disk/name" if disk is a template.

I can change the error string to something like `Disk name must be provided when creating a new disk that is not a template`.

vsphere/resource_vsphere_virtual_machine.go
alekssaul: you can have multiple disks but only one template disk type ex:
```
 disk {
    template = "/nyc/vm/templates/coreos_1506.0.0"
    datastore = "dscluster/datastore2"
    size = "10"
    type = "thin"
  }

  disk {
    name = "test1"
    size = "20"
    type = "thin"
  }
```
Works. If you were to attempt multiple disks such as 
```
  disk {
    template = "/nyc/vm/templates/coreos_1506.0.0"
    datastore = "dscluster/datastore2"
    size = "10"
    type = "thin"
  }

  disk {
    template = "/nyc/vm/templates/coreos_1437.0.0"
    datastore = "dscluster/datastore2"
    size = "10"
    type = "thin"
  }
```
error generated with `[ERROR] Only one bootable disk or template may be given` which is also a limitation on master; https://github.com/terraform-providers/terraform-provider-vsphere/blob/730e48bd327190815b8cd04d0cacbff1d7e70a66/vsphere/resource_vsphere_virtual_machine.go#L855 . This is because template disk is expected to be the boot disk, and can only have one boot disk.

alekssaul:
Hi @vancluever @grubernaut ,

Thank you for your patience and apologies about the delay.

Added a commit with test, which look good based on my setup; https://gist.github.com/alekssaul/bde8dba4df4b1d8564066690dbc8b4af . Also added comments to reviews which I hope makes sense. 

Please let me know if any questions or concerns.

vsphere/resource_vsphere_virtual_machine.go
vancluever: Hey @alekssaul, I'm kind of wondering if maybe @grubernaut was asking if maybe the error _condition_ was not correct here.

It's checking if the the value of a `map[string]interface{}` is a valid string, I'm assuming, but the conditional fires if:

* The key is not present (`!ok`)
* **and if only also if** `v` is not empty (`&& v != ""`)

Since the zero value of a map[string]interface is `nil`, the value of `v` when `!ok` is true will always be `nil`.

To have the condition match the message, I think you would want:

```
if v, ok := disk["template"].(string); !ok || v == ""
```

This condition would be true if the key was not present OR if `v` was an empty string.

vancluever:
Hey @alekssaul - sorry for the delay here. I tried to run the tests but got this:

```
		* vsphere_virtual_machine.template_resize: resource pool 'hashi-resource-default' not found
```

I looked at the commit log and it seems a few months old - there have been some major bugfixes to the project that fix some of these issues. I rebased off of master and tried again and all worked - can you rebase your code on top of master and re-push?

vancluever:
Hey @alekssaul - looks like the test failed after all, even after re-basing:

```
--- FAIL: TestAccVSphereVirtualMachine_templateResize (363.30s)
	testing.go:449: Step 0 error: Error applying: 1 error(s) occurred:
		
		* vsphere_virtual_machine.template_resize: 1 error(s) occurred:
		
		* vsphere_virtual_machine.template_resize: vm 'terraform-test' not found

```

Can you re-base on your end and let me know?

ze42:
Trying to use that PR... rebased on currest master (branch can be found here: https://github.com/ze42/terraform-provider-vsphere/tree/vm-resize-disk )

  * Works on first attempt, with a simple template, with or without actual resize
  * Fails to create a VM from a template with multiple disks, even without resize (works fine without those 2 commits)

```
Error applying plan:

1 error(s) occurred:

* vsphere_virtual_machine.innovation-vpn-1: 1 error(s) occurred:

* vsphere_virtual_machine.innovation-vpn-1: ServerFaultCode: 
Required property device is missing from data object of type VirtualDeviceConfigSpec

while parsing serialized DataObject of type vim.vm.device.VirtualDeviceSpec
at line 2, column 989

while parsing property "deviceChange" of static type ArrayOfVirtualDeviceConfigSpec

while parsing serialized DataObject of type vim.vm.ConfigSpec
at line 2, column 812

while parsing property "config" of static type VirtualMachineConfigSpec

while parsing serialized DataObject of type vim.vm.CloneSpec
at line 2, column 216

while parsing call information for method CloneVM_Task
at line 2, column 66

while parsing SOAP body
at line 2, column 60

while parsing SOAP envelope
at line 2, column 0

while parsing HTTP request for method clone
on object of type vim.VirtualMachine
at line 1, column 0

```

vancluever:
Hey all,

With #244, we have completely de-coupled cloning from template from the actual `disk` sub-resources. There are now rules to how `disk` sub-resources are structured when cloning from template:

* Resizing is supported as long as you are not using `linked_clone`. The new size has to be greater than the template's disk size - shrinking disks is not supported.
* Resizing is not supported on `linked_clone`. The disk size need to be exactly the same size as the template.
* You need to have at least the same amount of disks as exists in the template. The resource does support adding extra disks on top of these. Disks are sorted by unit number and then distributed to each disk sub-resource in the same `unit_number` order specified in the `disk` sub-resources.

There has been a new `vsphere_virtual_machine` data source created to assist with locating disk sizes as well, which can assist with getting disk sizes when working with linked clones so you don't have to manually remember the size of each disk.

As this addresses most, if not all of the concerns here, I am going to close this PR. If you are curious about the new implementation ahead of time, I recommend checking the documentation linked in the PR.

Thanks @alekssaul for all of your work on this!

PR #106, opened 2017-07-20
modified split to use '] '

angelichorsey:
Necessary for datastores that have spaces in their names.


angelichorsey:
Realized there was more Split() fixes for datastores with spaces in their name.

PR #107, opened 2017-07-21
Enhancement: Added resource for snapshot and revert snapshot

idmsubs:
Please review code changes implemented for vSphere snapshot management resources.

Resources implemented are :
a. Snapshot management (Create and delete snapshot)
b. Revert snapshot

Referred https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md for following standards for Terraform provider.

Please suggest changes if required any.


girishramnani:
@idmsubs revert snapshot seems more like an action then a resource so instead of having a resource if there is a `revert_to_this` or similar named key on the snapshot resource. This key if is kept true then the state of the system is on that snapshot, Making it more declarative then procedural.

Any suggestions?.

idmsubs:
@girishramnani that can be possible, but to avoid additional checks for mandatory and non-mandatory input parameters of create_snapshot and revert_snapshot I have created separate resource for revert_snapshot. Like for create we require "description", "memory", "quiesce", "remove_children", "consolidate", and for revert requires "suppress_power_on" - these are irrelevant in both and with suggested way use will have provide all values in any of case (revert flag = true/false). 

Let me know your thoughts on this.

girishramnani:
Yeah totally get that, but I am bit concerned about a state where someone removes the snapshot which is referred by `snapshot_name` of revert_snapshot.

There is also one more option. What if the `virtual machine` resource has an optional field called `snapshot_name` which can be interpolated as `${vsphere_snapshot.snap.id}`

this can also help in cleanup process i.e. remove the snapshot if the vm is destroyed ( if the user wants that ).

NOTE - here `snap` is a `snapshot` resource that was created.

idmsubs:
> about a state where someone removes the snapshot which is referred by snapshot_name of revert_snapshot.
if someone removes snapshot manually and tries to remove it again with resource, apply will fail as per my implementation.

> one more option.
this way is possible, but consider actual use with this option. User will create virtual machine (as its new, no snapshot name will be provided). Later whenever user wants to create snapshot, user will have to do apply for create virtual machine, with providing mandatory values of virtual machine just for snapshot create.  In that we can add this option for create virtual disk also :)

> remove the snapshot if the vm is destroyed
this user still can do with multiple resources in microservice, right? User can delete snapshot and then delete VM.

girishramnani:
`Later whenever user wants to create snapshot, user will have to do apply for create virtual machine` - Correct me if I am wrong but I think this is how life cycle would look like - 

If the user wants to create a snapshot then they would add a `vsphere_snapshot` resource which would look like this.

```

resource "vsphere_virtual_machine" "testing" {
     ...
}
resource "vsphere_snapshot" "Test_terraform_cases"{
       vm_name = "snapshot"
 	snapshot_name = "SnapshotForTestingTerraform"
	description = "This is snpashot created for testing and will be deleted."
 	memory = "true"
 	quiesce = "true"
 	remove_children = "false"
 	consolidate = "true"
 }

```

The moment user want to revert to the specific snapshot they will provide the snapshot name to the vm

```

resource "vsphere_virtual_machine" "testing" {
     ...
     snapshot_name ="${vsphere_snapshot.Test_terraform_cases.id}"
}
resource "vsphere_snapshot" "Test_terraform_cases"{
       vm_name = "testing"
 	snapshot_name = "SnapshotForTestingTerraform"
	description = "This is snpashot created for testing and will be deleted."
 	memory = "true"
 	quiesce = "true"
 	remove_children = "false"
 	consolidate = "true"
 }

```

This will trigger a update lifecycle method in the vm resource where you can `find the vm and apply the snapshot` ( the logic present in the create lifecycle of the revert snapshot ).

Any thoughts on this approach?

vsphere/resource_vsphere_snapshot_test.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		return nil
+	}
+}
+
+const testAccCheckVmSnapshotConfig_basic = `
+resource "vsphere_snapshot" "Test_terraform_cases"{
+ 
+	snapshot_name = "SnapshotForTestingTerraform"
```
girishramnani: Could you please add values for required parameters such as `vm_name` and `folder`.

vsphere/resource_vsphere_snapshot_test.go
```diff
@@ -0,0 +1,116 @@
...
+	if v := os.Getenv("VSPHERE_VM_NAME"); v == "" {
+		t.Fatal("env variable VSPHERE_VM_NAME must be set for acceptance tests")
+	}
+
+	if v := os.Getenv("VSPHERE_VM_FOLDER"); v == "" {
+		t.Fatal("env variable VSPHERE_VM_FOLDER must be set for acceptance tests")
+	}
+}
+
+func TestAccVmSnanpshot_Basic(t *testing.T) {
```
girishramnani: Could you please solve the typo here.

vsphere/resource_vsphere_snapshot_test.go
```diff
@@ -0,0 +1,117 @@
...
+			return fmt.Errorf("Error while getting the snapshot %v", snapshot)
+		}
+
+		return nil
+	}
+}
+
+const testAccCheckVmSnapshotConfig_basic = `
+resource "vsphere_snapshot" "Test_terraform_cases"{
+  	vm_name = "vmForTesting"
```
vancluever: Also just noticed here - can you make sure this curly bracer is properly spaced (ie: `resource "vsphere_snapshot" "Test_terraform_cases" {`)? If you have a way to filter specific text through `terraform fmt -`, that would be a good idea.

vancluever:
Hey @idmsubs - just noticed that there's no response yet to this one. Would you be able to follow up and address the review comments and concerns I had about the business case of these resources?

I'll keep the PR open for another week. Also, if it's not a good time, we can always just close it off and you can re-submit it at a later time.

Thanks again!

idmsubs:
@vancluever we are working on making required changes for incorporating review comments, hoping to get those done and committed by early in next week. For your complete usecase understanding question, yes we will post those details too.

Sorry for delayed response.

vancluever:
Hey @idmsubs, that would be helpful! Will wait to hear from you more on this one.

idmsubs:
@vancluever we have fixed your review comments and also resolved conflicts. Please review and let me know if any additional comments. For detailed usecase explanation will send it today.

vancluever:
Hey @idmsubs, one thing I still don't necessarily see here is the use case for this resource. Would you be able to review the concerns that I had regarding the necessity or feasibility for a VM snapshot resource in Terraform, given the concerns surrounding snapshot age and the potential for technical issues and data loss, and get back to me with your thoughts?

Again, I'm concerned around the fact that VM snapshots are not disk snapshots, are not designed to stick around for a long time, and have a lot of potential for data loss issues if not treated properly. Also, the revert resource does not really seem like it belongs in Terraform and would just be better suited for a CLI call at any in particular point in time.

Thanks!

idmsubs:
@vancluever vSphere snapshots are basically restore_points created of VM which can be used later to make VM running on particular restore_point by simply reverting snapshot.

Consider deployment of server with steps as (using Terraform) : 
- VM deployment with vSphere provider
- Addition of disk to VM
- Installation & configuration of some complex software on deployed VM
- Taking snapshot to create restore_point

Now, with use of deployed VM, user might require to take snapshot later to create additional restore_points. Similarly, can use revert resource to reverting back to particular snapshot, this will help in reverting back to particular point with less time along with user data being reverted. 

In case use doesn't use snapshot, if something went wrong with VM user will require to destroy VM, request for new VM deployment, this will also cause user data loss at particular point to which user can revert using snapshot.

I understand with Terraform infrastructure as a service support, snapshot resources might not help during complete infrastructure creation but they will definitely help for maintaining infrastructure. Please let me know if this helps in understanding use cases I am trying to achieve.

idmsubs:
@vancluever could you help me in completing review process, so I can merge changes.

Thanks

vancluever:
Hey @idmsubs, sorry for being silent for the last week, but I have been busy with trying to roll out some new resources and wanted to make sure I had the time to give this a thoughtful enough response.

I still personally have some big reservations about this resource. I understand the workflow that you have laid out above, but personally don't necessarily think it agrees with the philosophy of the Terraform workflow, as you have mentioned.

Rather than creating snapshots that, in my opinion, can be fragile over the course of a lifetime of a VM (especially when it could pan out to a long snapshot chain like you have laid out above), I would rather the engineer lean on configuration management for the scenario that you have laid out above, or backups, or a combination of both.

The ideal scenario would be using Packer to build a base virtual machine, which can abstract the complex configuration that you have mentioned above to the point where it can be generalized and turned into a template. Then, using Terraform, the VM could be either fully cloned, creating an independent copy, or a linked clone could be made, creating an "instance" off of the base image (more akin to what you would see in a modern cloud).

From there, backups (not snapshots) should be used to back up the machine, and regular restore tests should happen to ensure that the application can be restored to a reasonably recent working state.

Nonetheless, I want to recognize your hard work here and also the fact that not everyone Terraforms exactly the same - so I want to merge this, with the stipulation that the `vsphere_virtual_machine_snapshot_revert` resource be removed, as I really can't justfiy having this kind of resource in the provider. Instead, to support reverting to an existing snapshot, someone should be taking the ID of the snapshot produced by TF (as an output for example) and using an OOB process to revert it (ie: using `govc vm.snapshot.revert`).  If you can do that, we can move towards merging this.

I have also left more review comments inline in the current diff.

vsphere/resource_vsphere_snapshot.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+func resourceVSphereSnapshot() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceVSphereSnapshotCreate,
+		Read:   resourceVSphereSnapshotRead,
+		Delete: resourceVSphereSnapshotDelete,
+
+		Schema: map[string]*schema.Schema{
+			"vm_id": {
```
vancluever: Can you change this to reference the the UUID of the virtual machine versus the VM path name?

`vsphere_virtual_machine` supplies this via the `uuid` attribute, so it can be fetched in Terraform. From there, you can get the MoRef for the VM via the [`FindAllByUuid`](https://godoc.org/github.com/vmware/govmomi/object#SearchIndex.FindByUuid) method.

vsphere/resource_vsphere_snapshot.go
```diff
@@ -0,0 +1,186 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"datacenter": {
```
vancluever: After the VM reference has been changed to UUID, you can get rid of this attribute.

vsphere/resource_vsphere_snapshot.go
```diff
@@ -0,0 +1,186 @@
...
+		},
+	}
+}
+
+func resourceVSphereSnapshotCreate(d *schema.ResourceData, meta interface{}) error {
+	vm, err := findVM(d, meta)
+	if err != nil {
+		return fmt.Errorf("Error while getting the VirtualMachine :%s", err)
+	}
+	task, err := vm.CreateSnapshot(context.TODO(), d.Get("snapshot_name").(string), d.Get("description").(string), d.Get("memory").(bool), d.Get("quiesce").(bool))
```
vancluever: I'd like to get out of the pattern of the vSphere provider using general blocking contexts for these operations - this could potentially mean the user may have to wait 20 minutes for a failed API call to return.

Can you make sure to replace all calls to `context.TODO()` with a pattern similar to:

```go
ctx, cancel := context.WithTimeout(context.Background(), defaultAPITimeout) // This is 5 mins
defer cancel()
// your function that needs a context here
```

This will ensure that we put a timeout on function calls that are more reasonable for the operations that they are trying to perform.

Use a separate context for each call. If you find the code getting a little messy, break the code up into different functions. I'd recommend no more than 2 contexts per function with the latter being for waiting on tasks - like what is below:

```go
ctx, cancel := context.WithTimeout(context.Background(), defaultAPITimeout) // This is 5 mins
defer cancel()
task, err := object.FunctionThatNeedsContext(ctx)
if err != nil {
  return err
}
tctx, tcancel := context.WithTimeout(context.Background(), defaultAPITimeout)
defer tcancel()
info, err := task.WaitForResult(tctx, nil)
// Process info.Result here
```

vsphere/resource_vsphere_snapshot.go
```diff
@@ -0,0 +1,186 @@
...
+	log.Printf("[INFO] Task created for Create Snapshot: %v", task)
+	err = task.Wait(context.TODO())
+
+	if err != nil {
+		log.Printf("[ERROR] Error While waiting for the Task for Create Snapshot: %v", err)
+		return fmt.Errorf(" Error While waiting for the Task for Create Snapshot: %s", err)
+	}
+	log.Printf("[INFO] Create Snapshot completed %v", d.Get("snapshot_name").(string))
+	log.Println("[INFO] Managed Object Reference: " + taskInfo.Result.(types.ManagedObjectReference).Value)
+	d.SetId(taskInfo.Result.(types.ManagedObjectReference).Value)
```
vancluever: :+1: for using the MOID as the resource ID!

vsphere/resource_vsphere_revert_snapshot.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func resourceVSphereSnapshotRevert(d *schema.ResourceData, meta interface{}) error {
+	vm, err := findVM(d, meta)
+	if err != nil {
+		return fmt.Errorf("Error while getting the VirtualMachine :%s", err)
+	}
+	task, err := vm.RevertToSnapshot(context.TODO(), d.Get("snapshot_id").(string), d.Get("suppress_power_on").(bool))
+	if err != nil {
+		log.Printf("[ERROR] Error While Creating the Task for Revert Snapshot: %v", err)
```
vancluever: A note on logs: Terraform currently does nothing with `INFO` and `ERROR` messages that is anything more meaningful than `DEBUG`... can all of these messages be changed to `DEBUG`?

vsphere/resource_vsphere_snapshot.go
```diff
@@ -0,0 +1,186 @@
...
+			return nil
+		}
+		log.Printf("[ERROR] Error While finding the Snapshot: %v", err)
+		return fmt.Errorf("Error while finding the Snapshot :%s", err)
+	}
+	log.Printf("[INFO] Snapshot found: %v", snapshot)
+	return nil
+}
+
+func findVM(d *schema.ResourceData, meta interface{}) (*object.VirtualMachine, error) {
```
vancluever: Once this is changed to find by UUID, can you move this function to a new file called `virtual_machine_helper.go`? I want to try and start ensuring that all helpers are organized by their inparticular MO.

Also, make sure the function name is self-documenting - in this case, it would be `virtualMachineFromUUID`.

website/docs/r/virtual_machine_snapshot.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+
+# vsphere\_virtual\_machine\_snapshot
+
+Provides a VMware vSphere virtual machine snapshot resource. This can be used to create,
+delete, and revert virtual machine's snapshot.
+
+## Example Usage
+
+```hcl
+resource "vsphere_virtual_machine_snapshot" "demo1"{
```
vancluever: Can you filter this through `terraform fmt` so that it is correctly formatted?

vsphere/resource_vsphere_snapshot_test.go
```diff
@@ -0,0 +1,122 @@
...
+			return fmt.Errorf("Error while getting the snapshot %v", snapshot)
+		}
+
+		return nil
+	}
+}
+
+func testAccCheckVSphereVMSnapshotConfig_basic(vmId, snapshotName, description, memory, quiesce string) string {
+	return fmt.Sprintf(`
+	resource "vsphere_virtual_machine_snapshot" "Test_terraform_cases" {
```
vancluever: Can you filter this through `terraform fmt` so that it is correctly formatted?

website/docs/index.html.markdown
```diff
@@ -1,209 +1,229 @@
----
```
vancluever: Can you remove this file from the PR?

I will probably be revising the provider index doc soon, and at that point it will contain a more terse example, something more akin to the [AWS provider's index page](https://www.terraform.io/docs/providers/aws/index.html).

website/docs/r/virtual_machine_snapshot.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+1. for resource vsphere_virtual_machine_snapshot
+
+* `vm_id` - (Required) The virtual machine id (FolderName/VM_Name)
+* `snapshot_name` - (Required) New name for the snapshot.
+* `description` - (Required) New description for the snapshot.
+* `memory` - (Required) If the memory flag set to true, a dump of the internal state of the virtual machine is included in the snapshot.
+* `quiesce` - (Required) If the quiesce flag set to true, and the virtual machine is powered on when the snapshot is taken, VMware Tools is used to quiesce the file system in the virtual machine.
+* `datacenter` - (Optional) The name of a Datacenter in which the virtual machine exists for which we are taking snapshot
+* `remove_children` - (Optional) Flag to specify removal of the entire snapshot subtree.
+* `consolidate` - (optional) If set to true, the virtual disk associated with this snapshot will be merged with other disk if possible. Defaults to true.
```
vancluever: `optional` -> `Optional`

vancluever:
Hey @idmsubs, I think this LGTM now but there are some changes I will be making (while preserving your commits in place) pertaining to items that were missed from the review. I will be merging shortly.

Thank you very much for your work on this and your patience on the process!

vancluever:
Hey @idmsubs, just wanted to let you know the post-merge edits can be seen in the commit log in master. One significant change is I have changed the `vm_uuid` parameter to `virtual_machine_uuid`, just to keep with our convention of using full-word parameters (for the most part).

Anyway, this should be in next week's release, so stay tuned and thanks again!

PR #110, opened 2017-07-25
Enhancement: Added resource for license management

girishramnani:
Please review the implementation of license resource.

The following capabilities are provided by this resource - 

1) Adding license ( with optional labels )
2) Removing license
3) Updating labels attached to a license

Terraform version - v0.9.11
Vsphere version used for acceptance tests - Version 5.5.0 Build 4063694


vancluever:
Hi @girishramnani! Thank you very much for submitting this PR - it's going to be very useful!

I have been testing this against a standalone ESXi server and there seems to be some issues. I think this stems from the fact that the host API differs somewhat from the vCenter API. There are a couple of formatting issues as well that I hope could be looked at, just to make the tests look a bit better.

I will start a review and put the relevant information in, but just FYI a good reference to check what needs to happen to support both standalone host and vCenter when working with licenses can be found in the `govc` [`license.add` call](https://github.com/vmware/govmomi/blob/master/govc/license/add.go#L76).

If you have any questions or need any help feel free to ask! Looking forward to getting this one merged!

vancluever:
Hey @girishramnani - just wanted to let you know the review is still pending - sorry for the delay! In the meantime can you check out what I mentioned regarding handling the API that this points against (host vs vCenter) and make the needed changes?

PS: This is the line that needs fixing in `Create`: https://github.com/terraform-providers/terraform-provider-vsphere/pull/110/files/fb2068625fe0a3c63647156cb53e68e3d28ca7cd#diff-6da7e5115133f8bf52016c3cc74c771cR98

girishramnani:
Done, so now based on the api type different method to add license will be used.

website/docs/r/license.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+  
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `license_key` - (Required) The key of the license which is needed to be added to vpshere
+* `label` - (Optional) The key value pair of labels that has to be attached to the license key. One key can have multiple labels.
```
vancluever: I think this could probably do with more detail. Something like:

>  * `label` - (Optional) A set of labels to apply to the license key. May be specified multiple times to supply multiple label keys.
> 
> The values for `labels` are:
>  * `key` - (Required) The key for the label.
>  * `values` - (Required) A list of string values to assign to the label key.
> 

The latter note on `values` also takes into consideration my notes on the `labels` attribute in the schema itself - so check there for more details.

website/docs/r/license.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Attributes Reference
+
+The following attributes are exported:
```
vancluever: Can this section be cleaned up a bit too? Something like:

>  * `edition_key` - The product edition of the license key.
>  * `total` - Total number of units (example: vCPUs) contained in the license.
>  * `used` - The number of units (example: vCPUs) assigned to this license.
>  * `name` - The display name for the license.

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,287 @@
...
+	"errors"
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/license"
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/types"
+	"golang.org/x/net/context"
```
vancluever: Can this be changed to the standard library `context`? So just `context`, versus `golang.org/x/net/context`.

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,287 @@
...
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"value": &schema.Schema{
+							Type:     schema.TypeString,
```
vancluever: Noticed that you mention that labels can be have multiple values - can this be switched to a `TypeSet` so that all values can be specified in a single resource?

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,287 @@
...
+		Delete: resourceVSphereLicenseDelete,
+
+		Schema: map[string]*schema.Schema{
+			"license_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"label": &schema.Schema{
+				Type:     schema.TypeList,
```
vancluever: This should be a `TypeSet` so that conflicting entries can't be added.

Also, you will need a hashing function as well when converting it to a set - the stock [`HashResource` function](https://github.com/hashicorp/terraform/blob/master/helper/schema/set.go#L23) should help too.

vsphere/resource_vsphere_license.go
girishramnani: I don't think one label can have multiple values. Where have I mentioned that?

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil
+	}
+
+	if info := getLicenseInfoFromKey(d.Get("license_key").(string), manager); info != nil {
+		log.Println("[INFO] Setting the values")
+		d.Set("edition_key", info.EditionKey)
+		d.Set("total", info.Total)
+		d.Set("used", info.Used)
+		d.Set("name", info.Name)
+
```
girishramnani: If i update the labels here, then the state provided by terraform files and terraform state stored will drift apart, for example if server has labels `x:y` and local state shows it as `x:yy` then the server state has to be changed not the other way around. That's why only computed properties are set as they are the one which the user is not responsible for.

vsphere/resource_vsphere_license_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+func TestVSphereLicenseLabelsToMap(t *testing.T) {
+
+	labelMap, err := labelsToMap(labelStub)
+
+	if err != nil {
+		t.Fatal("Error ", err)
+	}
+
+	if value, ok := labelMap["Hello"]; !ok || value != "World" {
```
girishramnani: I think it would be better to add a check for the keys that are not being tested here.

website/docs/r/license.html.markdown
girishramnani: `values` is not a list of string, `value` is a string. Does the api support multiple values for a single key?

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,275 @@
...
+	} else {
+		return ErrNoSuchKeyFound
+	}
+
+	return nil
+
+}
+
+// resourceVSphereLicenseUpdate check for change in labels of the key and updates them.
+// Change in key would remove it from
```
girishramnani: Sorry, this sentence was not supposed to be there.

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,275 @@
...
+			},
+			"label": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
```
girishramnani: Wouldn't `ForceNew` here mean that if `key`of this label change that means create a new label instead of changing?

website/docs/r/license.html.markdown
vancluever: I think my confusion was in this part:

> One key can have multiple labels

I did double check as well, and a label can't have multiple values - sorry about that! So I think what you are saying here is that a in particular license can have multiple labels, right? I think in that case, it might be a good idea to just drop this sentence altogether, considering the double meaning of `key` (license key and label key) may cause confusion.

girishramnani:
yeah, but is there a better name for the label's key attribute?

vsphere/resource_vsphere_license_test.go
vancluever: So you are saying test the entire set? I think that makes sense too - just as long as the fixture as a whole ultimately has use. :slightly_smiling_face:

vsphere/resource_vsphere_license.go
vancluever: This is actually the intended behaviour of Terraform - it should detect this kind of drift (ie: if a user manually adds a label) and bring it back to the correctly configured state. It encourages management to be done out of configuration, treating it as the source of truth. 

When someone wants to ignore drift, the correct thing to do in a Terraform configuration is to add [`ignore_changes`](https://www.terraform.io/docs/configuration/resources.html#ignore_changes) for the attributes they want to ignore:

```
resource "vsphere_license" "license" {
  ...

  lifecycle {
    ignore_changes = ["label"]
  }
}
```

vsphere/resource_vsphere_license.go
vancluever: I was confused about this - sorry again! I put some notes in the documentation comment relevant to this.

vsphere/resource_vsphere_license.go
vancluever: PS: In light of the misunderstanding over label data - this could possibly be moved to a `TypeMap` to make things easier as well, seeing as no complex resource stuff needs to happen on this being that it's just a set of key/value pairs.

girishramnani:
But then in the next apply terraform will detect a difference in the local
state and the state specified by the terraform files which will trigger an
update. Due to this labels added by the user manually would be removed. Am
I correct?

vsphere/resource_vsphere_license.go
girishramnani: But then in the next apply terraform will detect a difference in the local
state and the state specified by the terraform files which will trigger an
update. Due to this labels added by the user manually would be removed. Am
I correct?

website/docs/r/license.html.markdown
girishramnani: instead of `key`, I can write `license_key can have multiple labels`

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,275 @@
...
+
+func resourceVSphereLicenseRead(d *schema.ResourceData, meta interface{}) error {
+
+	log.Println("[INFO] Running the read method")
+
+	client := meta.(*govmomi.Client)
+	manager := license.NewManager(client.Client)
+
+	// the key was never present. Should set the ID to false.
+	if _, ok := d.GetOk("license_key"); !ok {
```
girishramnani: Yeah, makes sense. Removing this part.

vsphere/resource_vsphere_license.go
girishramnani: Could you please clarify which approach is going to be used here, `TypeMap` or `TypeSet`

girishramnani:
@vancluever thank you for such a quick and thorough review. I have done the changes according to the comments could you please review them.

website/docs/r/license.html.markdown
```diff
@@ -32,15 +28,15 @@ resource "vsphere_license" "licenseKey" {
 
 The following arguments are supported:
 
-* `license_key` - (Required) The key of the license which is needed to be added to vpshere
-* `label` - (Optional) The key value pair of labels that has to be attached to the license key. One key can have multiple labels.
+* `license_key` - (Required) The license key to add.
+* `labels` - (Optional) The key value pair of labels that has to be attached to the license key.
```
vancluever: Looking good here! Thank you for the fix!

Just a couple of remaining nitpicks on the `labels` line:
 * Can we indicate that it's a map now?
 * Can we say `key/value` instead of `key value`?
 * It says `has to be attached`, but it's optional.

Something like:

> `labels` - (Optional) A map of key/value pairs to be attached as labels (tags) to the license key.

vsphere/resource_vsphere_license.go
vancluever: Hey @girishramnani, that is correct, but as mentioned above, this is intended Terrafrom behaviour - resources should always be set up so that they reconcile state with config, even if resources were intentionally modified by a user outside of Terraform. `ignore_changes` is explicitly used to disregard such changes when that's intentional.

vsphere/resource_vsphere_license_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+func testAccVSphereLicenseInvalidConfig() string {
+
+	// quite sure this key cannot be valid
+	return `resource "vsphere_license" "foo" {
+  					license_key = "HN422-47193-58V7M-03086-0JAN2"
+			}`
+}
+
+func testAccVSphereLicenseWithLabelConfig(labels map[string]string) string {
```
vancluever: We are getting close here - remember though that we don't necessarily need to render the labels here dynamically - it's better to just use a static string, and roll the function into a much more simpler form:

```go
func testAccVSphereLicenseWithLabelConfig() string {
  return fmt.Sprintf(`
resource "vsphere_license" "foo" {
  license_key = "%s"

  labels {
    VpxClientLicenseLabel = "Hello World"
    TestTitle = "FooBar"
  }  
}
`, os.Getenv("VSPHERE_LICENSE"))
}
```

vsphere/resource_vsphere_license_test.go
```diff
@@ -0,0 +1,225 @@
...
+	for key, value := range labels {
+		val += fmt.Sprintf(`
+			%s = "%s"
+		`, key, value)
+
+	}
+	return val
+}
+
+func testAccVSphereLicenseBasicConfig() string {
```
vancluever: As per the label generator here, remember we should style this function the same way:

```go
func testAccVSphereLicenseWithLabelConfig() string {
  return fmt.Sprintf(`
resource "vsphere_license" "foo" {
  license_key = "%s"
}
`, os.Getenv("VSPHERE_LICENSE"))
}
```

vancluever:
Hey @girishramnani thanks for the updates! Added some comments inline for the docs and tests stuff that still needs to be adjusted.

The following items are still outstanding:

 * Ensure labels are being saved to state on `Read` calls
 * Fix spurious whitespace at the start of functions - there are still several places that have them - it would be useful to go over the whole set of changes and address this. One was also added in 8f6e12739060a5f19c0271105d6b9ac5a968196d. I think maybe there's a misunderstanding that these lines should be *removed*, not added.
 * Just had a couple of nitpicks about language on the `labels` attribute and then that's good to go.
 * We are making progress on the tests but I still would like to see the config generation functions cleaned up a bit.

After this work, this also needs to be fully tested against an ESXi host. From what I read in the [`LicenseManager`](http://pubs.vmware.com/vsphere-6-5/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvim.LicenseManager.html) object, tags are ignored by `AddLicense` on ESXi, which might be causing the errors I was seeing in these tests on ESXi (the test returned tag not found after a successful call to add the license).

Thank you for your patience and continued work on this!

girishramnani:
@vancluever thanks once again for the review, I have changed the code according to the comments, could you please review it.

vancluever:
Hey @girishramnani - making more progress!

Here are the current test results:

### vCenter

```
=== RUN   TestAccVSphereLicenseBasic
--- PASS: TestAccVSphereLicenseBasic (3.32s)
=== RUN   TestAccVSphereLicenseInvalid
--- PASS: TestAccVSphereLicenseInvalid (0.99s)
=== RUN   TestAccVSphereLicenseWithLabels
--- FAIL: TestAccVSphereLicenseWithLabels (0.00s)
	testing.go:428: Step 0 error: Error loading configuration: Error parsing /tmp/tf-test756435681/main.tf: At 6:21: Unknown token: 6:21 IDENT FooBar
```

### Standalone ESXi

This first round of tests was with a license that I don't think is valid for ESXi at all (vSAN). I think better error trapping needs to happen to ensure that invalid license types for a node (not just invalid keys) are caught.

```
=== RUN   TestAccVSphereLicenseBasic
--- FAIL: TestAccVSphereLicenseBasic (1.35s)
	testing.go:428: Step 0 error: Error applying: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: The key was not found
	testing.go:492: Error destroying resource! WARNING: Dangling resources
		may exist. The full state and error is shown below.
		
		Error: Error refreshing: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: vsphere_license.foo: The key was not found
		
		State: <nil>
=== RUN   TestAccVSphereLicenseInvalid
--- FAIL: TestAccVSphereLicenseInvalid (1.31s)
	testing.go:422: Step 0, expected error:
		
		Error applying: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: The key was not found
		
		To match:
		
		License file not found
		
		
	testing.go:492: Error destroying resource! WARNING: Dangling resources
		may exist. The full state and error is shown below.
		
		Error: Error refreshing: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: vsphere_license.foo: The key was not found
		
		State: <nil>
=== RUN   TestAccVSphereLicenseWithLabels
--- FAIL: TestAccVSphereLicenseWithLabels (0.00s)
	testing.go:428: Step 0 error: Error loading configuration: Error parsing /tmp/tf-test872458016/main.tf: At 6:21: Unknown token: 6:21 IDENT FooBar
```

This second test was with an Enterprise Plus key. This one succeeds minus the issues with labels and the broken config that is causing issues with all of the tests (see below).

```
=== RUN   TestAccVSphereLicenseBasic
--- PASS: TestAccVSphereLicenseBasic (3.28s)
=== RUN   TestAccVSphereLicenseInvalid
--- FAIL: TestAccVSphereLicenseInvalid (1.32s)
	testing.go:422: Step 0, expected error:
		
		Error applying: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: The key was not found
		
		To match:
		
		License file not found
		
		
	testing.go:492: Error destroying resource! WARNING: Dangling resources
		may exist. The full state and error is shown below.
		
		Error: Error refreshing: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: vsphere_license.foo: The key was not found
		
		State: <nil>
=== RUN   TestAccVSphereLicenseWithLabels
--- FAIL: TestAccVSphereLicenseWithLabels (0.00s)
	testing.go:428: Step 0 error: Error loading configuration: Error parsing /tmp/tf-test762915504/main.tf: At 6:21: Unknown token: 6:21 IDENT FooBar
```

So looks like the config needs to be fixed on the label test, and also the outstanding label tests need to be corrected against the ESXi hosts (might be an idea to kick back an error if a user is trying to define labels on an ESXi server versus vCenter, to prevent confusion).

Also, the formatting needs to be changed on your config generation functions still (again, change them to a simple form where the raw string is not indented and you are using `os.Getenv` directly in the `fmt.Sprintf` call).

Thank you for the continued attention on this one!

vsphere/resource_vsphere_license_test.go
```diff
@@ -0,0 +1,193 @@
...
+	// precheck already checks if this is present or not
+	key := os.Getenv("VSPHERE_LICENSE")
+	return fmt.Sprintf(`resource "vsphere_license" "foo" {
+  							license_key = "%s"
+						}
+						`, key)
+}
+
+func testAccVSphereLicenseDestroy(s *terraform.State) error {
+	client := testAccProvider.Meta().(*govmomi.Client)
```
vancluever: Hey @girishramnani, while not necessarily a panic that affects users, this panics when there is an error setting up the client (ie: if one forgets to skip SSL verification). Can you add something to trap nil in this situation and return an error?

girishramnani: @vancluever So should I change it in all the places? Like other tests and methods.

girishramnani:
Hello @vancluever, I have solved the invalid config. Could you please run the tests once more and provide the results. Also a small doubt, could you please help me understand why `TestAccVSphereLicenseInvalid` test is failing in cases of VSAN key and enterprise key? This test doesn't even depend on the license key.

vancluever:
Hey @girishramnani, I think the issues with both the invalid license tests and the issues with adding the VSAN key might be due to the differences between the vCenter and host APIs. I tried a completely invalid key and got the same kind of error. Just using `govc license.add -dump ABCDE-11111-11111-11111-11111`:

## vCenter

The error that the code currently looks for is displayed here correctly.

```
(license.licenseOutput) (len=1 cap=1) {
    (types.LicenseManagerLicenseInfo) {
        DynamicData: (types.DynamicData) {
        },
        LicenseKey: (string) "",
        EditionKey: (string) "",
        Name: (string) "",
        Total: (int32) 0,
        Used: (int32) 0,
        CostUnit: (string) "",
        Properties: ([]types.KeyAnyValue) (len=4 cap=4) {
            (types.KeyAnyValue) {
                DynamicData: (types.DynamicData) {
                },
                Key: (string) (len=8) "lc_error",
                Value: (int32) 5
            },
            (types.KeyAnyValue) {
                DynamicData: (types.DynamicData) {
                },
                Key: (string) (len=19) "localizedDiagnostic",
                Value: (types.LocalizableMessage) {
                    DynamicData: (types.DynamicData) {
                    },
                    Key: (string) (len=38) "com.vmware.vim.vc.license.error.decode",
                    Arg: ([]types.KeyAnyValue) (len=1 cap=4) {
                        (types.KeyAnyValue) {
                            DynamicData: (types.DynamicData) {
                            },
                            Key: (string) (len=10) "diagnostic",
                            Value: (string) (len=19) "licenseFileNotFound"
                        }
                    },
                    Message: (string) (len=22) "License file not found"
                }
            },
            (types.KeyAnyValue) {
                DynamicData: (types.DynamicData) {
                },
                Key: (string) (len=10) "diagnostic",
                Value: (string) (len=37) "License is not valid for this product"
            },
            (types.KeyAnyValue) {
                DynamicData: (types.DynamicData) {
                },
                Key: (string) (len=9) "Localized",
                Value: (types.ArrayOfKeyAnyValue) {
                    KeyAnyValue: ([]types.KeyAnyValue) <nil>
                }
            }
        },
        Labels: ([]types.KeyValue) <nil>
    }
}
```


## ESXi

Here, the error isn't, so the code is assuming that the add is successful. This is the same dump profile I get when trying to add the VSAN key.

```
(license.licenseOutput) (len=1 cap=1) {
    (types.LicenseManagerLicenseInfo) {
        DynamicData: (types.DynamicData) {
        },
        LicenseKey: (string) (len=29) "ABCDE-11111-11111-11111-11111",
        EditionKey: (string) "",
        Name: (string) "",
        Total: (int32) 0,
        Used: (int32) 0,
        CostUnit: (string) "",
        Properties: ([]types.KeyAnyValue) (len=46 cap=64) {
            (types.KeyAnyValue) {
                DynamicData: (types.DynamicData) {
                },
                Key: (string) (len=10) "diagnostic",
                Value: (string) (len=37) "License is not valid for this product"
            },
            (types.KeyAnyValue) {
                DynamicData: (types.DynamicData) {
                },
                Key: (string) (len=11) "ProductName",
                Value: (string) (len=17) "VMware ESX Server"
            },
        ...
    }
}
```

It looks like both errors have the common "License is not valid for this product" un-localized diagnostic `KeyAnyValue` - maybe we should filter off of that instead? If we do that, I think there's a good chance all of these tests will pass.

PS: Looks like the raw strings for the config generation functions still need to be fixed. Remember that it needs to be **un**-indented. Example: [this config](https://github.com/terraform-providers/terraform-provider-vsphere/pull/110/files#diff-2b6a073299181154110b8dd5c8a5de24R85) should be generated more like:

```
func testAccVSphereLicenseWithLabelConfig() string {
  return fmt.Sprintf(`
resource "vsphere_license" "foo" {
  license_key = "%s"
  labels {
    VpxClientLicenseLabel = "Hello World"
    TestTitle = "FooBar"
  }		 	
}
`, os.Getenv("VSPHERE_LICENSE"))
}
```

PPS: Also, can you rebase this PR off of `master` and re-vendor the license packages of `gvmomi` off of v0.15.0? Both of these things have changed since this PR was submitted.

girishramnani:
Hello @vancluever , I have added error and Error checks for ESXi server, removed indentation and checked for invalid license instead of `decode.error`. Could you please review it.

website/docs/r/license.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+* `license_key` - (Required) The license key to add.
+* `labels` - (Optional) A map of key/value pairs to be attached as labels (tags) to the license key.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `edition_key` - The product edition of the license key.
+* `total` - Total number of units (example: vCPUs) contained in the license.
```
vancluever: Sorry I missed this before @girishramnani but this needs to be changed to CPUs as well - I don't think I've seen a license have vCPUs at all.

girishramnani: Sure, changing.

girishramnani:
Hello @vancluever , changed the docs are per the request.

vancluever:
Hey @girishramnani, getting close now. Tests all pass on ESXi, but now on vCenter we get:

```
=== RUN   TestAccVSphereLicenseBasic
--- PASS: TestAccVSphereLicenseBasic (2.10s)
=== RUN   TestAccVSphereLicenseInvalid
--- PASS: TestAccVSphereLicenseInvalid (0.54s)
=== RUN   TestAccVSphereLicenseWithLabels
--- FAIL: TestAccVSphereLicenseWithLabels (0.61s)
	testing.go:422: Step 0, expected error:

		Error applying: 1 error(s) occurred:

		* vsphere_license.foo: 1 error(s) occurred:

		* vsphere_license.foo: No description for lc_error code

		To match:

		Labels are not allowed in ESXi


FAIL
exit status 1
```

I think [this](https://github.com/terraform-providers/terraform-provider-vsphere/pull/110/files#diff-2b6a073299181154110b8dd5c8a5de24R73) is the source of the issue - can we split the tests up into vSphere and ESXi-specific tests? Maybe let's add the `VSPHERE_TEST_ESXI` environment variable to indicate if we are testing against ESXi or not and skip a tests based on the results.

The last item would be the test code - the raw strings are still indented - sorry if this is not 100% clear but I need it laid out in a fashion similar to [this](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_ssm_maintenance_window_task_test.go#L130). If there's maybe something keeping you from doing this (ie: IDE issues), we can just merge this and make the change after the fact. Let me know!

girishramnani:
Hello @vancluever , Sorry about the space issue. I have split the test in VCenter and ESXi specific tests and also removed the extra space. Thank you for your patience on this PR.

vancluever:
Hey @girishramnani - getting closer!

Test results:

### vCenter

```
=== RUN   TestAccVSphereLicenseBasic
--- PASS: TestAccVSphereLicenseBasic (3.37s)
=== RUN   TestAccVSphereLicenseInvalid
--- PASS: TestAccVSphereLicenseInvalid (0.89s)
=== RUN   TestAccVSphereLicenseWithLabelsOnVCenter
--- FAIL: TestAccVSphereLicenseWithLabelsOnVCenter (0.91s)
	testing.go:428: Step 0 error: Error applying: 1 error(s) occurred:
		
		* vsphere_license.foo: 1 error(s) occurred:
		
		* vsphere_license.foo: No description for lc_error code
=== RUN   TestAccVSphereLicenseWithLabelsOnESXiServer
--- SKIP: TestAccVSphereLicenseWithLabelsOnESXiServer (0.00s)
	resource_vsphere_license_test.go:109: VSPHERE_TEST_ESXI must be set to true for this acceptance test
```

### ESXi

```
=== RUN   TestAccVSphereLicenseBasic
--- PASS: TestAccVSphereLicenseBasic (3.31s)
=== RUN   TestAccVSphereLicenseInvalid
--- PASS: TestAccVSphereLicenseInvalid (0.79s)
=== RUN   TestAccVSphereLicenseWithLabelsOnVCenter
--- SKIP: TestAccVSphereLicenseWithLabelsOnVCenter (0.00s)
	resource_vsphere_license_test.go:103: VSPHERE_TEST_ESXI must not be set for this acceptance test
=== RUN   TestAccVSphereLicenseWithLabelsOnESXiServer
--- PASS: TestAccVSphereLicenseWithLabelsOnESXiServer (0.71s)
```

Can you review what might be going on with `TestAccVSphereLicenseWithLabelsOnVCenter` and fix accordingly?

Also - still need to fix one config generator (it could actually be moved to a constant). I will comment it.

Thanks!

vsphere/resource_vsphere_license_test.go
```diff
@@ -0,0 +1,225 @@
...
+func testAccVspherePreLicenseESXiServerIsSetCheck(t *testing.T) {
+	key, err := strconv.ParseBool(os.Getenv("VSPHERE_TEST_ESXI"))
+	if err != nil || !key {
+		t.Skip("VSPHERE_TEST_ESXI must be set to true for this acceptance test")
+	}
+}
+
+func testAccVSphereLicenseInvalidConfig() string {
+	// quite sure this key cannot be valid
+	return `resource "vsphere_license" "foo" {
```
vancluever: Hey @girishramnani, looks like this one is still indented - might be worthwhile to just move this to a constant too as nothing in it is being generated.

girishramnani:
Hello @vancluever, I cannot reproduce the `lc_error` on VCenter  could you please share the verbose logs.

girishramnani:
Hello @vancluever, this is the output of my tests

```

TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run License -timeout 120m
?       bitbucket.org/girishramnani/terraform_vsphere    [no test files]
=== RUN   TestAccVSphereLicenseBasic
--- PASS: TestAccVSphereLicenseBasic (1.47s)
=== RUN   TestAccVSphereLicenseInvalid
--- PASS: TestAccVSphereLicenseInvalid (0.16s)
=== RUN   TestAccVSphereLicenseWithLabelsOnVCenter
--- PASS: TestAccVSphereLicenseWithLabelsOnVCenter (0.78s)
=== RUN   TestAccVSphereLicenseWithLabelsOnESXiServer
--- SKIP: TestAccVSphereLicenseWithLabelsOnESXiServer (0.00s)
        resource_vsphere_license_test.go:115: VSPHERE_TEST_ESXI must be set to true for this acceptance test
PASS
ok      bitbucket.org/girishramnani/terraform_vsphere/vsphere    2.455s

```

Vsphere version used for acceptance tests - Version 5.5.0 Build 4063694. I am not able to reproduce the error, could you please provide me the `govc` debug logs.

vancluever:
Hey @girishramnani - interesting! I am testing against vSphere 6.5 U1, so there definitely is a version discrepancy there that we may need to investigate. I will make a note to get the info you need tomorrow morning!

vancluever:
Hey @girishramnani, here's an XML of one the request's output:

```
<?xml version="1.0" encoding="UTF-8"?>
<soapenv:Envelope xmlns:soapenc="http://schemas.xmlsoap.org/soap/encoding/" xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soapenv:Body>
    <AddLicenseResponse xmlns="urn:vim25">
      <returnval>
        <licenseKey/>
        <editionKey/>
        <name/>
        <total>0</total>
        <costUnit/>
        <properties>
          <key>lc_error</key>
          <value xsi:type="xsd:int">-1</value>
        </properties>
        <properties>
          <key>localizedDiagnostic</key>
          <value xsi:type="LocalizableMessage">
            <key>com.vmware.vim.vc.license.error.decode</key>
            <arg>
              <key>diagnostic</key>
              <value xsi:type="xsd:string">unknownError</value>
            </arg>
            <message>Unknown license decode error</message>
          </value>
        </properties>
        <properties>
          <key>diagnostic</key>
          <value xsi:type="xsd:string">No description for lc_error code</value>
        </properties>
        <properties>
          <key>Localized</key>
          <value xsi:type="ArrayOfKeyAnyValue"/>
        </properties>
      </returnval>
    </AddLicenseResponse>
  </soapenv:Body>
</soapenv:Envelope>
```

This is happening on creation. I kind of wonder if we need to work around this by doing an initial create without the labels and then update it afterwards. [Source](https://github.com/vmware/govmomi/blob/f9184c1d704efa615d419dd8d1dae1ade94701d1/govc/license/add.go#L69). Will check and then update the issue.

vancluever:
Hey @girishramnani, looks like this was indeed the issue.

Here is a very non-optimal sketch of my updated workflow:

```
		info, err = manager.Add(context.TODO(), key, nil)
		if err != nil {
			return err
		}
		if err = DecodeError(info); err != nil {
			return err
		}
		d.SetId(info.LicenseKey)
		return resourceVSphereLicenseUpdate(d, meta)
```

After this, the test passes:

```
=== RUN   TestAccVSphereLicenseWithLabelsOnVCenter
--- PASS: TestAccVSphereLicenseWithLabelsOnVCenter (4.32s)
```

I think there is some re-factoring that could be done too. I will add annotations.

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,246 @@
...
+	var info types.LicenseManagerLicenseInfo
+	var err error
+	switch t := client.ServiceContent.About.ApiType; t {
+	case "HostAgent":
+		// Labels are not allowed in ESXi
+		if len(finalLabels) != 0 {
+			return errors.New("Labels are not allowed in ESXi")
+		}
+
+		info, err = manager.Update(context.TODO(), key, finalLabels)
```
vancluever: Since labels can't be used on ESXi period, it might be good to just remove `finalLabels` here and just use `nil`.

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,246 @@
...
+	case "VirtualCenter":
+		info, err = manager.Add(context.TODO(), key, finalLabels)
```
vancluever: As mentioned in the PR main comments, this needs to be re-factored into a two-step process - one where the license is added, and another where the labels are added afterwards. I think what you could probably do is break off part of the logic in `Update` into its own function, and then call out to it here. There is possibly some other re-factoring that could be done here too, considering the possible divergence of workflows.

vsphere/resource_vsphere_license.go
```diff
@@ -0,0 +1,246 @@
...
+	client := meta.(*govmomi.Client)
+	manager := license.NewManager(client.Client)
+
+	if key, ok := d.GetOk("license_key"); ok {
+		licenseKey := key.(string)
+		if !isKeyPresent(licenseKey, manager) {
+			return ErrNoSuchKeyFound
+		}
+
+		if d.HasChange("labels") {
```
vancluever: Again, this is the logic that should probably be spun off into its own function so it can be used in `Create` as well to add the labels as a separate step after the initial addition of the license is performed.

girishramnani:
Hello @vancluever, I have updated the code according to your comments could you please review it.

girishramnani:
Hello @vancluever, the above commit a small fix which adds a error check statement for add license.

vsphere/resource_vsphere_license.go
girishramnani: Done

girishramnani:
Rebasing

girishramnani:
Hello @vancluever, thank you for merging master. Not rebasing.

vancluever:
@girishramnani - of course - it's important to us that work gets preserved! I needed to fix the conflicts on the website doc index and wanted to make sure I didn't have to force push.

This now LGTM! I am merging this shortly - but thank you **very** much for your patience on this one and your response to my feedback! I will be releasing a new version of the provider next week which will have your work in it :+1: 

Thanks again!

PR #111, opened 2017-07-25
virtual machine - add annotation argument

verdel:
In our environment, we use the annotation field to store additional information about the virtual machine. I added the ability to use this field when creating a virtual machine.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -86,6 +86,7 @@ type virtualMachine struct {
 	vcpu                  int32
 	memoryMb              int64
 	memoryAllocation      memoryAllocation
+	annotation            string
 	template              string
 	networkInterfaces     []networkInterface
 	hardDisks             []hardDisk
@@ -153,6 +154,11 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 				Optional: true,
 				Default:  0,
 				ForceNew: true,
+			},
+
+			"annotation": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
 			},
 
 			"datacenter": &schema.Schema{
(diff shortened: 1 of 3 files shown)


vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1776,6 +1794,7 @@ func (vm *virtualMachine) setupVirtualMachine(c *govmomi.Client) error {
 		Flags: &types.VirtualMachineFlagInfo{
 			DiskUuidEnabled: &vm.enableDiskUUID,
 		},
+		Annotation: vm.annotation,
```
grubernaut: Does this fail if `vm.annotation` is `""`?

verdel: No, this doesn`t fail with empty string in vm.annotaton. So everything works fine as with an empty argument and absolutely without it. In both of these situations vm.annotation takes a value "". [vsphere/resource_vsphere_virtual_machine.go (L710-L714)](https://github.com/verdel/terraform-provider-vsphere/blob/8624dabc7e5517fbd09eeaf78b25c7eb74f6193f/vsphere/resource_vsphere_virtual_machine.go#L710-L714)

rismoney:
is this only the notes field, or can it handle custom annotations?

verdel:
It is only for the notes fields. This option cannot manage custom attributes or tags.

rismoney:
was that a limitation of this PR or govmomi?
The reason i ask is because I would love to be able to add "commit: sha1" as an annotation/custom attribute:value.  Notes is already used by 3rd party polluters.  Appreciate the work btw! kudos on merge.

verdel:
It is a limitation of this PR only. Notes(annotation in vsphere model) and custom attributes it is different properties in vsphere object model.

Issue #64, opened 2017-06-13
provider/vsphere: not possible to set disk order on a virtual machine


hashibot:
_This issue was originally opened by @moserke as hashicorp/terraform#10225. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Is there a way to specify the disk order for vsphere virtual machines?  I am currently having a problem in that I need to know which /dev the disk is getting attached to and it's non-deterministic, because it depends on the SCSI device it's attached to, and that order is not being maintained.  Sometimes it will be SCSI 0:2 so /dev/sdb sometimes SCSI 0:3 so /dev/sdc.  Can I specify the SCSI device number so I can force the dev order?

notuscloud:
Hi, I've been facing the issue with terraform 0.9.11. It's quite a pain that it's not possible to order disk creation

vancluever:
Hey all,

This inparticular feature is now available in the 1.x version of `vsphere_virtual_machine`. See the [resource documentation](https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html) for more details on usage.

Thanks!

PR #115, opened 2017-07-31
Changes the disk schema from TypeSet to TypeList

r2bit:
My rather crude way to fix (or rather change the behaviour) #64.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -124,7 +124,7 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 		Update: resourceVSphereVirtualMachineUpdate,
 		Delete: resourceVSphereVirtualMachineDelete,
 
-		SchemaVersion: 1,
+		SchemaVersion: 2,
 		MigrateState:  resourceVSphereVirtualMachineMigrateState,
 
 		Schema: map[string]*schema.Schema{
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	"golang.org/x/net/context"
+	"reflect"
 )
 
 var DefaultDNSSuffixes = []string{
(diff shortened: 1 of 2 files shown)


catsby:
For users with existing disks, what do they get when they upgrade to this new code? Do they get a diff in their plan?

girishramnani:
@catsby probably not and as this is a schema change implementing a [StateMigrateFunc](https://godoc.org/github.com/hashicorp/terraform/helper/schema#StateMigrateFunc) could be required to support users who already have resources.

More about MigrateFunc is in [Source Code](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource.go#L47)

r2bit:
@catsby Unfortunately it will already crash in the planning state - it will attempt to read the old TypeSet state in as TypeList (as defined in the schema change) and crash on <nil>. The crash itself can be quite easily handled, but it does not give a smooth migration path, still.. I will look into StateMigrateFunc, although at first glance it does not look like it would help, since i'm not changing elements, but the schema itself.

girishramnani:
@r2bit according to the doc of migrateFunc

```
// MigrateState is responsible for updating an InstanceState with an old
// version to the format expected by the current version of the Schema.

```

The function is responsible for transforming the local state ( which was created using older schema ) to the newer state ( newer schema ). This will be called before refresh and hence the local state will be updated before terraform reaches the ResourceRead method.

So I think StateMigrateFunc could solve the problem.

r2bit:
@girishramnani Thanks, I'll give it a shot in the evening :)

r2bit:
I added migration. However, there's still one open issue: how to handle reordering? Reorder disks on VM by force (i.e. detach and attach all unordered disks) or simply ignore the positions of existing disks.
The current solution will do the latter to avoid any data loss during forceful disk detach (and FS unmount by extension). But i feel that is not the correct approach - TF state should reflect the exact state after successful apply (?)

vancluever:
Hey all,

This seems to be a rather old PR (sorry about that), and how has been largely superseded by #244. The rewrite does not move disks to lists, but rather exposes a `unit_number` attribute that can be used to set the unit number directly on the SCSI bus, allowing for 60 disks to possibly be attached across 4 SCSI controllers.

The only real issue remaining with disks still being sets are that diffs are rather ugly as we don't have any way of intelligently lining up sets in a diff in TF core. However, I think this was not necessarily an objective of this PR as it seems to attempt to address #64, which `unit_number` should sufficiently do.

I still want to explore moving disks to lists at a later date to address the diff user experience issues, however for v1.0.0, we will be sticking with sets, as order has been addressed now and we do have documentation on how to read the disk diff in the new version of the docs, which should help mitigate any confusion.

Closing this now considering all of the above. Thanks for all the effort on this!

PR #116, opened 2017-08-04
 Add multiple virtual core support closes #42 

### LINKED ISSUE #42: vsphere: Multiple cores per vcpu
_This issue was originally opened by @datroup as hashicorp/terraform#8446. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,

I cannot find this in any of the documentation, but when creating vm's in vsphere with terraform we're unable to set the number of cores per vcpu. Is this possible, or can this be a feature request? I know from a technical point of view it doesn't matter but we have a requirement for this to be set.

D

erbriones:
Add support for customizing the number of cores per socket. Enforces
that the number of virtual CPUs is evenly divisible by the selected
number of virtual cores for the optimal socket size.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -84,6 +84,7 @@ type virtualMachine struct {
 	resourcePool          string
 	datastore             string
 	vcpu                  int32
+	vcores_per_socket     int32
 	memoryMb              int64
 	memoryAllocation      memoryAllocation
 	annotation            string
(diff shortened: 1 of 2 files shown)


angelichorsey:
This feature is actually needed considering how Windows licensing works, and it looks like it was easier than I thought it would be by making the user verify a mod of zero for vcpu and cores per socket.

The only issue I see is that `vcores_per_socket` detracts from the style of Go and the file it is within. I can't find a consistent online explanation for naming, but the Kernighan Golang book states to use camel case over underscore separation.

Using `vcoresPerSocket` would be better IMO.

PR #121, opened 2017-08-09
Test environment RC file, update base and drop inner README

vancluever:
A couple of things:

 * Adding the ability to configure your environment from an included rc-ish makefile from your own home directory (882c466)
 * Clean up/move the README to the top level and remove the old one (00543d3). I want this to be the source of truth along with the docs on the official website, and the one in the `vsphere` directory seems redundant on a few levels. I may restore some of the docs on how to create a test role in the future, or just direct contributors towards using strictly test environments (ie: using nested ESXi) to do plugin development.


vancluever:
PS: I'm still reviewing the other tests to see what variables are required for them as well.

vancluever:
Thanks @mbfrahry! Going to merge now, will drop in a small PR for the other config vars once they are identified.

PR #126, opened 2017-08-11
Enhancement: New resource datacenter

jorgenunez:
Add a new resource ```vsphere_datacenter```, used as the primary container for many objects, including all the current resources for this provider.

Tested on vSphere 6.5.

### DIFF
--- a/vsphere/resource_vsphere_datacenter_test.go
+++ b/vsphere/resource_vsphere_datacenter_test.go
@@ -12,14 +12,21 @@ import (
 	"golang.org/x/net/context"
 )
 
+const resourceName = "vsphere_datacenter.testDC"
+
+const testAccCheckVSphereDatacenterConfig = `
+  resource "vsphere_datacenter" "testDC" {
+    name = "testDC"
+  }`
+
+const testAccCheckVSphereDatacenterConfigSubfolder = `
+  resource "vsphere_datacenter" "testDC" {
+    name = "testDC"
+    folder = "%s"
+  }`
+
 // Create a datacenter on the root folder
 func TestAccVSphereDatacenter_createOnRootFolder(t *testing.T) {
-	const resourceName = "vsphere_datacenter.testDC"
-
-	const testAccCheckVSphereDatacenterConfig = `
-    resource "vsphere_datacenter" "testDC" {
-	    name = "testDC"
-    }`
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -36,14 +43,7 @@ func TestAccVSphereDatacenter_createOnRootFolder(t *testing.T) {
 
 // Create a datacenter on a subfolder
 func TestAccVSphereDatacenter_createOnSubfolder(t *testing.T) {
-	const resourceName = "vsphere_datacenter.testDC"
 	dcFolder := os.Getenv("VSPHERE_DC_FOLDER")
-
-	const testAccCheckVSphereDatacenterConfig = `
-    resource "vsphere_datacenter" "testDC" {
-	    name = "testDC"
-      folder = "%s"
-    }`
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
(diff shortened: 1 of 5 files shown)


vancluever:
Hey @jorgenunez - thanks for this - will review as soon as we can! :+1:

website/docs/r/datacenter.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+sidebar_current: "docs-vsphere-resource-datacenter"
+description: |-
+  Provides a VMware vSphere datacenter resource. This can be used as the primary container of inventory objects such as hosts and virtual machines.
+---
+
+# vsphere\_datacenter
+
+Provides a VMware vSphere datacenter resource. This can be used as the primary container of inventory objects such as hosts and virtual machines.
+
+No updates are possible.
```
vancluever: Can this sentence be fleshed out a bit? While the context might be clear to us as contributors and engineers it might not necessarily be to someone reading this document.

You could also move this to the argument reference and make it a warning by using the `~>` prefix. Something like:

```
## Argument Reference

~> **NOTE**: Datacenters cannot be changed once they are created. Modifying any of these attributes will force a new resource!
```

jorgenunez:
Hi @vancluever, thank you for the feedback. I will take a look at it asap.

vsphere/resource_vsphere_datacenter_test.go
```diff
@@ -0,0 +1,115 @@
...
+		CheckDestroy: testAccCheckVSphereDatacenterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(
+					testAccCheckVSphereDatacenterConfig,
+					name,
+					name,
+				),
+				Check: resource.ComposeTestCheckFunc(
+					waitForCreation,
```
jorgenunez: You are absolutely right, I missed this race condition. I have used ```WaitForState``` both for creation and deletion, I think it should be ok now.

website/docs/r/datacenter.html.markdown
jorgenunez: I added the note you mentioned. In the beginning I put it like that more than anything to show that as implemented here there can be no updates, but I agree that it wasn't clear enough. Saying that it can't be updated is a small lie though, as far as I know there is at least one case in which an update within vSphere to a datacenter would be possible, which would be moving the datacenter to a different folder. 

So far I haven't implemented update but will look into it. I would like to finish two other resources I'm working on (distributed virtual switches and distributed portgroups) and then come back to this if that's ok.

jorgenunez:
Hey @vancluever, I implemented your feedback. For the extra test I chose to add an extra env variable to get it working but I would rather use the ```vsphere_folder``` resource, I just didn't know exactly how, if you can point me to some good example I would appreciate it.

vancluever:
Hey @jorgenunez - thanks!

I think that this is good for now - `vsphere_folder` may only work for VMs at this time so the extra folder variable is fine for now.

Will review shortly!

vsphere/resource_vsphere_datacenter_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/find"
+	"golang.org/x/net/context"
+)
+
+// Create a datacenter on the root folder
+func TestAccVSphereDatacenter_createOnRootFolder(t *testing.T) {
+	const resourceName = "vsphere_datacenter.testDC"
+
+	const testAccCheckVSphereDatacenterConfig = `
```
vancluever: Can we move these constants to package scope? We don't really have many constants that have function scope.

vsphere/resource_vsphere_datacenter_test.go
```diff
@@ -0,0 +1,124 @@
...
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("[ERROR] No ID is set")
+		}
+
+		client := testAccProvider.Meta().(*govmomi.Client)
+		finder := find.NewFinder(client.Client, true)
+
+		path := rs.Primary.Attributes["name"]
+		if _, ok := rs.Primary.Attributes["folder"]; ok {
+			path = rs.Primary.Attributes["folder"] + path
```
vancluever: I think this is why `TestAccVSphereDatacenter_createOnSubfolder` is failing - the string needs to be concatenated with a `/` here.

jorgenunez: Yeah, this is the issue. I specified the folder for my own tests with a trailing ```/``` so it was passing for me. I added the ```/``` here now. It might be the case that someone specifies the folder with or without a trailing ```/```, but both work now. I will make it a nicer when I start to implement the update action.

vancluever:
Hey @jorgenunez - this LGTM now! I just made a small formatting change to the test fixtures so that they were properly aligned, as I didn't think it was big enough of a deal to kick back to you. Thank you so much for this PR and I am merging it now!

Issue #16, opened 2017-06-13
vSphere provider doesn't work correctly if secondary/alias addresses are added


hashibot:
_This issue was originally opened by @deasmi as hashicorp/terraform#6326. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I am using aliases/secondary IP addresses for some services which get added by puppet rather than configured but the vSphere template.

As such when I re-run terraform plan it detect the IP address and then think it needs to make a change.

I think this is a bug as these are not the primary addresses as can be seen here.

```
[deasmi@desktop:~/projects/terraform-zelotus] $ terraform plan zelotus.com/
Refreshing Terraform state prior to plan...

<.. CUT...>
vsphere_virtual_machine.orch1: Refreshing state... (ID: terraform/orch1)
<.. CUT ..>

The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed.

Note: You didn't specify an "-out" parameter to save this plan, so when
"apply" is called, Terraform can't guarantee this is what will execute.

~ vsphere_virtual_machine.orch1
    network_interface.0.ipv4_address:       "192.168.xx.50" => "192.168.xx.27"
    network_interface.0.ipv4_prefix_length: "32" => "24"


Plan: 0 to add, 1 to change, 0 to destroy.
[deasmi@desktop:~/projects/terraform-zelotus] $

```

On the server itself 

```
[root@orch1:~] # ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: ens192: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 00:50:56:9e:c5:85 brd ff:ff:ff:ff:ff:ff
    inet 192.168.xx.27/24 brd 192.168.xx.255 scope global ens192
       valid_lft forever preferred_lft forever
[...]

PR #128, opened 2017-08-11
Fix ip mapping from guest to terraform

### LINKED ISSUE #127: vpshere plugin returns the last defined ipadres from the IpStack instead of the first
### Terraform Version
Terraform v0.9.11
terraform-provider-vsphere - master

### Affected Resource(s)
- vsphere_virtual_machine

```hcl
resource "vsphere_virtual_machine" "acceptance-test" {
        name          = "test"
        vcpu          = 1
        memory        = 1024
        datacenter    = "DC0"
        cluster       = "CL0"
        resource_pool = "CL0/RP0"
        network_interface {
                label              = "VM network"
                ipv4_address       = "10.5.8.8"
                ipv4_prefix_length = "24"
                ipv4_gateway       = "10.5.8.1"
		ipv6_address       = "2001:XXXX::8"
		ipv6_prefix_length = "56"
		ipv6_gateway       = "2001:XXXX::1"

        }
        disk { ... }
}
```

### Expected Behavior
no changes in the VM when i run plan/apply for an existing VM with no changes

### Actual Behavior
terraform plan

case where ipv6 returns link-local
~ vsphere_virtual_machine.test
    network_interface.0.ipv6_address:       "fe80::250:56ff:febf:7f29" => "2001:XXXX::8"
    network_interface.0.ipv6_prefix_length: "64" => "56"

case where the last ip is the floating ip between 2 keepalived instances
~ vsphere_virtual_machine.test
    network_interface.0.ipv4_address:       "10.5.8.100" => "10.5.8.8"
    network_interface.0.ipv4_prefix_length: "32" => "24"

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`

mjrider:
- only accept the first ip adres per family instead of the last one
  This is a problem which machines with floating ipaddresses

- for ipv6, reject link-local ipadresses, those are auto-generated

fixes #127


vancluever:
Thanks @mjrider  - as mentioned will look at this soon :+1:

mjrider:
needs some more work:
- ipv6_gateway: same filtering as for ipv6_address, filter away local scope addresses
- conflicts with #129 ebcause of changes on the same lines

vancluever:
Hey @mjrider - so a weird side-effect here: Using a config like this for network:

```
  network_interface {
    label              = "public"
    ipv4_address       = "172.16.1.150"
    ipv4_prefix_length = 24
    ipv4_gateway       = "172.16.1.1"
  }
```

With an IP address added local to the machine (ie: `ip addr add 172.16.1.151/24 dev eth0`) leads to the following in the state:

```
                            "network_interface.#": "1",
                            "network_interface.0.adapter_type": "",
                            "network_interface.0.ip_address": "",
                            "network_interface.0.ipv4_address": "172.16.1.150",
                            "network_interface.0.ipv4_gateway": "172.16.1.1",
                            "network_interface.0.ipv4_prefix_length": "24",
                            "network_interface.0.ipv6_address": "172.16.1.151",
                            "network_interface.0.ipv6_gateway": "",
                            "network_interface.0.ipv6_prefix_length": "32",
                            "network_interface.0.key": "4000",
                            "network_interface.0.label": "public",
                            "network_interface.0.mac_address": "00:50:56:94:22:d5",
                            "network_interface.0.subnet_mask": "",
```

Note how `172.16.1.151` wound up as the `ipv6_address`.

Looks like this might be because how you are checking for an IPv6 address with net.IP - `To16` will actually convert an IPv4 address to an IPv6 "address":

```
[DEBUG] p.To16 == net.IP{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xff, 0xff, 0xac, 0x10, 0x1, 0x97}
```

What you can do, however, is check for `nil` on `To4`. Check https://play.golang.org/p/cKUjXkswjy for an example.

mjrider:
this should fix #16

vancluever:
Hey @mjrider - found another one - this is more just a spurious diff issue, stemming from the normalization of IPv6 addresses:

```
  ~ vsphere_virtual_machine.foo
      network_interface.0.ipv6_address: "fd00::1" => "fd00::0001"
```

Can we some code in to make sure this is skipped in TF (ie: through maybe a [`DiffSuppressFunc`](https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L59) or a [`StateFunc`](https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L111)), whatever one you think might be more appropriate here?

mjrider:
@vancluever i added a  DiffSuppressFunc for ipv4 and ipv6 address and gateway

vancluever:
Hey @mjrider - this fixes the normalization issues on IPv6 addresses - thanks!

Looks like, however, we are still having issues with IPv6 addresses - I added another IPv6 address to the interface (`fd00::2/112`) and am having the same issue now:

```
  ~ vsphere_virtual_machine.foo
      network_interface.0.ipv6_address: "fd00::2" => "fd00::0001"
```

I honestly think that what we will ultimately need to do is separate the guest network info from the network information supplied in Terraform, which is ultimately a configuration specification of sorts (I would have to dig into the API to know 100%). Since re-configuration isn't allowed on regular vSphere anyway, having this set to a `ForceNew` anyway should not necessarily be out of the realm of possibility, with the guest network information being saved to a purely computed hierarchy that does not cause state drift.

This is a longer term thing though and I while IPv6 is important I'd imagine that there are a lot more people with this issue using IPv4, so I'm going to merge this one now. Maybe if you wanted to pick up something, you could see how we could check config to determine if we should even write an IP address to state :)

Thanks for the contribution!

PR #129, opened 2017-08-11
Discovery network interfaces from pci devices instead from Guest information

mjrider:
This change enumerates the pci devices to list the netwerkinterfaces.
After that , the guest information is used to provide the ipstack data.

What does this fix:
- vsphere guest info sometimes misses the network label
- network interfaces without ip are no longer skipped


vancluever:
Hey @mjrider - thanks for this!

Just a question - is this complete or are you still working on it? If it's done can you remove the WIP heading from the subject so that I can schedule the review and testing accordingly?

Thanks!

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1057,32 +1062,54 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("Invalid disks to set: %#v", disks)
 	}
 
+	// start
 	networkInterfaces := make([]map[string]interface{}, 0)
+
+	deviceList, err := vm.Device(context.TODO())
```
vancluever: @mjrider just a note here - am I right in assuming that `err` is not being caught here?

Also = I notice this device list is being sourced off of the `vm` MO instead of the `mvm` MO, although the device being logged in debug is being sourced off of `mvm`. Is this an error or by design?

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1057,32 +1062,54 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
+	deviceList = deviceList.SelectByType((*types.VirtualEthernetCard)(nil))
+	log.Printf("Device list %+v", mvm.Config.Hardware.Device)
```
vancluever: Can we get the `[DEBUG]` prefix here?

Looks like there are some other entries that need this as well - can we get those there as well?

vancluever:
LGTM @mjrider - thanks for the fixes! Merging this now and will be in the release next week :+1:

Thanks again for the great work here!

PR #136, opened 2017-08-22
Host networking resources: vSwitch and port group management

vancluever:
This PR introduces two new resources:

 * `vsphere_host_virtual_switch`: Manage a virtual switch on an ESXi host
 * `vsphere_host_port_group`: Manage a port group on an ESXi host

These resources demonstrate the first of a larger effort to get essential network management capabilities into the provider. Using these resources, one can now set up virtual networks on ESXi hosts, which can be used in virtual machines after that.

From here, we can look at setting up:

 * VMkernel NICs for additional host configuration
 * Distributed virtual switches and port groups

All of these are probably up on deck next, but definitely the distributed virtual switches and port groups will have high value, as this is more than likely what people are really using when setting up new networks to connect virtual machines to. This is an important first step in getting there though, as we can take the understanding of host networking in the vSphere API and now apply it to the more complex VDS resources.

Feedback is welcome and looking forward to getting this merged!


vancluever:
Hey @radeksimko - will do this. My bad - I think I got a little carried away here and ended up with a bit of a big changeset. Will do that and re-submit.

PR #138, opened 2017-08-22
New resource: vsphere_host_virtual_switch

vancluever:
This commit adds a new resource, `vsphere_host_virtual_switch`, which can be used to manage a virtual switch on an ESXi host.
    
Docs included, and tests have been run against both ESXi and vCenter.

### DIFF
--- a/vsphere/host_network_policy_structure.go
+++ b/vsphere/host_network_policy_structure.go
@@ -1,0 +1,286 @@
+package vsphere
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+const (
+	hostNetworkPolicyNicTeamingPolicyModeLoadbalanceIP     = "loadbalance_ip"
+	hostNetworkPolicyNicTeamingPolicyModeLoadbalanceSrcMac = "loadbalance_srcmac"
+	hostNetworkPolicyNicTeamingPolicyModeLoadbalanceSrcID  = "loadbalance_srcid"
+	hostNetworkPolicyNicTeamingPolicyModeFailoverExplicit  = "failover_explicit"
+)
+
+var hostNetworkPolicyNicTeamingPolicyAllowedValues = []string{
+	hostNetworkPolicyNicTeamingPolicyModeLoadbalanceIP,
+	hostNetworkPolicyNicTeamingPolicyModeLoadbalanceSrcMac,
+	hostNetworkPolicyNicTeamingPolicyModeLoadbalanceSrcID,
+	hostNetworkPolicyNicTeamingPolicyModeFailoverExplicit,
+}
+
+// schemaHostNetworkPolicy returns schema items for resources that need to work
+// with a HostNetworkPolicy, such as virtual switches and port groups.
+func schemaHostNetworkPolicy() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		// HostNicTeamingPolicy/HostNicFailureCriteria
+		"check_beacon": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Enable beacon probing. Requires that the vSwitch has been configured to use a beacon. If disabled, link status is used only.",
+		},
+
+		// HostNicTeamingPolicy/HostNicOrderPolicy
+		"active_nics": &schema.Schema{
+			Type:        schema.TypeList,
+			Description: "List of active network adapters used for load balancing.",
+			Elem:        &schema.Schema{Type: schema.TypeString},
+		},
+		"standby_nics": &schema.Schema{
+			Type:        schema.TypeList,
+			Description: "List of standby network adapters used for failover.",
+			Elem:        &schema.Schema{Type: schema.TypeString},
+		},
+
+		// HostNicTeamingPolicy
+		"teaming_policy": &schema.Schema{
+			Type:         schema.TypeString,
+			Optional:     true,
+			Description:  "The network adapter teaming policy. Can be one of loadbalance_ip, loadbalance_srcmac, loadbalance_srcid, or failover_explicit.",
+			ValidateFunc: validation.StringInSlice(hostNetworkPolicyNicTeamingPolicyAllowedValues, false),
+		},
+		"notify_switches": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "If true, the teaming policy will notify the broadcast network of a NIC failover, triggering cache updates.",
+		},
+		"failback": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "If true, the teaming policy will re-activate failed interfaces higher in precedence when they come back up.",
+		},
+
+		// HostNetworkSecurityPolicy
+		"allow_promiscuous": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Enable promiscuous mode on the network. This flag indicates whether or not all traffic is seen on a given port.",
+		},
+		"allow_forged_transmits": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Controls whether or not the virtual network adapter is allowed to send network traffic with a different MAC address than that of its own.",
+		},
+		"allow_mac_changes": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Controls whether or not the Media Access Control (MAC) address can be changed.",
+		},
+
+		// HostNetworkTrafficShapingPolicy
+		"shaping_average_bandwidth": &schema.Schema{
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "The average bandwidth in bits per second if shaping is enabled on the port.",
+		},
+		"shaping_burst_size": &schema.Schema{
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "The maximum burst size allowed in bytes if shaping is enabled on the port.",
+		},
+		"shaping_enabled": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "True if the traffic shaper is enabled on the port.",
+		},
+		"shaping_peak_bandwidth": &schema.Schema{
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "The peak bandwidth during bursts in bits per second if traffic shaping is enabled on the port.",
+		},
+	}
+}
+
+// expandHostNicFailureCriteria reads certain ResourceData keys and returns a
+// HostNicFailureCriteria.
+func expandHostNicFailureCriteria(d *schema.ResourceData) *types.HostNicFailureCriteria {
+	obj := &types.HostNicFailureCriteria{}
+
+	if v, ok := d.GetOkExists("check_beacon"); ok {
+		obj.CheckBeacon = &([]bool{v.(bool)}[0])
+	}
+
+	// These fields are deprecated and are set only to make things work. They are
+	// not exposed to Terraform.
+	obj.CheckSpeed = "minimum"
+	obj.Speed = 10
+	obj.CheckDuplex = &([]bool{false}[0])
+	obj.FullDuplex = &([]bool{false}[0])
+	obj.CheckErrorPercent = &([]bool{false}[0])
+	obj.Percentage = 0
+
+	return obj
+}
+
+// flattenHostNicFailureCriteria reads various fields from a
+// HostNicFailureCriteria into the passed in ResourceData.
+func flattenHostNicFailureCriteria(d *schema.ResourceData, obj *types.HostNicFailureCriteria) error {
+	if obj.CheckBeacon != nil {
+		d.Set("check_beacon", obj.CheckBeacon)
+	}
+	return nil
+}
+
+// expandHostNicOrderPolicy reads certain ResourceData keys and returns a
+// HostNicOrderPolicy.
+func expandHostNicOrderPolicy(d *schema.ResourceData) *types.HostNicOrderPolicy {
+	obj := &types.HostNicOrderPolicy{}
+	activeNics, activeOk := d.GetOkExists("active_nics")
+	standbyNics, standbyOk := d.GetOkExists("standby_nics")
+	if !activeOk && !standbyOk {
+		return nil
+	}
+	obj.ActiveNic = sliceInterfacesToStrings(activeNics.([]interface{}))
+	obj.StandbyNic = sliceInterfacesToStrings(standbyNics.([]interface{}))
+	return obj
+}
+
+// flattenHostNicOrderPolicy reads various fields from a HostNicOrderPolicy
+// into the passed in ResourceData.
+func flattenHostNicOrderPolicy(d *schema.ResourceData, obj *types.HostNicOrderPolicy) error {
+	if obj == nil {
+		return nil
+	}
+	if err := d.Set("active_nics", sliceStringsToInterfaces(obj.ActiveNic)); err != nil {
+		return err
+	}
+	if err := d.Set("standby_nics", sliceStringsToInterfaces(obj.StandbyNic)); err != nil {
+		return err
+	}
+	return nil
+}
+
+// expandHostNicTeamingPolicy reads certain ResourceData keys and returns a
+// HostNicTeamingPolicy.
+func expandHostNicTeamingPolicy(d *schema.ResourceData) *types.HostNicTeamingPolicy {
+	obj := &types.HostNicTeamingPolicy{
+		Policy: d.Get("teaming_policy").(string),
+	}
+	if v, ok := d.GetOkExists("failback"); ok {
+		obj.RollingOrder = &([]bool{!v.(bool)}[0])
+	}
+	if v, ok := d.GetOkExists("notify_switches"); ok {
+		obj.NotifySwitches = &([]bool{v.(bool)}[0])
+	}
+	obj.FailureCriteria = expandHostNicFailureCriteria(d)
+	obj.NicOrder = expandHostNicOrderPolicy(d)
+
+	// These fields are deprecated and are set only to make things work. They are
+	// not exposed to Terraform.
+	obj.ReversePolicy = &([]bool{true}[0])
+
+	return obj
+}
+
+// flattenHostNicTeamingPolicy reads various fields from a HostNicTeamingPolicy
+// into the passed in ResourceData.
+func flattenHostNicTeamingPolicy(d *schema.ResourceData, obj *types.HostNicTeamingPolicy) error {
+	if obj.RollingOrder != nil {
+		d.Set("failback", !*obj.RollingOrder)
+	}
+	if obj.NotifySwitches != nil {
+		d.Set("notify_switches", obj.NotifySwitches)
+	}
+	d.Set("teaming_policy", obj.Policy)
+	if err := flattenHostNicFailureCriteria(d, obj.FailureCriteria); err != nil {
+		return err
+	}
+	if err := flattenHostNicOrderPolicy(d, obj.NicOrder); err != nil {
+		return err
+	}
+	return nil
+}
+
+// expandHostNetworkSecurityPolicy reads certain ResourceData keys and returns
+// a HostNetworkSecurityPolicy.
+func expandHostNetworkSecurityPolicy(d *schema.ResourceData) *types.HostNetworkSecurityPolicy {
+	obj := &types.HostNetworkSecurityPolicy{}
+	if v, ok := d.GetOkExists("allow_promiscuous"); ok {
+		obj.AllowPromiscuous = &([]bool{v.(bool)}[0])
+	}
+	if v, ok := d.GetOkExists("allow_forged_transmits"); ok {
+		obj.ForgedTransmits = &([]bool{v.(bool)}[0])
+	}
+	if v, ok := d.GetOkExists("allow_mac_changes"); ok {
+		obj.MacChanges = &([]bool{v.(bool)}[0])
+	}
+	return obj
+}
+
+// flattenHostNetworkSecurityPolicy reads various fields from a
+// HostNetworkSecurityPolicy into the passed in ResourceData.
+func flattenHostNetworkSecurityPolicy(d *schema.ResourceData, obj *types.HostNetworkSecurityPolicy) error {
+	if obj.AllowPromiscuous != nil {
+		d.Set("allow_promiscuous", *obj.AllowPromiscuous)
+	}
+	if obj.ForgedTransmits != nil {
+		d.Set("allow_forged_transmits", *obj.ForgedTransmits)
+	}
+	if obj.MacChanges != nil {
+		d.Set("allow_mac_changes", *obj.MacChanges)
+	}
+	return nil
+}
+
+// expandHostNetworkTrafficShapingPolicy reads certain ResourceData keys and
+// returns a HostNetworkTrafficShapingPolicy.
+func expandHostNetworkTrafficShapingPolicy(d *schema.ResourceData) *types.HostNetworkTrafficShapingPolicy {
+	obj := &types.HostNetworkTrafficShapingPolicy{
+		AverageBandwidth: int64(d.Get("shaping_average_bandwidth").(int)),
+		BurstSize:        int64(d.Get("shaping_burst_size").(int)),
+		PeakBandwidth:    int64(d.Get("shaping_peak_bandwidth").(int)),
+	}
+	if v, ok := d.GetOkExists("shaping_enabled"); ok {
+		obj.Enabled = &([]bool{v.(bool)}[0])
+	}
+	return obj
+}
+
+// flattenHostNetworkTrafficShapingPolicy reads various fields from a
+// HostNetworkTrafficShapingPolicy into the passed in ResourceData.
+func flattenHostNetworkTrafficShapingPolicy(d *schema.ResourceData, obj *types.HostNetworkTrafficShapingPolicy) error {
+	if obj.Enabled != nil {
+		d.Set("shaping_enabled", *obj.Enabled)
+	}
+	d.Set("shaping_average_bandwidth", obj.AverageBandwidth)
+	d.Set("shaping_burst_size", obj.BurstSize)
+	d.Set("shaping_peak_bandwidth", obj.PeakBandwidth)
+	return nil
+}
+
+// expandHostNetworkPolicy reads certain ResourceData keys and returns a
+// HostNetworkPolicy.
+func expandHostNetworkPolicy(d *schema.ResourceData) *types.HostNetworkPolicy {
+	obj := &types.HostNetworkPolicy{
+		Security:      expandHostNetworkSecurityPolicy(d),
+		NicTeaming:    expandHostNicTeamingPolicy(d),
+		ShapingPolicy: expandHostNetworkTrafficShapingPolicy(d),
+	}
+	return obj
+}
+
+// flattenHostNetworkPolicy reads various fields from a HostNetworkPolicy into
+// the passed in ResourceData.
+func flattenHostNetworkPolicy(d *schema.ResourceData, obj *types.HostNetworkPolicy) error {
+	if err := flattenHostNetworkSecurityPolicy(d, obj.Security); err != nil {
+		return err
+	}
+	if err := flattenHostNicTeamingPolicy(d, obj.NicTeaming); err != nil {
+		return err
+	}
+	if err := flattenHostNetworkTrafficShapingPolicy(d, obj.ShapingPolicy); err != nil {
+		return err
+	}
+	return nil
+}
--- a/vsphere/host_network_policy_structure.go
+++ b/vsphere/host_network_policy_structure.go
@@ -187,7 +187,8 @@ func expandHostNicTeamingPolicy(d *schema.ResourceData) *types.HostNicTeamingPol
 // into the passed in ResourceData.
 func flattenHostNicTeamingPolicy(d *schema.ResourceData, obj *types.HostNicTeamingPolicy) error {
 	if obj.RollingOrder != nil {
-		d.Set("failback", !*obj.RollingOrder)
+		v := *obj.RollingOrder
+		d.Set("failback", !v)
 	}
 	if obj.NotifySwitches != nil {
 		d.Set("notify_switches", obj.NotifySwitches)
--- a/vsphere/host_network_policy_structure.go
+++ b/vsphere/host_network_policy_structure.go
@@ -108,16 +108,16 @@ func expandHostNicFailureCriteria(d *schema.ResourceData) *types.HostNicFailureC
 	obj := &types.HostNicFailureCriteria{}
 
 	if v, ok := d.GetOkExists("check_beacon"); ok {
-		obj.CheckBeacon = &([]bool{v.(bool)}[0])
+		obj.CheckBeacon = boolPtr(v.(bool))
 	}
 
 	// These fields are deprecated and are set only to make things work. They are
 	// not exposed to Terraform.
 	obj.CheckSpeed = "minimum"
 	obj.Speed = 10
-	obj.CheckDuplex = &([]bool{false}[0])
-	obj.FullDuplex = &([]bool{false}[0])
-	obj.CheckErrorPercent = &([]bool{false}[0])
+	obj.CheckDuplex = boolPtr(false)
+	obj.FullDuplex = boolPtr(false)
+	obj.CheckErrorPercent = boolPtr(false)
 	obj.Percentage = 0
 
 	return obj
@@ -168,17 +168,17 @@ func expandHostNicTeamingPolicy(d *schema.ResourceData) *types.HostNicTeamingPol
 		Policy: d.Get("teaming_policy").(string),
 	}
 	if v, ok := d.GetOkExists("failback"); ok {
-		obj.RollingOrder = &([]bool{!v.(bool)}[0])
+		obj.RollingOrder = boolPtr(!v.(bool))
 	}
 	if v, ok := d.GetOkExists("notify_switches"); ok {
-		obj.NotifySwitches = &([]bool{v.(bool)}[0])
+		obj.NotifySwitches = boolPtr(v.(bool))
 	}
 	obj.FailureCriteria = expandHostNicFailureCriteria(d)
 	obj.NicOrder = expandHostNicOrderPolicy(d)
 
 	// These fields are deprecated and are set only to make things work. They are
 	// not exposed to Terraform.
-	obj.ReversePolicy = &([]bool{true}[0])
+	obj.ReversePolicy = boolPtr(true)
 
 	return obj
 }
@@ -207,13 +207,13 @@ func flattenHostNicTeamingPolicy(d *schema.ResourceData, obj *types.HostNicTeami
 func expandHostNetworkSecurityPolicy(d *schema.ResourceData) *types.HostNetworkSecurityPolicy {
 	obj := &types.HostNetworkSecurityPolicy{}
 	if v, ok := d.GetOkExists("allow_promiscuous"); ok {
-		obj.AllowPromiscuous = &([]bool{v.(bool)}[0])
+		obj.AllowPromiscuous = boolPtr(v.(bool))
 	}
 	if v, ok := d.GetOkExists("allow_forged_transmits"); ok {
-		obj.ForgedTransmits = &([]bool{v.(bool)}[0])
+		obj.ForgedTransmits = boolPtr(v.(bool))
 	}
 	if v, ok := d.GetOkExists("allow_mac_changes"); ok {
-		obj.MacChanges = &([]bool{v.(bool)}[0])
+		obj.MacChanges = boolPtr(v.(bool))
 	}
 	return obj
 }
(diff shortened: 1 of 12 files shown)


vancluever:
Done!

I will hold off on messing with #139 until this one is merged to reduce the changeset over there too. :)

vsphere/host_network_policy_structure.go
```diff
@@ -0,0 +1,273 @@
...
+			Description: "If true, the teaming policy will re-activate failed interfaces higher in precedence when they come back up.",
+		},
+
+		// HostNetworkSecurityPolicy
+		"allow_promiscuous": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Enable promiscuous mode on the network. This flag indicates whether or not all traffic is seen on a given port.",
+		},
+		"forged_transmits": &schema.Schema{
```
tombuildsstuff: I think this might be better as `enable_forced_transmits`?

vsphere/host_network_policy_structure.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// flattenHostNicTeamingPolicy reads various fields from a HostNicTeamingPolicy
+// into the passed in ResourceData.
+func flattenHostNicTeamingPolicy(d *schema.ResourceData, obj *types.HostNicTeamingPolicy) error {
+	if obj.RollingOrder != nil {
+		d.Set("failback", !*obj.RollingOrder)
+	}
+	if obj.NotifySwitches != nil {
+		d.Set("notify_switches", *obj.NotifySwitches)
```
tombuildsstuff: is the `*` needed here - I think d.Set will handle that for us?

vsphere/host_network_policy_structure.go
```diff
@@ -0,0 +1,273 @@
...
+
+// expandHostNetworkTrafficShapingPolicy reads certain ResourceData keys and
+// returns a HostNetworkTrafficShapingPolicy.
+func expandHostNetworkTrafficShapingPolicy(d *schema.ResourceData) *types.HostNetworkTrafficShapingPolicy {
+	obj := &types.HostNetworkTrafficShapingPolicy{
+		AverageBandwidth: int64(d.Get("shaping_average_bandwidth").(int)),
+		BurstSize:        int64(d.Get("shaping_burst_size").(int)),
+		PeakBandwidth:    int64(d.Get("shaping_peak_bandwidth").(int)),
+	}
+	if v, ok := d.GetOkExists("shaping_enabled"); ok {
```
tombuildsstuff: should this be in a child-object?

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+// schemaHostVirtualSwitchBondBridge returns schema items for resources that
+// need to work with a HostVirtualSwitchBondBridge, such as virtual switches.
+func schemaHostVirtualSwitchBondBridge() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		// HostVirtualSwitchBeaconConfig
+		"beacon_interval": &schema.Schema{
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "Determines how often, in seconds, a beacon should be sent to probe for the validity of a link.",
+			Default:     1,
```
tombuildsstuff: does this want some validation for the supported ranges?

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+
+		// LinkDiscoveryProtocolConfig
+		"link_discovery_operation": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Whether to advertise or listen for link discovery. Valid values are advertise, both, listen, and none.",
+			Default:     "listen",
```
tombuildsstuff: does this want a validation func for these values?

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+		"link_discovery_protocol": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "The discovery protocol type. Valid values are cdp and lldp.",
```
tombuildsstuff: can we add a `ValidateFunc` here?

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+	d.Set("beacon_interval", obj.Interval)
+	return nil
+}
+
+// expandLinkDiscoveryProtocolConfig reads certain ResourceData keys and
+// returns a LinkDiscoveryProtocolConfig.
+func expandLinkDiscoveryProtocolConfig(d *schema.ResourceData) *types.LinkDiscoveryProtocolConfig {
+	obj := &types.LinkDiscoveryProtocolConfig{
+		Operation: d.Get("link_discovery_operation").(string),
+		Protocol:  d.Get("link_discovery_protocol").(string),
```
tombuildsstuff: would this make more sense as a sub-object?

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+// schemaHostVirtualSwitchSpec returns schema items for resources that need to
+// work with a HostVirtualSwitchSpec, such as virtual switches.
+func schemaHostVirtualSwitchSpec() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		// HostVirtualSwitchSpec
+		"mtu": &schema.Schema{
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Description: "The maximum transmission unit (MTU) of the virtual switch in bytes.",
+			Default:     1500,
```
tombuildsstuff: should we be applying validation here?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	}
+	mergeSchema(s, schemaHostVirtualSwitchSpec())
+
+	// Transform any necessary fields in the schema that need to be updated
+	// specifically for this resource.
+	s["active_nics"].Required = true
+	s["standby_nics"].Required = true
+
+	s["teaming_policy"].Default = "loadbalance_srcid"
```
tombuildsstuff: 🤔 this should come from an enum if possible

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+func resourceVSphereHostVirtualSwitchCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*govmomi.Client)
+	host := d.Get("host").(string)
+	datacenter := d.Get("datacenter").(string)
+	ns, err := hostNetworkSystemFromName(client, host, datacenter)
+	if err != nil {
+		return fmt.Errorf("error loading network system: %s", err)
+	}
+
+	timeout := time.Duration(float64(d.Timeout(schema.TimeoutCreate)) * 0.8)
```
tombuildsstuff: as discussed on Slack - there's no guarantee a cancel would complete in `0.2 * timeout`?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+	if err := ns.AddVirtualSwitch(ctx, d.Get("name").(string), spec); err != nil {
+		return fmt.Errorf("error adding host vSwitch: %s", err)
+	}
+
+	d.SetId(d.Get("name").(string))
+	return resourceVSphereHostVirtualSwitchRead(d, meta)
+}
+
+func resourceVSphereHostVirtualSwitchRead(d *schema.ResourceData, meta interface{}) error {
+	timeout := time.Duration(float64(d.Timeout(schema.TimeoutCreate)) * 0.8)
```
tombuildsstuff: as discussed on Slack - there's no guarantee a cancel would complete in `0.2 * timeout`?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+func resourceVSphereHostVirtualSwitchUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*govmomi.Client)
+	host := d.Get("host").(string)
+	datacenter := d.Get("datacenter").(string)
+	ns, err := hostNetworkSystemFromName(client, host, datacenter)
+	if err != nil {
+		return fmt.Errorf("error loading network system: %s", err)
+	}
+
+	timeout := time.Duration(float64(d.Timeout(schema.TimeoutCreate)) * 0.8)
```
tombuildsstuff: as discussed on Slack - there's no guarantee a cancel would complete in `0.2 * timeout`?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func resourceVSphereHostVirtualSwitchDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*govmomi.Client)
+	ns, err := hostNetworkSystemFromName(client, d.Get("host").(string), d.Get("datacenter").(string))
+	if err != nil {
+		return fmt.Errorf("error loading network system: %s", err)
+	}
+
+	timeout := time.Duration(float64(d.Timeout(schema.TimeoutCreate)) * 0.8)
```
tombuildsstuff: as discussed on Slack - there's no guarantee a cancel would complete in `0.2 * timeout`?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	if err := ns.RemoveVirtualSwitch(ctx, d.Id()); err != nil {
+		return fmt.Errorf("error deleting host vSwitch: %s", err)
+	}
+
+	d.SetId("")
```
tombuildsstuff: I think this is done automatically once the delete returns nil?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+
+	timeout := time.Duration(float64(d.Timeout(schema.TimeoutCreate)) * 0.8)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	spec := expandHostVirtualSwitchSpec(d)
+	if err := ns.AddVirtualSwitch(ctx, d.Get("name").(string), spec); err != nil {
+		return fmt.Errorf("error adding host vSwitch: %s", err)
+	}
+
+	d.SetId(d.Get("name").(string))
```
tombuildsstuff: _minor_ given this is accessed twice, worth pulling this out into a variable?

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,254 @@
...
+		rs, ok := s.RootModule().Resources["vsphere_host_virtual_switch.switch"]
+		if !ok {
+			return errors.New("vsphere_host_virtual_switch.switch not found in state")
+		}
+
+		client := testAccProvider.Meta().(*govmomi.Client)
+		id := rs.Primary.ID
+		host := os.Getenv("VSPHERE_ESXI_HOST")
+		datacenter := os.Getenv("VSPHERE_DATACENTER")
+		timeout := time.Minute * 5
```
tombuildsstuff: worth making a `TestClientWrapper` for these variables?

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+variable "host_nic0" {
+  type    = "string"
+  default = "%s"
+}
+
+variable "host_nic1" {
+  type    = "string"
+  default = "%s"
+}
```
tombuildsstuff: _minor_ we can remove the `type` from these

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,254 @@
...
+  active_nics  = ["${var.host_nic0}"]
+  standby_nics = []
+}
+`, os.Getenv("VSPHERE_ESXI_HOST"), os.Getenv("VSPHERE_DATACENTER"), os.Getenv("VSPHERE_HOST_NIC0"))
+}
+
+func testAccResourceVSphereHostVirtualSwitchConfigStandbyLink() string {
+	return fmt.Sprintf(`
+variable "esxi_host" {
+  type    = "string"
```
tombuildsstuff: _minor_ we can remove the types here

website/docs/r/host_virtual_switch.html.markdown
```diff
@@ -0,0 +1,156 @@
...
+
+The `vsphere_host_virtual_switch` resource can be used to manage vSphere
+standard switches on an ESXi host. These switches can be used as a backing for
+standard port groups, which can be managed by the
+[`vsphere_host_port_group`][host-port-group] resource.
+
+For an overview on vSphere networking concepts, see [this page][ref-vsphere-net-concepts].
+
+[host-port-group]: /docs/providers/vsphere/r/host_port_group.html
+[ref-vsphere-net-concepts]: https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.networking.doc/GUID-2B11DBB8-CB3C-4AFF-8885-EFEA0FC562F4.html
```
tombuildsstuff: not seen this syntax before, but I'm assuming this renders correctly?

vsphere/host_network_policy_structure.go
vancluever: Hey @tombuildsstuff, I have been trying to avoid unnecessary nesting whenever possible. I have lost too many dev hours on trying to work around the issues of dealing with nested config stuff in TF - the inheritance model of this resource in conjunction with `vsphere_host_port_group` also makes things even more complicated. So going forward I am only reserving using lists and sets for things that need it because they need to deal with multiple values and that's it. A virtual switch or port group can only have one policy, of which traffic shaping options are a part of, so this does not need to be nested here.

vsphere/host_virtual_switch_structure.go
vancluever: Hey @tombuildsstuff, as mentioned [above](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r135106704) I'd like to stay away from nested values there's better first-class support for general nested config modeling in TF. This does not need a list or set either to support multiple values so I'm just leaving it flat.

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+		Protocol:  d.Get("link_discovery_protocol").(string),
+	}
+	return obj
+}
+
+// flattenLinkDiscoveryProtocolConfig reads various fields from a
+// LinkDiscoveryProtocolConfig into the passed in ResourceData.
+func flattenLinkDiscoveryProtocolConfig(d *schema.ResourceData, obj *types.LinkDiscoveryProtocolConfig) error {
+	d.Set("link_discovery_operation", obj.Operation)
+	d.Set("link_discovery_protocol", obj.Protocol)
```
vancluever: Again the issues with nested values - [original response](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r135106704).

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+// expandHostVirtualSwitchBondBridge reads certain ResourceData keys and
+// returns a HostVirtualSwitchBondBridge.
+func expandHostVirtualSwitchBondBridge(d *schema.ResourceData) *types.HostVirtualSwitchBondBridge {
+	obj := &types.HostVirtualSwitchBondBridge{
+		NicDevice: sliceInterfacesToStrings(d.Get("network_adapters").([]interface{})),
```
vancluever: I'm trying to think of what might be good - `network_adapter_names` seems a little too redundant - they are not device IDs, but basically the friendly names of the NICs (`ie: vmnic0`). I personally think that documentation should be able to shore this up okay enough - but what are your thoughts?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+	s := map[string]*schema.Schema{
+		"name": &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "The name of the virtual switch.",
+			Required:    true,
+			ForceNew:    true,
+		},
+		"host": &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "The name of the host to put this virtual switch on. This is ignored if connecting directly to ESXi, but required if not.",
```
vancluever: The problem here is that our provider targets ESXi and vCenter - so we need to accommodate both cases. With ESXi this option is superfluous because there is only one host and datacenter the provider can target - the host that the provider is connected to, and the "default" datacenter.

Hence, I would rather have an error trigger when someone tries to use this with vCenter without these options specified, versus have someone enter dummy data when working with ESXi:

Good: 

```hcl
resource "vsphere_host_virtual_switch" "esxiVSwitch0" {
  name = "esxiVSwitch0"
}
```

Bad:

```hcl
resource "vsphere_host_virtual_switch" "esxiVSwitch0" {
  name       = "esxiVSwitch0"
  host       = "unused"
  datacenter = "unused"
}
```

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,135 @@
...
+			Optional:    true,
+			ForceNew:    true,
+		},
+		"datacenter": &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "The path to the datacenter the host is located in. This is ignored if connecting directly to ESXi. If not specified on vCenter, the default datacenter is used.",
```
vancluever: This is tied to the [rationale for the host option](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r135109717).

website/docs/r/host_virtual_switch.html.markdown
vancluever: [It does](https://daringfireball.net/projects/markdown/syntax#link)!

vsphere/host_virtual_switch_structure.go
vancluever: Unfortunately vSphere doesn't really publish an allowable range, so I don't necessarily want to guess... but I have added a validation for a negative value.

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: Hmm, I thought that it's been generally recommended to use explicit variable type declaration in TF code ever since functionality for it was released in 0.6.10. Was I mistaken in that? I can change it, although it will take some getting used to for me writing my configs with some variables with defaults being implicit and the ones without defaults having `types`.

Like this just seems a little inconsistent to me:

```
variable "has_default" {
  default = "implied"
}

variable "no_default" {
  type = "string"
}
```

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: Thoughts [above](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r135161820) on this one.

vancluever:
Hey @tombuildsstuff! I think I addressed/commented on all of these now. Can you give it another look?

vsphere/resource_vsphere_host_virtual_switch.go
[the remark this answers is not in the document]
vancluever: Hey @tombuildsstuff, just thought I'd mention that I'm refactoring this to use data sources after all :) #144 and #146 aim to supply datacenter and host data sources that can be used to load in MOIDs for datacenters and host systems. The network resources will now take a host ID as a required field and the ambiguity of defaults vs vcenter vs ESXi has been offloaded to the data sources.

vancluever:
Refactored to use host MoRef from `vsphere_host` data source, which should help clean up some of the ambiguity with how this used to handle default host/datacenter stuff (this logic has now been passed off to the data sources).

Also, we now generate a bit more of a unique style ID, based off of a custom prefix (`tf-HostVirtualSwitch`), the HostSystem's MoRef ID, and the name of the virtual switch. The reason I think this is necessary is because a host virtual switch lacks any sort of real uniquely identifying criteria - both a virtual switch name and "key" (an attribute that is derived from the name of the virtual switch) are both subject to change - a name can be changed (although we do not support it in TF) and the key changes with the name. Even if they don't change, name or key alone are not enough to identify a specific virtual switch as there needs to be something to locate the host it is set up on. Hence the custom ID.

The refactor had the side effect of just squashing most of these commits - the other outstanding commits here are from the `vsphere_host` data source and v0.10.3 vendor PRs which are currently open.

These changes will be added to the `vsphere_host_port_group` resource as well.

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func testAccResourceVSphereHostVirtualSwitchPreCheck(t *testing.T) {
+	if os.Getenv("VSPHERE_HOST_NIC0") == "" {
+		t.Skip("set VSPHERE_HOST_NIC0 to run vsphere_host_virtual_switch acceptance tests")
+	}
+	if os.Getenv("VSPHERE_HOST_NIC1") == "" {
+		t.Skip("set VSPHERE_HOST_NIC1 to run vsphere_host_virtual_switch acceptance tests")
+	}
+	if os.Getenv("VSPHERE_ESXI_HOST") == "" {
```
tombuildsstuff: should this not be being caught in the test provider initialisation?

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,253 @@
...
+			return err
+		}
+		ns, err := hostNetworkSystemFromHostSystemID(vars.client, hsID)
+		if err != nil {
+			return fmt.Errorf("error loading host network system: %s", err)
+		}
+
+		_, err = hostVSwitchFromName(vars.client, ns, name)
+		if err != nil {
+			if err.Error() == fmt.Sprintf("could not find virtual switch %s", name) && !expected {
```
tombuildsstuff: I think the `!` isn't needed here?

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,253 @@
...
+resource "vsphere_host_virtual_switch" "switch" {
+  name           = "vSwitchTerraformTest"
+  host_system_id = "${data.vsphere_host.esxi_host.id}"
+
+  network_adapters = ["${var.host_nic0}", "${var.host_nic1}"]
+
+  active_nics  = ["${var.host_nic0}", "${var.host_nic1}"]
+  standby_nics = []
+}
+`, os.Getenv("VSPHERE_HOST_NIC0"), os.Getenv("VSPHERE_HOST_NIC1"), os.Getenv("VSPHERE_DATACENTER"), os.Getenv("VSPHERE_ESXI_HOST"))
```
tombuildsstuff: _minor_ 🤔 feels like it may be worth returning some kind of helper object from the testProvider code, rather than putting magic strings throughout the codebase?

vsphere/resource_vsphere_host_virtual_switch_test.go
tombuildsstuff: > Was I mistaken in that? 

I'm not sure, from what I can see:

> If no default is provided, the type is assumed to be string.

it sounds like we can rely on that assumption long-term. There's no issue with leaving these in, but `go vet` will fail if these aren't strings, given they're marked as `%s` - so I think we could safely remove it (but it's no big deal)

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+resource "vsphere_host_virtual_switch" "switch" {
+  name           = "vSwitchTerraformTest"
+  host_system_id = "${data.vsphere_host.esxi_host.id}"
+
+  network_adapters = ["${var.host_nic0}", "${var.host_nic1}"]
+
+  active_nics  = ["${var.host_nic0}", "${var.host_nic1}"]
+  standby_nics = []
```
tombuildsstuff: we should be able to remove this, since it should be the default value?

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,253 @@
...
+  name           = "vSwitchTerraformTest"
+  host_system_id = "${data.vsphere_host.esxi_host.id}"
+
+  network_adapters = ["${var.host_nic0}", "${var.host_nic1}"]
+
+  active_nics    = ["${var.host_nic0}"]
+  standby_nics   = ["${var.host_nic1}"]
+  teaming_policy = "failover_explicit"
+}
+`, os.Getenv("VSPHERE_HOST_NIC0"), os.Getenv("VSPHERE_HOST_NIC1"), os.Getenv("VSPHERE_DATACENTER"), os.Getenv("VSPHERE_ESXI_HOST"))
```
tombuildsstuff: (same here)

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -0,0 +1,253 @@
...
+resource "vsphere_host_virtual_switch" "switch" {
+  name           = "vSwitchTerraformTest"
+  host_system_id = "${data.vsphere_host.esxi_host.id}"
+
+  network_adapters = ["${var.host_nic0}"]
+
+  active_nics  = ["${var.host_nic0}"]
+  standby_nics = []
+}
+`, os.Getenv("VSPHERE_HOST_NIC0"), os.Getenv("VSPHERE_DATACENTER"), os.Getenv("VSPHERE_ESXI_HOST"))
```
tombuildsstuff: (same here)

vsphere/structure_helper.go
```diff
@@ -0,0 +1,38 @@
...
+	var d []interface{}
+	for _, v := range s {
+		d = append(d, v)
+	}
+	return d
+}
+
+// mergeSchema merges the map[string]*schema.Schema from src into dst. Safety
+// against conflicts is enforced by panicing.
+func mergeSchema(dst, src map[string]*schema.Schema) {
```
tombuildsstuff: _minor_ longer term this probably belongs in TF Core with some tests, but it's not a blocker by any means

website/docs/r/host_virtual_switch.html.markdown
```diff
@@ -0,0 +1,173 @@
...
+  teaming_policy = "failover_explicit"
+
+  allow_promiscuous      = false
+  allow_forged_transmits = false
+  allow_mac_changes      = false
+
+  shaping_enabled           = true
+  shaping_average_bandwidth = 50000000
+  shaping_peak_bandwidth    = 100000000
+  shaping_burst_size        = 1000000000
```
tombuildsstuff: 🤔 not a blocker - we've got this in OPC land too, we probably want to expose an interpolation here to make it possible to specify this in MB and translate that to bytes?

website/docs/r/host_virtual_switch.html.markdown
```diff
@@ -0,0 +1,173 @@
...
+* `link_discovery_operation` - (String, optional) Whether to `advertise` or
+  `listen` for link discovery traffic. Default: `listen`.
+* `link_discovery_protocol` - (String, optional) The discovery protocol type.
+  Valid types are `cpd` and `lldp`. Default: `cdp`.
+
+### Policy Options
+
+The following options relate to how network traffic is handled on this virtual
+switch. It also controls the NIC failover order. This subset of options is
+shared with the [`vsphere_host_port_group`][host-port-group] resource, in which
```
tombuildsstuff: `[`vsphere_host_port_group`][host-port-group]` - does the second one want brackets?

website/docs/r/host_virtual_switch.html.markdown
```diff
@@ -0,0 +1,173 @@
...
+
+* `network_adapters` - (Array of strings, required) The network interfaces to
+  bind to the bridge.
+* `beacon_interval` - (Integer, optional) The interval, in seconds, that a NIC
+  beacon packet is sent out. This can be used with
+  [`check_beacon`](#check_beacon) to offer link failure capability beyond link
+  status only. Default: `1`.
+* `link_discovery_operation` - (String, optional) Whether to `advertise` or
+  `listen` for link discovery traffic. Default: `listen`.
+* `link_discovery_protocol` - (String, optional) The discovery protocol type.
```
tombuildsstuff: Maybe this is consistent with the other docs in vmware land - but in other providers I've seen:

```
* `foo` (Optional) Setting this to `blah` does something awesome. Changing this forces a new resource to be created
```

again, not a blocker but should we be consolidating on this?

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,137 @@
...
+
+	s["shaping_enabled"].Default = false
+
+	return &schema.Resource{
+		Create: resourceVSphereHostVirtualSwitchCreate,
+		Read:   resourceVSphereHostVirtualSwitchRead,
+		Update: resourceVSphereHostVirtualSwitchUpdate,
+		Delete: resourceVSphereHostVirtualSwitchDelete,
+		Schema: s,
+	}
```
tombuildsstuff: are we supporting Import for this resource too?

vsphere/host_network_policy_structure.go
```diff
@@ -0,0 +1,286 @@
...
+
+		// HostNetworkSecurityPolicy
+		"allow_promiscuous": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Enable promiscuous mode on the network. This flag indicates whether or not all traffic is seen on a given port.",
+		},
+		"allow_forged_transmits": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
```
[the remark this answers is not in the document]
vancluever: Hey @tombuildsstuff - these are defaulted in `resource_host_virtual_switch.go`. They need to be non-defaulted here so that the schema can be shared with `vsphere_host_port_group` for inheritance purposes.

vsphere/host_network_policy_structure.go
```diff
@@ -0,0 +1,286 @@
...
+			Description: "Controls whether or not the virtual network adapter is allowed to send network traffic with a different MAC address than that of its own.",
+		},
+		"allow_mac_changes": &schema.Schema{
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Controls whether or not the Media Access Control (MAC) address can be changed.",
```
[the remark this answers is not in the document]
vancluever: As [above](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r136396879)

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,198 @@
...
+		return err
+	}
+	return nil
+}
+
+// saveHostVirtualSwitchID sets a special ID for a host virtual switch,
+// composed of the MOID for the concerned HostSystem and the virtual switch's
+// key.
+func saveHostVirtualSwitchID(d *schema.ResourceData, hsID, name string) {
+	d.SetId(fmt.Sprintf("%s:%s:%s", hostVirtualSwitchIDPrefix, hsID, name))
```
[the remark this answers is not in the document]
vancluever: Yeah, that's true - the MOID will never have a ":" in it, and the `SplitN` call is capped at 3 splits. So the remainder will go in the vSwitch name (where a `:` is a legit character).

vsphere/host_virtual_switch_structure.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// splitHostVirtualSwitchID splits a vsphere_host_virtual_switch resource ID
+// into its counterparts: the prefix, the HostSystem ID, and the virtual switch
+// name.
+func splitHostVirtualSwitchID(raw string) (string, string, error) {
+	s := strings.SplitN(raw, ":", 3)
+	if len(s) < 3 || s[0] != hostVirtualSwitchIDPrefix || s[1] == "" || s[2] == "" {
```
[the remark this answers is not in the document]
vancluever: The count should always be 3 (on part of `SplitN`), but I think I just affirmed your assessment :wink:

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -0,0 +1,137 @@
...
+	s["teaming_policy"].Default = hostNetworkPolicyNicTeamingPolicyModeLoadbalanceSrcID
+	s["check_beacon"].Default = false
+	s["notify_switches"].Default = true
+	s["failback"].Default = true
+
+	s["allow_promiscuous"].Default = false
+	s["allow_forged_transmits"].Default = true
+	s["allow_mac_changes"].Default = true
+
+	s["shaping_enabled"].Default = false
```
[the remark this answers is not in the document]
vancluever: Yeah, this is the best way I could think of doing it for now as these items are shared between resources, and the defaults can't be defined in the other resource so that `GetOkExists` works correctly (or at least my understanding of `GetOkExists`).

It would be a neat idea if eventually we got to a place where we could _transform_ schemas where we could have a base schema, and then make changes to items in the base via something a little more substantial than a single assignment.

vsphere/resource_vsphere_host_virtual_switch.go
vancluever: Import will be added eventually, but for now it is not going to be supported in this resource (DM'ed you on slack)

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: This variable is not caught as it's not necessary on all tests, but it's necessary here.

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: It needs to be there as it's a required field on `vsphere_host_virtaul_switch` (not on `vsphere_host_port_group` though).

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: Yeah I can't remember 100%. IMO, it's not a matter of whether or not that's going to change longer term for me, it's more a matter of what looks consistent for me and what matches my personal style.

Been thinking a bit more about this actually, and I think changing this kind of goes with my general philosophy when writing TF code, which is mainly treating it like how I write go code, and we have implicit types in value declarations there so why not here as well? So will change this now :smile:

vsphere/resource_vsphere_host_virtual_switch.go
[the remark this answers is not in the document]
tombuildsstuff: yeah that sounds sensible. Agree about transforming schema's - there's a couple of other options too given the number of fields this could work with a model object or even just using constants for the field names used throughout the codebase (I've been bitten by magic strings too many times!) - but that's not a blocker by any means just one to 🤔 about :)

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: Yeah, I'd like to maybe take this offline as something to think about and maybe come up with something that we can do a re-factor push later on. Right now I think it's readable enough to scale reasonably well - personally longer term, once the provider becomes mature enough to support it, I'd love to do away with the vast majority of these variables and have the resources self-host these kinds of things if at all possible.

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: [As above](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r136406733)

vsphere/resource_vsphere_host_virtual_switch_test.go
vancluever: [As above](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r136406733)

vsphere/structure_helper.go
vancluever: I think this might be another job for the [schema transformer](https://github.com/terraform-providers/terraform-provider-vsphere/pull/138#discussion_r136401357)! :wink:

website/docs/r/host_virtual_switch.html.markdown
vancluever: Yeah, might be something to take offline and think about, but yeah right now the API expects bytes, so that's what needs to be put in :stuck_out_tongue:

website/docs/r/host_virtual_switch.html.markdown
vancluever: Naw, links with references are `[text][ref]` - kind of like the numbered form (`[text][1]`).

website/docs/r/host_virtual_switch.html.markdown
vancluever: I think I've seen a number of styles, not just in this provider but in other providers as well.

At this point in time, I think there'd need to be a bigger push towards consistency and one singular convention in the provider docs here in general. For now I think I will leave it and then add a doc review as a longer term TODO to make things consistent.

vancluever:
Hey @tombuildsstuff! I think I've addressed all of the points here! Thank you very much for your thorough review and the help! :+1:

PR #137, opened 2017-08-22
vendor: Update TF to v0.10.2

vancluever:
In particular, we are using `GetOkExists` in the port group resources so
that we can determine if inheritance is needed on certain network policy
attributes properly - and it was renamed from `GetOkRaw` in v0.10.1.

PR #139, opened 2017-08-22
New resource: vsphere_host_port_group

vancluever:
This commit adds a new resource, `vsphere_host_port_group`, which can be used to manage a virtual switch on an ESXi host.

Docs included, and tests have been run against both ESXi and vCenter.

### DIFF
--- a/vsphere/host_port_group_structure.go
+++ b/vsphere/host_port_group_structure.go
@@ -1,0 +1,141 @@
+package vsphere
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+const hostPortGroupIDPrefix = "tf-HostPortGroup"
+
+// schemaHostPortGroupSpec returns schema items for resources that
+// need to work with HostPortGroupSpec, such as port groups.
+func schemaHostPortGroupSpec() map[string]*schema.Schema {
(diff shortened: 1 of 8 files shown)


radeksimko:
Do you mind rebasing & removing the vendor update from this PR as it was merged as part of https://github.com/terraform-providers/terraform-provider-vsphere/pull/137

Thanks

vancluever:
Will do the same here as I did for #138 after that one is merged (so we don't have to include the vSwitch commits here too).

vancluever:
This one has been updated to use IDs gathered from `vsphere_host` as well. Some of the suggestions that were made in the review of #138 have been added as well, and since it's been squashed hopefully it should be a lot easier to review in general. Let me know if there's any questions!

vsphere/host_port_group_structure.go
```diff
@@ -0,0 +1,141 @@
...
+func schemaHostPortGroupSpec() map[string]*schema.Schema {
+	s := map[string]*schema.Schema{
+		// HostPortGroupSpec
+		"name": &schema.Schema{
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "The name of the port group.",
+			ForceNew:    true,
+		},
+		"vlan_id": &schema.Schema{
```
mbfrahry: Thoughts on adding `ValidateFunc: validation.IntBetween(0, 4095),`

vancluever: Will add it!

vancluever:
Thanks for the review @mbfrahry! Added the `ValidateFunc` and am merging now!

PR #141, opened 2017-08-26
New data source: vsphere_vmfs_disks

vancluever:
This adds the `vsphere_vmfs_disks` data source, which allows someone to discover the storage devices that can be used for VMFS datastores on a given ESXi host.

This is the first part of a 3-part push to add datastore support to the vSphere provider - specifically, the output of this is intended to be fed into a `vsphere_vmfs_datastore`, for which a PR will be put in after I finish working on that (will be starting work now that this is done).

This is WIP because I want to make sure that #138 and #139 are merged first, in addition to hashicorp/terraform#15931 so that I can use `ValidateRegexp` out of there, although the latter item might be moot as I have already picked out the commit in question with `govendor fetch`.

### DIFF
--- a/vsphere/data_source_vsphere_vmfs_disks.go
+++ b/vsphere/data_source_vsphere_vmfs_disks.go
@@ -1,0 +1,89 @@
+package vsphere
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"sort"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func dataSourceVSphereVmfsDisks() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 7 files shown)


vsphere/data_source_vsphere_vmfs_disks.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func dataSourceVSphereVmfsDisks() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceVSphereVmfsDisksRead,
+
+		Schema: map[string]*schema.Schema{
+			"host": &schema.Schema{
+				Type:        schema.TypeString,
+				Description: "The name of the host to put this virtual switch on. This is ignored if connecting directly to ESXi, but required if not.",
```
mjrider: is this description right? The datasource name suggest it's about disks, not switches

vancluever:
Well that's embarrassing! Yeah, that's still there from copying it from the switch resource. Thanks @mjrider!

vancluever:
This PR was squashed and re-factored to just take a host ID now instead of a host and datacenter name.

Nothing else has really changed, so I am merging this one now!

PR #142, opened 2017-08-29
New resource: vsphere_vmfs_datastore

vancluever:
This is the next part of the push to add storage features to the provider.

This PR introduces the `vsphere_vmfs_datastore` resource, which can be
used to manage VMFS-based datastores.

Currently only whole disk support is included, however import
functionality is planned so that one can import existing datastores that
don't have full-disk extents, although whole disks will only be allowed
to be added from there.

Support for disk expansion may also be planned, depending on business
case.

### DIFF
--- a/vsphere/datastore_summary_structure.go
+++ b/vsphere/datastore_summary_structure.go
@@ -1,0 +1,72 @@
+package vsphere
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// schemaDatastoreSummary returns schema items for resources that
+// need to work with a DatastoreSummary.
+func schemaDatastoreSummary() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		// Note that the following fields are not represented in the schema here:
+		// * Name (more than likely the ID attribute and will be represented in
+		// resource schema)
+		// * Type (redundant attribute as the datastore type will be represented by
+		// the resource)
+		"accessible": &schema.Schema{
+			Type:        schema.TypeBool,
+			Description: "The connectivity status of the datastore. If this is false, some other computed attributes may be out of date.",
+			Computed:    true,
+		},
+		"capacity": &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Maximum capacity of the datastore, in bytes.",
+			Computed:    true,
+		},
+		"free_space": &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Available space of this datastore, in bytes.",
+			Computed:    true,
+		},
+		"maintenance_mode": &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "The current maintenance mode state of the datastore.",
+			Computed:    true,
+		},
+		"multiple_host_access": &schema.Schema{
+			Type:        schema.TypeBool,
+			Description: "If true, more than one host in the datacenter has been configured with access to the datastore",
+			Computed:    true,
+		},
+		"uncommitted_space": &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Total additional storage space, in bytes, potentially used by all virtual machines on this datastore.",
+			Computed:    true,
+		},
+		"url": &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "The unique locator for the datastore.",
+			Computed:    true,
+		},
+	}
+}
+
+// flattenDatastoreSummary reads various fields from a DatastoreSummary into
+// the passed in ResourceData.
+func flattenDatastoreSummary(d *schema.ResourceData, obj *types.DatastoreSummary) error {
+	d.Set("accessible", obj.Accessible)
+	d.Set("capacity", obj.Capacity)
+	d.Set("free_space", obj.FreeSpace)
+	d.Set("maintenance_mode", obj.MaintenanceMode)
+	d.Set("multiple_host_access", obj.MultipleHostAccess)
+	d.Set("uncommitted_space", obj.Uncommitted)
+	d.Set("url", obj.Url)
+
+	// Set the name attribute off of the name here - since we do not track this
+	// here we check for errors
+	if err := d.Set("name", obj.Name); err != nil {
+		return err
+	}
+	return nil
+}
--- a/vsphere/datastore_summary_structure.go
+++ b/vsphere/datastore_summary_structure.go
@@ -21,12 +21,12 @@ func schemaDatastoreSummary() map[string]*schema.Schema {
 		},
 		"capacity": &schema.Schema{
 			Type:        schema.TypeInt,
-			Description: "Maximum capacity of the datastore, in bytes.",
+			Description: "Maximum capacity of the datastore, in MB.",
 			Computed:    true,
 		},
 		"free_space": &schema.Schema{
 			Type:        schema.TypeInt,
-			Description: "Available space of this datastore, in bytes.",
+			Description: "Available space of this datastore, in MB.",
 			Computed:    true,
 		},
 		"maintenance_mode": &schema.Schema{
@@ -41,7 +41,7 @@ func schemaDatastoreSummary() map[string]*schema.Schema {
 		},
 		"uncommitted_space": &schema.Schema{
 			Type:        schema.TypeInt,
-			Description: "Total additional storage space, in bytes, potentially used by all virtual machines on this datastore.",
+			Description: "Total additional storage space, in MB, potentially used by all virtual machines on this datastore.",
 			Computed:    true,
 		},
 		"url": &schema.Schema{
@@ -56,11 +56,11 @@ func schemaDatastoreSummary() map[string]*schema.Schema {
 // the passed in ResourceData.
 func flattenDatastoreSummary(d *schema.ResourceData, obj *types.DatastoreSummary) error {
 	d.Set("accessible", obj.Accessible)
-	d.Set("capacity", obj.Capacity)
-	d.Set("free_space", obj.FreeSpace)
+	d.Set("capacity", byteToMB(obj.Capacity))
+	d.Set("free_space", byteToMB(obj.FreeSpace))
 	d.Set("maintenance_mode", obj.MaintenanceMode)
 	d.Set("multiple_host_access", obj.MultipleHostAccess)
-	d.Set("uncommitted_space", obj.Uncommitted)
+	d.Set("uncommitted_space", byteToMB(obj.Uncommitted))
 	d.Set("url", obj.Url)
 
 	// Set the name attribute off of the name here - since we do not track this
--- a/vsphere/datastore_summary_structure.go
+++ b/vsphere/datastore_summary_structure.go
@@ -36,7 +36,7 @@ func schemaDatastoreSummary() map[string]*schema.Schema {
 		},
 		"multiple_host_access": &schema.Schema{
 			Type:        schema.TypeBool,
-			Description: "If true, more than one host in the datacenter has been configured with access to the datastore",
+			Description: "If true, more than one host in the datacenter has been configured with access to the datastore.",
 			Computed:    true,
 		},
 		"uncommitted_space": &schema.Schema{
--- a/vsphere/host_data_store_system_helper.go
+++ b/vsphere/host_data_store_system_helper.go
@@ -58,7 +58,7 @@ func diskSpecForCreate(dss *object.HostDatastoreSystem, name string) (*types.Vmf
 	defer cancel()
 	options, err := dss.QueryVmfsDatastoreCreateOptions(ctx, disk.DevicePath)
 	if err != nil {
-		return nil, fmt.Errorf("could not get disk creation options for %s: %s", name, err)
+		return nil, fmt.Errorf("could not get disk creation options for %q: %s", name, err)
 	}
 	var option *types.VmfsDatastoreOption
 	for _, o := range options {
@@ -68,7 +68,7 @@ func diskSpecForCreate(dss *object.HostDatastoreSystem, name string) (*types.Vmf
 		}
 	}
 	if option == nil {
-		return nil, fmt.Errorf("cannot use entire disk on device %s for datastore", name)
+		return nil, fmt.Errorf("device %q is not available as a new whole-disk device for datastore", name)
 	}
 	return option.Spec.(*types.VmfsDatastoreCreateSpec), nil
 }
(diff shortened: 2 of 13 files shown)


vancluever:
This has been rebased/squashed now that a large number of other new resources and work has been merged. Docs are still pending, as is import.

vsphere/datastore_summary_structure.go
```diff
@@ -0,0 +1,72 @@
...
+			Computed:    true,
+		},
+		"multiple_host_access": &schema.Schema{
+			Type:        schema.TypeBool,
+			Description: "If true, more than one host in the datacenter has been configured with access to the datastore",
+			Computed:    true,
+		},
+		"uncommitted_space": &schema.Schema{
+			Type:        schema.TypeInt,
+			Description: "Total additional storage space, in bytes, potentially used by all virtual machines on this datastore.",
```
tombuildsstuff: would this make sense to return MB here rather than bytes? I can't think of a scenario when this would be helpful as Bytes?

vsphere/datastore_summary_structure.go
```diff
@@ -0,0 +1,72 @@
...
+	d.Set("capacity", obj.Capacity)
+	d.Set("free_space", obj.FreeSpace)
+	d.Set("maintenance_mode", obj.MaintenanceMode)
+	d.Set("multiple_host_access", obj.MultipleHostAccess)
+	d.Set("uncommitted_space", obj.Uncommitted)
+	d.Set("url", obj.Url)
+
+	// Set the name attribute off of the name here - since we do not track this
+	// here we check for errors
+	if err := d.Set("name", obj.Name); err != nil {
```
tombuildsstuff: I don't see the need for only this field?

vsphere/host_data_store_system_helper.go
```diff
@@ -0,0 +1,147 @@
...
+	disk, err := availableScsiDisk(dss, name)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultAPITimeout)
+	defer cancel()
+	options, err := dss.QueryVmfsDatastoreCreateOptions(ctx, disk.DevicePath)
+	if err != nil {
+		return nil, fmt.Errorf("could not get disk creation options for %s: %s", name, err)
```
tombuildsstuff: can we make this first argument `%q` to quote it?

vsphere/host_data_store_system_helper.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	var option *types.VmfsDatastoreOption
+	for _, o := range options {
+		if _, ok := o.Info.(*types.VmfsDatastoreAllExtentOption); ok {
+			option = &o
+			break
+		}
+	}
+	if option == nil {
+		return nil, fmt.Errorf("cannot use entire disk on device %s for datastore", name)
```
tombuildsstuff: could we add some more information to this message and change the formatting argument to `%q`?

vsphere/host_data_store_system_helper.go
```diff
@@ -0,0 +1,147 @@
...
+	disk, err := availableScsiDisk(dss, name)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultAPITimeout)
+	defer cancel()
+	options, err := queryVmfsDatastoreExtendOptions(ctx, dss, ds, disk.DevicePath, true)
+	if err != nil {
+		return nil, fmt.Errorf("could not get disk extension options for %s: %s", name, err)
```
tombuildsstuff: can we use `%q` as the first formatting argument here?

vsphere/host_data_store_system_helper.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	var option *types.VmfsDatastoreOption
+	for _, o := range options {
+		if _, ok := o.Info.(*types.VmfsDatastoreAllExtentOption); ok {
+			option = &o
+			break
+		}
+	}
+	if option == nil {
+		return nil, fmt.Errorf("cannot use entire disk on device %s for datastore", name)
```
tombuildsstuff: can we switch out the formatting argument to `%q` and provide a better reason for this?

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+	waitForDeletePending   = "waitForDeletePending"
+	waitForDeleteCompleted = "waitForDeleteCompleted"
+	waitForDeleteError     = "waitForDeleteError"
+)
+
+// formatCreateRollbackError defines the verbose error for extending a disk on
+// creation where rollback is not possible.
+const formatCreateRollbackErrorUpdate = `
+WARNING: Dangling resource!
+There was an error extending your datastore with disk: %s:
```
tombuildsstuff: can this formatting argument become `%q` to quote it?

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+%s
+Additionally, there was an error removing the created datastore:
+%s
+You will need to remove this datastore manually before trying again.
+`
+
+// formatUpdateInconsistentState defines the verbose error when the setting
+// state failed in the middle of an update opeartion. This is an error that
+// will require repair of the state before TF can continue.
+const formatUpdateInconsistentState = `
```
tombuildsstuff: could we not make use of `d.Partial()` to continue & retry any pending steps?

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+WARNING: Inconsistent state!
+Terraform was able to add disk %s, but could not save the update to the state.
```
tombuildsstuff: `%s` -> `%q`

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+		"host_system_id": &schema.Schema{
+			Type:        schema.TypeString,
+			Description: "The managed object ID of the host to set up the datastore on.",
+			Required:    true,
+		},
+		"disks": &schema.Schema{
+			Type:        schema.TypeList,
+			Description: "The disks to add to the datastore.",
+			Required:    true,
+			MinItems:    1,
```
tombuildsstuff: given this is required `MinItems` should be inferred?

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating datastore with disk %s: %s", disk, err)
+	}
+
+	// Now add any remaining disks.
+	for _, disk := range disks[1:] {
+		spec, err := diskSpecForExtend(dss, ds, disk.(string))
+		if err != nil {
+			// We have to destroy the created datastore here.
+			if remErr := removeDatastore(dss, ds); remErr != nil {
```
tombuildsstuff: as above, can we track this with `d.Partial` to allow us to re-attempt this in the a future plan?

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+				return fmt.Errorf(formatCreateRollbackErrorUpdate, disk, err, remErr)
+			}
+			return fmt.Errorf("error fetching datastore extend spec for disk %s: %s", disk, err)
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), defaultAPITimeout)
+		defer cancel()
+		if _, err := extendVmfsDatastore(ctx, dss, ds, *spec); err != nil {
+			if remErr := removeDatastore(dss, ds); remErr != nil {
+				// We could not destroy the created datastore and there is now a dangling
+				// resource. We need to instruct the user to remove the datastore
```
tombuildsstuff: +1

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+				// manually.
+				return fmt.Errorf(formatCreateRollbackErrorUpdate, disk, err, remErr)
+			}
+			return fmt.Errorf("error extending datastore with disk %s: %s", disk, err)
```
tombuildsstuff: `%s` -> `%q`

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+	if err != nil {
+		return fmt.Errorf("could not get properties for datastore: %s", err)
+	}
+	if err := flattenDatastoreSummary(d, &props.Summary); err != nil {
+		return err
+	}
+
+	// We also need to update the disk list from the summary.
+	var disks []string
+	for _, disk := range props.Info.(*types.VmfsDatastoreInfo).Vmfs.Extent {
```
tombuildsstuff: there's a ton of potential crashes we're not tackling here? I've seen use use flatten methods to track this in other providers

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	// Maintain a copy of the disks that have been added just in case the update
+	// fails in the middle.
+	old2 := make([]interface{}, len(old.([]interface{})))
+	copy(old2, old.([]interface{}))
+
+	// Now we basically reverse what we did above when we were checking for
+	// removed disks, and add any new disks that have been added.
+	for _, v1 := range new.([]interface{}) {
```
tombuildsstuff: the more I read into this, the more this feels like it belongs in the SDK?

vsphere/resource_vsphere_vmfs_datastore.go
```diff
@@ -0,0 +1,324 @@
...
+
+	// This is a race that more than likely will only come up during tests, but
+	// we still want to guard against it - when working with datastores that end
+	// up mounting across multiple hosts, removing the datastore will fail if
+	// it's removed too quickly (like right away, for example). So we set up a
+	// very short retry waiter to make sure if the first attempt fails, the
+	// second one should probably succeed right away. We also insert a small
+	// minimum delay to make an honest first attempt at trying to delete the
+	// datastore without spamming the task log with errors.
+	deleteRetryFunc := func() (interface{}, string, error) {
```
tombuildsstuff: again, this feels like SDK logic

vsphere/resource_vsphere_vmfs_datastore.go
vancluever: What I want to avoid is someone attempting to enter an empty list here - ie: `disk = []`.

vsphere/datastore_summary_structure.go
vancluever: This is because `name` is not set in the accompanying `schemaDatastoreSummary`, as it's an attribute that may be controlled in different ways and go through different workflows depending on the resource. At the same time though, we want to flatten it from the summary data as it's one of the easiest ways to fetch it from the datastore. So rather than just drop the result on the floor if by any chance it's _not_ in the schema, I want to check for errors so that any edge cases get caught (more than likely in testing, so the user will never see it, but it has value nonetheless).

vancluever: This review item was helpful though as now that we use MoRefs for the IDs here on the datastore, we can support renaming of the datastore and also remove a ForceNew case if someone has changed it manually :) I just added that support.

vsphere/resource_vsphere_vmfs_datastore.go
vancluever: This isn't necessarily a problem with state that was already set, but more with state that could _not_ be set. When we are adding extents, we need to add the extent, then update the state in real time so that in the event that some of the disks were added, but some were not, the disks that were already added are properly saved to state, and we can pick up where we left off on the next Terraform run. This "inconsistent state" error is to catch the edge case where, for some reason, we could not set the `TypeList` properly and hence we are left with a situation were the disk was added, but the state was not updated, at which point the state would need to be manually repaired (although a `terraform refresh`) could actually fix it - I've updated the error message to reflect this.

vsphere/resource_vsphere_vmfs_datastore.go
vancluever: I think I'd like to keep the `Create` operation atomic - creating a datastore and adding extents is not a ridiculously expensive operation, and if a user is expecting a datastore to be created with a certain number of extents and one of the extents is no good for the datastore, I think it makes more sense to not have a datastore at all versus a partially created one.

vsphere/resource_vsphere_vmfs_datastore.go
vancluever: Since we are dealing in VMFS datastores only here, this is guaranteed to be `types.VmfsDatastoreInfo`. So if the assertion fails, this is a breaking bug that needs to be looked at ASAP, so I'm thinking the crash is warranted here.

vsphere/resource_vsphere_vmfs_datastore.go
vancluever: There is no method in `Datastore` or `HostDatastoreSystem` that would encapsulate this AFAIK, and even in the vSphere Client you need to extend datastores extent by extent.

I think it would be neat if it was in the SDK - my problem with that is if we start to send feature requests or pull requests over to VMware, are now blocked by VMware's schedule for getting those changes merged upstream so that we can have clean code. I would rather implement things ourselves for now, especially while we are pushing to get a reasonable feature set in the provider, while keeping things clean enough so that if said functionality becomes available upstream, refactoring should just be as simple as a drop-in replacement.

tombuildsstuff: That makes sense - that said there's nothing stopping us wrapping this outside of the resource which would be an intermediate step. This isn't a blocker by any means, more one to think about - but I'd argue it's worth refactoring this out if possible :)

vsphere/resource_vsphere_vmfs_datastore.go
vancluever: There is a property waiter, but it's unclear what property to follow here - I think the major factor in this race is if the disk is a shared VMFS disk or not (example: iSCSI).

This only started coming up when I started testing iSCSI VMFS disks versus local ones (local disks actually have another issue where you can't add extents through vCenter, you need to connect to the host directly, so I moved the vCenter tests over to use iSCSI disks that I provisioned from the lab NAS).

In the event of a shared VMFS datastore, vCenter will automatically try to mount it on every host in a cluster that can access it, and I think that part is causing issues - the task returns, yet other hosts are still scanning for changes and mounting/updating the datastore. During this transition, I think `multiple_host_access` will be false - but also at the same time, this is not really a stable thing to wait on, as there's too many variables that we would need to control (knowing that a disk is shared, or what hosts it exactly needs to be shared on, is currently out of the scope of this resource).

So I think this might be the simpler of the two scenarios. To that end I've updated things so that we specifically wait on the `ResourceInUse` error type instead of a string match (as per the other error matcher).

vancluever:
This is finally out of WIP! Finished off the docs yesterday, but as a bonus today, I added folder support - with a ton of scaffolding for better support for inventory for all major objects in general. But yeah - good to go!

vancluever:
Thanks @mbfrahry! Just discovered a diff issue on an import smoke test and I'm fixing that and then I'll be merging :+1:

PR #144, opened 2017-08-30
New data source: vsphere_datacenter

vancluever:
This data source fetches the managed object ID for a vSphere datacenter, and is
intended for use in other up-and-coming resources that will use a
managed object reference to a datacenter to fetch other resources (like
host systems), or tell resources where they need to be deployed to (may
eventually be used in `vsphere_virtual_machine`, for example).

### DIFF
--- a/vsphere/data_source_vsphere_datacenter_test.go
+++ b/vsphere/data_source_vsphere_datacenter_test.go
@@ -1,0 +1,100 @@
+package vsphere
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+var testAccDataSourceVSphereDatacenterExpectedRegexp = regexp.MustCompile("^datacenter-")
+
+func TestAccDataSourceVSphereDatacenter(t *testing.T) {
+	var tp *testing.T
+	testAccDataSourceVSphereDatacenterCases := []struct {
+		name     string
+		testCase resource.TestCase
+	}{
+		{
+			"basic",
+			resource.TestCase{
+				PreCheck: func() {
+					testAccPreCheck(tp)
+					testAccDataSourceVSphereDatacenterPreCheck(tp)
+				},
+				Providers: testAccProviders,
+				Steps: []resource.TestStep{
+					{
+						Config: testAccDataSourceVSphereDatacenterConfig(),
+						Check: resource.ComposeTestCheckFunc(
+							resource.TestMatchResourceAttr(
+								"data.vsphere_datacenter.dc",
+								"id",
+								testAccDataSourceVSphereDatacenterExpectedRegexp,
+							),
+							resource.TestMatchResourceAttr(
+								"data.vsphere_datacenter.dc",
+								"datacenter_id",
+								testAccDataSourceVSphereDatacenterExpectedRegexp,
+							),
+						),
+					},
+				},
+			},
+		},
+		{
+			"default",
+			resource.TestCase{
+				PreCheck: func() {
+					testAccPreCheck(tp)
+					testAccDataSourceVSphereDatacenterPreCheck(tp)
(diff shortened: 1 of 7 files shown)


website/docs/d/datacenter.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (String) The name of the datacenter. This can be a name or path.	If
+  not provided, the default datacenter is used.
+
+~> **NOTE:** `name` is ignored on ESXi, and is not required.
```
mbfrahry: Thoughts on adding that the default datacenter will be used when using ESXi? You mention it above but I think hammering it home isn't a bad idea.

vsphere/data_source_vsphere_datacenter_test.go
```diff
@@ -0,0 +1,100 @@
...
+		t.Skip("set VSPHERE_DATACENTER to run vsphere_datacenter acceptance tests")
+	}
+}
+
+func testAccDataSourceVSphereDatacenterConfig() string {
+	return fmt.Sprintf(`
+data "vsphere_datacenter" "dc" {
+  name = "%s"
+}
+`, os.Getenv("VSPHERE_DATACENTER"))
```
mbfrahry: Hopefully this isn't too nitpicky but I think that `os.Getenv("VSPHERE_DATACENTER"))` should be passed into the function instead of being called here. It'd be `func testAccDataSourceVSpohereDatacenterConfig(datacenter string) string {...}`

website/docs/d/datacenter.html.markdown
```diff
@@ -0,0 +1,39 @@
...
+
+## Attribute Reference
+
+* `id` - The managed object ID of this datacenter.
+* `datacenter_id` - (String) The managed object ID of this datacenter (same as
```
mbfrahry: Having a hard time seeing the justification for `datacenter_id` when it is the same as `id`. What're your thoughts on the benefits of `datacenter_id`?

vsphere/data_source_vsphere_datacenter_test.go
vancluever: I'm kind of on the fence with this one - one one hand I can see the merit of the full parameterization, allowing more flexibility of the generated config, on the other hand the only reason config is being generated in this case is to ensure that `VSPHERE_DATACENTER` can be passed in, otherwise it'd be a constant config like the default test, but with a static value for `name`. What I'd hate to see is this be a pattern where test cases end up scaling out and something like this has to be pasted in every test:

```
Config: testAccDataSourceVSphereDatacenterConfig(os.Getenv("VSPHERE_DATACENTER")),
```

Especially if the vast majority of the configs would take the parameter like that, it would start to seem redundant. Thoughts?

mbfrahry: I can stand behind that! I was also on the fence when I sent it away but you make valid points so I'm cool with keeping it this way!

website/docs/d/datacenter.html.markdown
vancluever: Done here! Lots of more clarification and have decided that we will _require_ this on `vsphere_host` no matter if we are dealing with ESXi or not as it just helps reduce ambiguity here (which is kind of the reason I'm writing these data sources in the first place).

website/docs/d/datacenter.html.markdown
vancluever: Removed this attribute now - my justification for this was kind of how the "name attribute" works in Chef where there is one attribute that ends up being the name/ID for a resource - but of course this is not Chef ;)

vancluever:
Hey @mbfrahry - addressed all of the above points - let me know when you have been able to take another look!

vancluever:
Thanks @mbfrahry!

PR #145, opened 2017-08-30
d/datacenter: Re-word "default" datacenter doc references

vancluever:
The "default" datacenter is actually for the most part a gvmomi thing, a
consequence of using the `DefaultDatacenter` calls in the find package.
This function simulates default datacenter fetching behaviour by
returning a single datacenter object from inventory if, and only if, a
singular datacenter object exists in inventory. This is always the case
on ESXi, but not on vCenter, and if there are multiple datacenters found
the call will fail. Docs have been updated to reflect this.


vancluever:
Thanks @catsby!

PR #148, opened 2017-08-31
r/host_virtual_switch: Doc update for ID attribute

vancluever:
The ID attribute is now a composite attribute that we assemble in code,
and not just the name of the resource, because that is not a unique
value or very telling of where the resource is actually located.

The ID change went in with #138, but I forgot to update the docs.


vancluever:
Just going to ship this one as it's a small doc update and I am releasing v0.2.1 shortly.

PR #149, opened 2017-09-05
New resource: vsphere_nas_datastore

vancluever:
This is the WIP PR for the new `vsphere_nas_datastore` resource.

It's nearly complete, but we need documentation and proper multiple host support. I was going to let Terraform naturally handle this, but race conditions happen if you try to create a second mount (basically a `CreateNasDatastore` with the exact same spec on another host) too close to the first. This can be seen in the current incarnation of the multi-host test where I have to use `depends_on` to chain the two resources.

The new version will just take multiple `host_system_ids`, which will ultimately simplify configurations and tests anyway.

Also, this PR is based on top of #142 so once that's merged I will rebase off of master instead.


vancluever:
WIP removed on this one too, now ready for review! After the `vsphere_vmfs_datastore` resource gets merged I will probably rebase this one on master to get rid of the irrelevant commits.

vsphere/host_nas_volume_structure.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		"type": &schema.Schema{
+			Type:        schema.TypeString,
+			Default:     "NFS",
+			Description: "The type of NAS volume. Can be one of NFS (to denote v3) or NFS41 (to denote NFS v4.1).",
+			ForceNew:    true,
+			Optional:    true,
+			ValidateFunc: validation.StringInSlice(
+				[]string{
+					string(types.HostFileSystemVolumeFileSystemTypeNFS),
```
mbfrahry: Loving all the constants

vsphere/resource_vsphere_nas_datastore_test.go
```diff
@@ -0,0 +1,455 @@
...
+					testAccPreCheck(tp)
+					testAccResourceVSphereNasDatastorePreCheck(tp)
+					testAccSkipIfEsxi(tp)
+				},
+				Providers: testAccProviders,
+				Steps: []resource.TestStep{
+					{
+						Config: testAccResourceVSphereNasDatastoreConfigMultiHost(),
+						Check: resource.ComposeTestCheckFunc(
+							testAccResourceVSphereNasDatastoreExists(true),
```
mbfrahry: It doesn't look like it but I'd just like to confirm that there isn't any other variable to check if it's basic/multi-host?

vsphere/host_nas_volume_structure.go
vancluever: :grinning:

vsphere/resource_vsphere_nas_datastore_test.go
vancluever: `Read` picks this up reasonably well (the host ID list is updated on refresh, and if there's a mismatch the test will fail with a non-empty plan)... we could probably have a harder check, but writing the check function might be a little overly complex as we would have to derive the host IDs from the host names in the environment variables (`VSPHERE_ESXI_HOST`, `VSPHERE_ESXI_HOST2`, and `VSPHERE_ESXI_HOST3`). I think this should be pretty good unless there's some sort of serious edge case that comes up.

vancluever:
Merging this now - thanks @mbfrahry!

PR #152, opened 2017-09-08
r/virtual_machine: Enforce powered on state

### LINKED ISSUE #134: Terraform crashed when powered off existing "vsphere_virtual_machine" resource then apply again
### Terraform Version
Terraform v0.10.1

### Affected Resource(s)
- terraform-provider-vsphere

```hcl
resource "vsphere_virtual_machine" "PengDemoRedhat65" {

    count = "${length(var.vms_address)}"

    datacenter = "DC"
    resource_pool = "/Resource_Pool"

    name = "${element(var.vms_name, count.index)}"
    dns_suffixes = "${var.dns_suffixes}"
    dns_servers  = "${var.dns_servers}"
    vcpu = "${element(var.vms_vcpu, count.index)}"
    memory = "${element(var.vms_memory, count.index)}"

    network_interface {
        label = "VM Network"
        ipv4_prefix_length = "22"
        ipv4_address = "${element(var.vms_address, count.index)}"
        ipv4_gateway = "${element(var.vms_gateway, count.index)}"
    }

    disk {
        datastore = "Automation"
        template = "Template_Redhat_6.5"
        type = "thin"
    }
}
```

### Debug Output
https://gist.github.com/Rocking80/758754e3837bc226a4c88c99327d7f9f

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.
Sorry cannot provide the crash.log which contains lots of private information.

### Expected Behavior
The shutdown machine can be powered on at 2nd `terraform apply`

### Actual Behavior
terraform crashed.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` succeeds and the vms are created on vcenter
2. shutdown one of remote vmware machine
3. `terraform apply` again

vancluever:
This commit adds something that we generally assume is correct in
Terraform - that an instance's power state is either one of two states:
on, or gone.

`vsphere_virtual_machine` has had a couple of issues that have been
attributed to power state in the last few months that have been fixed,
but the issue of ensuring that VMs are always powered on has not been
touched on yet.

In this commit, we add the `power_state` pseudo-computed variable,
working around the fact that the provider can't manipulate the diff
directly by using an immutable default to always indicate that a VM that
is powered off or suspended will be turned back on during the next
Terraform run. If a VM is powered off, this turns the VM back on during
the configuration modification phase at the end of the update.

Fixes #134.
Related to #103.

### DIFF
--- a/vsphere/helper_test.go
+++ b/vsphere/helper_test.go
@@ -116,8 +116,8 @@ func testGetVirtualMachine(s *terraform.State, resourceName string) (*object.Vir
 	return virtualMachineFromUUID(tVars.client, uuid)
 }
 
-// testGetVirtualMachineProperties is a convenience that adds an extra step to
-// testGetVirtualMachine to get the properties of a virtual machine.
+// testGetVirtualMachineProperties is a convenience method that adds an extra
+// step to testGetVirtualMachine to get the properties of a virtual machine.
 func testGetVirtualMachineProperties(s *terraform.State, resourceName string) (*mo.VirtualMachine, error) {
 	vm, err := testGetVirtualMachine(s, resourceName)
 	if err != nil {
--- a/vsphere/helper_test.go
+++ b/vsphere/helper_test.go
@@ -1,14 +1,18 @@
 package vsphere
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"regexp"
 	"testing"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
@@ -19,6 +23,9 @@ type testCheckVariables struct {
 
 	// The subject resource's ID.
 	resourceID string
+
+	// The subject resource's attributes.
+	resourceAttributes map[string]string
 
 	// The ESXi host that a various API call is directed at.
 	esxiHost string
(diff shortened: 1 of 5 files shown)


vsphere/helper_test.go
```diff
@@ -93,3 +101,59 @@ func testGetPortGroup(s *terraform.State, resourceName string) (*types.HostPortG
...
+	if err := task.Wait(tctx); err != nil {
+		return fmt.Errorf("error waiting for poweroff: %s", err)
+	}
+	return nil
+}
+
+// copyStatePtr returns a TestCheckFunc that copies the reference to the test
+// run's state to t. This allows access to the state data in later steps where
+// it's not normally accessible (ie: in pre-config parts in another test step).
+func copyStatePtr(t **terraform.State) resource.TestCheckFunc {
```
mbfrahry: Are the pointers to pointers necessary?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -330,6 +333,47 @@ func TestAccVSphereVirtualMachine_noPanicShutdown(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testBasicPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckVSphereVirtualMachineDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					copyStatePtr(&state),
```
mbfrahry: If we didn't do a pointer to a pointer with the above comment we could just pass `state` instead of `&state`. Unless I'm missing something

vancluever: Yeah, the double pointer is necessary because a `terrafrom.State` instance is always passed by reference. So we need to directly modify the value of the pointer, or else it's essentially passed by value and does not get retained. [Example of a failure of a pass by value function](https://play.golang.org/p/IevcXmWnxY). Unless I'm missing something. :)

A different approach would be to use `State.Copy`, but I don't necessarily want a _copy_ here, but instead the in-flight state.

vsphere/helper_test.go
vancluever: Double pointer rationale [here](https://github.com/terraform-providers/terraform-provider-vsphere/pull/152#discussion_r137841449).

vancluever:
Hey @mbfrahry, also, we actually want a non-empty plan here on powered down (there is actually a test in master testing this right now) because we always want to diff `poweredOff => poweredOn` to indicate that powered off is _not_ a valid power state to be in.

Unfortunately with the current state of affairs with `vsphere_virtrual_machine`, a powered off VM creates a non-empty plan even without this update, but that's going to be a longer-term battle.

Reference to the existing test:  https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/resource_vsphere_virtual_machine_test.go#L300

PR #154, opened 2017-09-09
website: Add resource categories

vancluever:
We are getting more resources now, and it will be worthwhile to start
departing them into categories now.

I am breaking down the categories mainly into their specific resource
classes as they are generally seen in vCenter:

* Hosts and Clusters
* VMs and Templates
* Storage
* Networking

There are other classes as well that will have resources that are not
necessarily related to be above that make logical sense:

* Administration - For vCenter and vSphere SSO related admin resources.
Licenses are in this category currently and items like tags and what not
will probably fall here as well.
* Inventory - For inventory management resources, such as datacenters
(which is a fancy top-level resource grouping) and folders (which can be
used across all four of the major resource classes).

### DIFF
--- a/website/vsphere.erb
+++ b/website/vsphere.erb
@@ -25,47 +25,70 @@
           </ul>
         </li>
 
-        <li<%= sidebar_current("docs-vsphere-resource") %>>
-          <a href="#">Resources</a>
+        <li<%= sidebar_current("docs-vsphere-resource-inventory") %>>
+          <a href="#">Inventory Resources</a>
           <ul class="nav nav-visible">
-            <li<%= sidebar_current("docs-vsphere-resource-datacenter") %>>
+            <li<%= sidebar_current("docs-vsphere-resource-inventory-datacenter") %>>
               <a href="/docs/providers/vsphere/r/datacenter.html">vsphere_datacenter</a>
             </li>
-            <li<%= sidebar_current("docs-vsphere-resource-file") %>>
-              <a href="/docs/providers/vsphere/r/file.html">vsphere_file</a>
-            </li>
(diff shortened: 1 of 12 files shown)


website/vsphere.erb
```diff
@@ -25,47 +25,70 @@
...
-            </li>
-            <li<%= sidebar_current("docs-vsphere-resource-license") %>>
-              <a href="/docs/providers/vsphere/r/license.html">vsphere_license</a>
-            </li>
-            <li<%= sidebar_current("docs-vsphere-resource-virtual-disk") %>>
+          </ul>
+        </li>
+
+        <li<%= sidebar_current("docs-vsphere-resource-vm") %>>
+          <a href="#">VM and Template Resources</a>
```
tombuildsstuff: it might be VMware terminology - but I'd argue `Template` could be removed here?

website/vsphere.erb
```diff
@@ -25,47 +25,70 @@
...
+              <a href="/docs/providers/vsphere/r/host_port_group.html">vsphere_host_port_group</a>
+            </li>
+            <li<%= sidebar_current("docs-vsphere-resource-networking-host-virtual-switch") %>>
+              <a href="/docs/providers/vsphere/r/host_virtual_switch.html">vsphere_host_virtual_switch</a>
+            </li>
+          </ul>
+        </li>
+
+        <li<%= sidebar_current("docs-vsphere-resource-admin") %>>
+          <a href="#">Administration Resources</a>
```
tombuildsstuff: can all of these sections be ordered alphabetically?

vancluever:
Hey @tombuildsstuff - thanks for the feedback! Will be implementing both points and then merging!

PR #157, opened 2017-09-14
r/virtual_machine: Refactor tests

vancluever:
Maintaining this file has become increasingly difficult. As such, 
I have re-factored the tests in the style as we are doing most
of the new tests - highly functional, static configs with a tolerable
level of repetition, and all tests set up as subtests, so every test is
easy to locate as part of the larger test case list.

I have also removed a few tests that currently do not function, for one
reason or another. A couple of these are due to present bugs in the
resource, but as pending work may deprecate the resource in its current
form, I think maintenance effort is better spent elsewhere as they are
not currently reported business critical bugs that are rendering the
resource unusable.

The tests were:

* A test with a CDROM virtual device - this test was panicking. As it's
hard to imagine who is using VMs in TF with CDROM devices attached, I
think this is a bit of an edge case.
* A test for a NIC with DHCP. In theory this should work, but it's hard
to test properly without dealing with races in VMware tools, so the test
itself probably did not have much value.
* A test to detach disks added out-of-band. This is indeed a bug as
`Read` does not seem to be picking up the extra disk and triggering the
needed diff. Again, this is something that will be picked up on the
refactor, or might look completely different afterwards.

Some other tests were consolidated - memory and CPU changes are now one
test, and the individual `keep_on_remove` test for disk devices was
dropped as we now test this with the test that adds an external disk via
`vsphere_virtual_disk`, where a lack of `keep_on_remove` on the attachment
actually causes issues during destroy as the disk is removed
prematurely, with the virtual machine.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine_test.go
+++ b/vsphere/resource_vsphere_virtual_machine_test.go
@@ -16,6 +16,15 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+const (
+	testAccResourceVSphereVirtualMachineDiskNameEager     = "terraform-test-extra-eager"
+	testAccResourceVSphereVirtualMachineDiskNameLazy      = "terraform-test-extra-lazy"
+	testAccResourceVSphereVirtualMachineDiskNameThin      = "terraform-test-extra-thin"
+	testAccResourceVSphereVirtualMachineDiskNameExtraVmdk = "terraform-test-vm-extra-disk.vmdk"
+	testAccResourceVSphereVirtualMachineStaticMacAddr     = "06:5c:89:2b:a0:64"
+	testAccResourceVSphereVirtualMachineAnnotation        = "Managed by Terraform"
+)
+
 func TestAccResourceVSphereVirtualMachine(t *testing.T) {
 	var tp *testing.T
 	var state *terraform.State
@@ -429,9 +438,9 @@ func testAccResourceVSphereVirtualMachineCheckExtraDisks() resource.TestCheckFun
 		}
 
 		var foundEager, foundLazy, foundThin bool
-		expectedEagerName := "terraform-test-extra-eager.vmdk"
-		expectedLazyName := "terraform-test-extra-lazy.vmdk"
-		expectedThinName := "terraform-test-extra-thin.vmdk"
+		expectedEagerName := testAccResourceVSphereVirtualMachineDiskNameEager + ".vmdk"
+		expectedLazyName := testAccResourceVSphereVirtualMachineDiskNameLazy + ".vmdk"
+		expectedThinName := testAccResourceVSphereVirtualMachineDiskNameThin + ".vmdk"
 
 		for _, dev := range props.Config.Hardware.Device {
 			if disk, ok := dev.(*types.VirtualDisk); ok {
@@ -474,10 +483,10 @@ func testAccResourceVSphereVirtualMachineCheckFolder(expected string) resource.T
 		if err != nil {
 			return fmt.Errorf("bad: %s", err)
 		}
-		expected, err := rootPathParticleVM.PathFromNewRoot(vm.InventoryPath, rootPathParticleVM, expected)
+		expected, err = rootPathParticleVM.PathFromNewRoot(vm.InventoryPath, rootPathParticleVM, expected)
 		actual := path.Dir(vm.InventoryPath)
 		if err != nil {
-			return fmt.Errorf("bad: %s", err)
+			return fmt.Errorf("bad: %s - inventory path", err)
 		}
 		if expected != actual {
 			return fmt.Errorf("expected path to be %s, got %s", expected, actual)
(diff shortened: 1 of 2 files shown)


vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+		}
+		return nil
 	}
+}
 
-	if v := os.Getenv("VSPHERE_IPV4_ADDRESS"); v == "" {
-		t.Fatal("env variable VSPHERE_IPV4_ADDRESS must be set for acceptance tests")
+// testAccResourceVSphereVirtualMachineCheckHostname is a check to check for a
+// VirtualMachine's hostname. The check uses guest info, so VMware tools needs
+// to be installed.
```
mbfrahry: What would happen if VMware tools weren't installed? Is there a way to check if the tools are installed before running tests that call this?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+// backings, and expects them in the exact order outlined in the function.
+func testAccResourceVSphereVirtualMachineCheckExtraDisks() resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		props, err := testGetVirtualMachineProperties(s, "vm")
+		if err != nil {
+			return err
+		}
+
+		var foundEager, foundLazy, foundThin bool
+		expectedEagerName := "terraform-test-extra-eager.vmdk"
```
mbfrahry: Is there any chance for these names to change? I'm thinking we make these constants

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+		expectedLazyName := "terraform-test-extra-lazy.vmdk"
```
mbfrahry: Thoughts on making these constants?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+		expectedThinName := "terraform-test-extra-thin.vmdk"
```
mbfrahry: Thoughts on making this a constant?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+// for the MAC address defined in the static MAC address test on the first
+// network interface.
+func testAccResourceVSphereVirtualMachineCheckStaticMACAddr() resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		props, err := testGetVirtualMachineProperties(s, "vm")
+		if err != nil {
+			return err
+		}
+		actual := props.Guest.Net[0].MacAddress
+		expected := "06:5c:89:2b:a0:64"
```
mbfrahry: Thoughts on making this a constant?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+
+// testAccResourceVSphereVirtualMachineCheckAnnotation is a check to ensure
+// that a VM's annotation is correctly set in the annotation test.
+func testAccResourceVSphereVirtualMachineCheckAnnotation() resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		props, err := testGetVirtualMachineProperties(s, "vm")
+		if err != nil {
+			return err
+		}
+		expected := "Managed by Terraform"
```
mbfrahry: Thoughts on constant?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -1,1726 +1,1548 @@
...
+		actual := props.Config.Annotation
```
mbfrahry: And constant here?

vsphere/resource_vsphere_virtual_machine_test.go
vancluever: If VMware tools is not installed, I think this test will either fail due to no data being present or panic, depending on how well populated `Guest` is populated in this scenario. In any case, it will fail, which is what we want it to do.

A good chunk of the resource right now really depends on VMware tools being installed on the template that you are basing it off of, and all of our tests (and most of the old tests as well) depend on a template. There might be a scenario in the future where we have a good use case where we need to deploy something with Terraform without VMware tools, like a virtual appliance - when we have that, we will probably need to create an environment variable that can be used to skip tests like these... but for now we don't have a test that would consume it, so it's not really necessary.

vsphere/resource_vsphere_virtual_machine_test.go
vancluever: This is the actual value which we need to pull from the VM state (`Config` here denotes the configuration section of the retrieved properties).

mbfrahry: My bad. Didn't notice the lack of ""

vancluever:
Thanks @mbfrahry! Just needed to review a couple of breaking tests before I merged here :+1:

PR #158, opened 2017-09-14
r/virtual_machine: Updated customization and net waiter behaviour

### LINKED ISSUE #140: Customization does not work for windows 2008 R2 template
### Terraform Version
v0.10.1

### Affected Resource(s)
 - terraform-provider-vsphere

```hcl
resource "vsphere_virtual_machine" "TerraformSampleMachine" {

    count = "${length(var.vms_address)}"

    datacenter = "${var.datacenter}"
    resource_pool = "${var.resource_pool}"

    name = "${element(var.vms_name, count.index)}"
    dns_suffixes = "${var.dns_suffixes}"
    dns_servers  = "${var.dns_servers}"
    vcpu = "${element(var.vms_vcpu, count.index)}"
    memory = "${element(var.vms_memory, count.index)}"


    network_interface {
        label = "VM Network"
        ipv4_prefix_length = "22"
        ipv4_address = "${element(var.vms_address, count.index)}"
        ipv4_gateway = "${element(var.vms_gateway, count.index)}"
    }

    disk {
        datastore = "${var.datastore}"
        template = "${var.templates["windows_2008_R2"]}"
        type = "thin"
    }
}
```

### Debug Output
https://gist.github.com/Rocking80/e9f55a5b5403ca1a629c798ed9e1b5fb 

### Expected Behavior
The windows 2008 R2 machine should be created and configured correctly. And `terraform apply` finish successfully.

### Actual Behavior
The windows 2008 R2 machine is created but network for it (ip, gateway, DNS) and hostname is not configured. And `terraform apply` is always printing 'Still creating' without ending.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
From the vcenter console, we can see the vmware tasks are done without error. 
tasks are:
1. clone virtual machine
2. Reconfigure virtual machine
3. Reconfigure virtual machine
4. Customize virtual machine guest OS
5. Power on virtual machine

After all tasks are finished and wait for half an hour, I am trying to logon to the machine, there is message prompt to restart the machine. I restarted the machine manually and still does not work. 

The vmware tools are latest on the template machine.

Windows 2012 R2 does work.

Don't know if this is a terraform issue or a vmware API issue?

vancluever:
This is a two-part change that ultimately changes/Fixes customization
workflow, but has some other implications:

* First off, we are now waiting for customization to complete by
watching VM events. This is the correct way to wait for customization to
complete as the actual customization task returns very quickly, while
the actual customization process can actually take a few minutes and
fail out-of-band of the initial task.

* Second off, we are now no longer waiting for all interfaces to have IP
addresses during Read, rather we are now waiting for a **routeable**
network. This fixes situations where a template may be configured for
DHCP and gets an auto-configuration address before actually getting an
IP. This seems to come up during Windows customization especially,
which is what we are trying to fix with this work, so having the event
watch fix without having this one would not fix the issue completely.

* The `wait_for_guest_net` option has also been added, which allows
someone to completely turn off the network waiter. This should help
alleviate some edge cases where NICs have not been configured with IP
addresses, and also allows someone to bypass this behaviour if they are
not configuring a gateway on the VM through either static configuration
of DHCP.

-----

Finally, as part of the additional testing work, the network waiter has been moved to the `Create` and `Update` functions of the resource, rather than the `Read` function. This is required especially on `Create` to ensure that created VMs can be rolled back properly by `terraform destroy`, which will do an initial refresh prior to destroy to get an updated state.

Fixes #140.

### DIFF
--- a/tf-vsphere-devrc.mk.example
+++ b/tf-vsphere-devrc.mk.example
@@ -22,6 +22,7 @@ export VSPHERE_ALLOW_UNVERIFIED_SSL ?= false
 # The following variables are shared across various tests. To ensure all tests
 # succeed, it's probably best to set all of these to valid values.
 export VSPHERE_TEMPLATE            ?= base-linux # VM template to clone
+export VSPHERE_TEMPLATE_WINDOWS    ?= base-win   # Windows VM template
 export VSPHERE_NETWORK_LABEL       ?= vm-network # Port group label
 export VSPHERE_NETWORK_LABEL_DHCP  ?= vm-network # Port group label for DHCP
 export VSPHERE_IPV4_ADDRESS        ?= 10.0.0.100 # Customization IP address
--- a/vsphere/event_helper.go
+++ b/vsphere/event_helper.go
@@ -11,6 +11,11 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+// A list of known event IDs that we use for querying events.
+const (
+	eventTypeCustomizationSucceeded = "CustomizationSucceeded"
+)
+
 // virtualMachineCustomizationWaiter is an object that waits for customization
 // of a VirtualMachine to complete, by watching for success or failure events.
 //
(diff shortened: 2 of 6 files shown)


vancluever:
Thanks @mbfrahry!

rismoney:
What I have found in my testing, is that OS customization can be reported to VMWare as completed, but that event can NOT be relied on.

The only way I have been able to conclusively determine if customization is complete is a local script that I wrote (power-cli) which scans guestcust.log for "Deleted Folder C:\\sysprep"  - Every other attempt has been either built around wait guessing, or eventing in vsphere which isn't accurate.

vancluever:
Hey @rismoney - thanks for the info. Have you been encountering issues with the new customization behaviour related to this? If so, do you mind putting in a new bug report with the specific issue?

The customization timeout has been adjusted to 10 minutes in v0.4.0 which should cover most cases, especially the timeout that happens here is exclusive to customization only. If you end up getting an error due to the event never showing up in vSphere, you will get an error similar to what is being seen in #160.

If you do see that, let us know in the new issue as well. This will help us gauge if this behaviour requires any further adjustments.

Thanks for your patience on this!

rismoney:
I have recently upgraded to 0.4.0 from original .1 and now terraformed bails out around 4-5 minutes.  The machine gets created and OS customization kicks off and completes but TF gets hosed. This problem has existed in 0.2+.  I wonder if there is something a problem with API change and vsphere.  I opened another issue with the tracelog.

I don't know if the issue I experience on customization is related or if there is a vsphere bug reporting it's completion before done.  I believe some sort of scheduled task runs that could be minutes after vsphere believes it to be done. Regardless the wait script I wrote handles this edge case flawlessly so I moved forward.

Issue #160, opened 2017-09-18
Windows 2012 customization does not finish in required by defaultAPITimeout time


krzkowalczyk:
Hi there,

After implementing PR [#158](https://github.com/terraform-providers/terraform-provider-vsphere/pull/158) we now check customization events to ensure the process has finished. Therefore in my test environment customization on win2012 takes more than 5min (which is set in defaultAPITimeout), and cause timeout error.

### Terraform Version
Terraform v0.10.5.

### Affected Resource(s)
- vsphere_virtual_machine

### Terraform Configuration Files
```hcl
resource "vsphere_virtual_machine" "mwawposhck003" {
  name               = "mwawposhck003"
  folder             = "terraformtest/HEALTHCHECK/Development"
  annotation         = "HealthCheck 03 - Windows 2012 Standard"
  vcpu               = 6
  memory             = 8192
  datacenter         = "MyDataCenter"
  cluster            = "MyCluster"
  dns_suffixes       = ["mydomain.pl"]
  dns_servers        = ["192.168.1.6", "192.168.1.8"]
  time_zone          = 100
  skip_customization = false

  network_interface {
    label       = "VM Network 896"
  }

  disk {
    datastore = "Dell_MD/Dell_MD_Storage01"
    template  = "terraformtest/W2K12R2STDUAT"
    type      = "thin"
  }
}
```

### Debug Output
https://gist.github.com/krzkowalczyk/e8b71b209b860445cf89e287c44507e2

### Panic Output
https://gist.github.com/krzkowalczyk/e8b71b209b860445cf89e287c44507e2

### Expected Behavior
Terraform waits for end of customization.
`defaultAPITimeout` should be increased, or there is a new APITimeout parameter available to configure in resource.

### Actual Behavior
Terraform gets timeout after 5min.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
It takes some time and at least 2 reboots after powering on, until node ends customization.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- [GH-158](https://github.com/terraform-providers/terraform-provider-vsphere/pull/158)

vancluever:
[...]

PR #162, opened 2017-09-18
increase defaultAPITimeout to 10min

krzkowalczyk:
This PR responds to issue [#160](https://github.com/terraform-providers/terraform-provider-vsphere/issues/160)

The default value of `defaultAPITimeout` in some cases like vmware customization is set to too low value.
The proposal is to set `defaultAPITimeout` to 10 minutes from 5 minutes. I hope it should be enough for all vsphere actions.

Alternative solution is to add a global parameter `api_timeout` to vsphere provider or even to populate it to specific resources.

### DIFF
--- a/vsphere/provider.go
+++ b/vsphere/provider.go
@@ -10,7 +10,7 @@ import (
 
 // defaultAPITimeout is a default timeout value that is passed to functions
 // requiring contexts, and other various waiters.
-var defaultAPITimeout = time.Minute * 5
+var defaultAPITimeout = time.Minute * 10
 
 // Provider returns a terraform.ResourceProvider.
 func Provider() terraform.ResourceProvider {
--- a/vsphere/resource_vsphere_virtual_machine_snapshot.go
+++ b/vsphere/resource_vsphere_virtual_machine_snapshot.go
@@ -63,7 +63,7 @@ func resourceVSphereVirtualMachineSnapshotCreate(d *schema.ResourceData, meta in
 	if err != nil {
 		return fmt.Errorf("Error while getting the VirtualMachine :%s", err)
 	}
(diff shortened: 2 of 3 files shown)


vancluever:
Hey @krzkowalczyk - sorry you are having the issues with customization! However, I want to keep the default timeout to 5 minutes as there are a number of calls that should almost never take 5 minutes, let alone 10, using this value. Also keep in mind that the default provider timeout is 20 minutes, and that most of the contexts using this timeout are not re-used, so while the risk is negligible, it would be a lot more possible in theory for the calls to run over the provider timeout, which will lead to misleading errors, one of the things we are trying to avoid by moving to properly timed contexts.

We will handle this one case though by increasing the timeout for customization specifically to 10 minutes for now to see how that goes. Keep in mind that after this call completes there is still the `waitForGuestVMNet` timeout as well, actually runs over the defined 5 minutes, I'm guessing due to backoff timers. We want to make sure both of these can run and theoretically timeout before the provider timeout.

I'm going to close this one in favour of the patch that we put in for the customization timeout specifically.

If 10 minutes proves to not be enough, we can revisit this and see about making the timeout configurable. I'd like to not be in the habit of allowing long customizations however - lengthy provisioning processes should be handled out-of-band from the initial creation of the resource, either through provisioners, or by pre-building the image with [Packer](https://www.packer.io/).

Thanks!

krzkowalczyk:
Hey @vancluever, thanks for reviewing this PR ! Your arguments make sens to me, so we can just set timeout for customization to 10min. 
Currently it takes ~8min to finish customization on non-ssd, where indeed the template is build by Packer, and the only customization we're looking to make is change hostname, timezone and administrator's password.

Thanks !

PR #164, opened 2017-09-19
New resource: vsphere_tag_category

vancluever:
This adds a new resource, `vsphere_tag_category`, which is the start of
the effort to get tag support into the provider. This allows one to
create and manage tag categories in the provider, which can then be
passed along to tag resources.

As part of this work, other work has been done to prime the provider for
the ability to manage multiple API connections at once, as the provider
now needs to connect to two to do its job. Furthermore, tags are not
supported on all versions of vSphere and not on ESXi, so extra
precautions needed to be undertaken to ensure these cases were properly
blocked.


vendor/vendor.json
```diff
@@ -11,8 +17,8 @@
 		{
 			"checksumSHA1": "+2yCNqbcf7VcavAptooQReTGiHY=",
 			"path": "github.com/apparentlymart/go-rundeck-api",
```
tombuildsstuff: why is Rundeck needed?

vsphere/resource_vsphere_tag_category.go
```diff
@@ -0,0 +1,251 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The display name of the category.",
+				Required:    true,
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Description: "The description of the category.",
+				Optional:    true,
```
tombuildsstuff: should this also be Computed?

vsphere/resource_vsphere_tag_category.go
```diff
@@ -0,0 +1,251 @@
...
+					false,
+				),
+			},
+			"associable_types": {
+				Type:        schema.TypeSet,
+				Description: "Object types to which this category's tags can be attached.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Required:    true,
+			},
+			"used_by": {
```
tombuildsstuff: I feel like this is named incorrectly?

vsphere/resource_vsphere_tag_category.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil, fmt.Errorf("could not get category for name %q: %s", name, err)
+	}
+
+	if len(cats) < 1 {
+		return nil, fmt.Errorf("category name %q not found", name)
+	}
+	if len(cats) > 1 {
+		// This should not happen but guarding just in case it does.
+		return nil, fmt.Errorf("category name %q returned multiple results, more specific name required", name)
+	}
```
tombuildsstuff: why not combine these to be: 
```
if len(cats) == 1) {
  d.SetId(cats[0].ID)
 return []*schema.ResourceData{d}, nil 
}

return nil, fmt.Errorf("Expected a single category, got '%d'", len(cats))

File: vsphere/resource_vsphere_tag_category_test.go [hunk 5]
```diff
@@ -0,0 +1,308 @@
...
+			resource.Test(t, tc.testCase)
+		})
+	}
+}
+
+func testAccResourceVSphereTagCategoryExists(expected bool) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		_, err := testGetTagCategory(s, "terraform-test-category")
+		if err != nil {
+			if strings.Contains(err.Error(), "Status code: 404") && !expected {
```
tombuildsstuff: is there no way of getting a concrete object back and checking the status code on that instead?

website/docs/r/tag_category.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+specific objects.
+
+For more information about tags, click [here][ext-tags-general]. For more
+information about tag categories specifically, click
+[here][ext-tag-categories].
+
+[ext-tags-general]: https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.vcenterhost.doc/GUID-E8E854DD-AA97-4E0C-8419-CE84F93C4058.html
+[ext-tag-categories]: https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.vcenterhost.doc/GUID-BA3D1794-28F2-43F3-BCE9-3964CB207FB6.html
+
+~> **NOTE:** Tagging support is unsupported on direct ESXi connections and
```
tombuildsstuff: this probably also warrants a section on the main vSphere page for "Required vSphere Versions" IMO (although, [checking the page](https://www.terraform.io/docs/providers/vsphere/index.html) perhaps it wants a separate page for that?)

website/docs/r/tag_category.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (String, required) The name of the tag.
+* `description` - (String, optional) A description for the tag.
+* `cardinality` - (String, required, forces new resource) The number of tags
```
tombuildsstuff: _minor_ in order providers we suffix the description with "Changing this forces a new resource to be created" rather than doing this

website/docs/r/tag_category.html.markdown
```diff
@@ -0,0 +1,107 @@
...
+  that can be assigned from this category to a single object at once. Can be
+  one of `SINGLE` (object can only be assigned one tag in this category), to
+  `MULTIPLE` (object can be assigned multiple tags in this category).
+* `associable_types` - (List of strings, required) A list object types that
+  this category is valid to be assigned to. For a full list, click
+  [here](#associable-object-types).
+
+~> **NOTE:** You can add associable types to a category, but you cannot remove
+them. Attempting to do so will result in an error.
```
tombuildsstuff: sounds like this should be `ForceNew`?

vendor/vendor.json
vancluever: `govendor add` strikes again :stuck_out_tongue:

I'm going to try this again against my fork using `govendor fetch`, and see if I get the same deps pulled in.

vsphere/resource_vsphere_tag_category.go
vancluever: Config should be the source of truth, even if there's no `description` defined, so this should not be computed so that it can be reverted to a blank description if set out-of-band.

vsphere/resource_vsphere_tag_category.go
vancluever: Yeah, I almost feel like removing this field altogether. I will do a bit more research on it. [API entry where it's referenced](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/1cd28284-3b72-4885-9e31-d1c6d9e26686/71ef7304-a6c9-43b3-a3cd-868b2c236c81/doc/operations/com/vmware/cis/tagging/category.get-operation.html).

vsphere/resource_vsphere_tag_category_test.go
vancluever: Nope. The library does not wrap an error type to pull it from ([reference](https://github.com/vmware/vic/blob/master/pkg/vsphere/tags/categories.go#L120)).

website/docs/r/tag_category.html.markdown
vancluever: _Adding_ types is a legitimate operation, removing them is not - so we can't be `ForceNew` here as that would block the case where someone wants to add a type to a category.

website/docs/r/tag_category.html.markdown
vancluever: Yeah, I'm still saving this work for a documentation overhaul. Keeping consistent with the style that I have on other docs for now, as that will allow me to easier fix this later. I haven't forgotten!

website/docs/r/tag_category.html.markdown
vancluever: Honestly that index page needs a massive overhaul anyway so I'm not surprised. For now I think I'm fine with it being out of date, but close to v1.0.0 (or during some downtime) I will work on updating it so that not only does it have this information, but the dev info is removed, and other steps are taken to make sure that it focuses on a summary of what the provider does and that's it, saving extra detail for each individual resource.

vsphere/resource_vsphere_tag_category.go
vancluever: There's 2 specific cases here - the first one is when an incorrect category is given and the one that will be the one most users will encounter (`len(cats) < 1`). Technically, it seems impossible for `len(cats) > 1` to happen at the moment, as category names need to be unique (adding a category with a different name but different associable types in TF or the UI will result in a name collision error). Yet `GetCategoriesByName` [returns multiple](https://github.com/vmware/vic/blob/master/pkg/vsphere/tags/categories.go#L169), even though its own logic searches for an [exact match](https://github.com/vmware/vic/blob/master/pkg/vsphere/tags/categories.go#L183). So the second case is just for brevity to match the SDK and I don't expect it to really happen. Regardless, given these two cases are so distinct I'd prefer to have this level of verbosity over a more general `len != 1` style error message, which I think is more confusing...

Also as I'm typing this I think the second error is a little misleading too - might end up re-wording this as well, but we still need to distinguish between "not found" (something a user may stumble on if they just mis-typed the name of the category) versus `len > 1` (which is more of a change of the expected API and something we will need to fix).

vancluever:
Merging now - thanks for the review @tombuildsstuff!

PR #167, opened 2017-09-19
New data source: vsphere_tag_category

vancluever:
This is the data source counterpart to the new `vsphere_tag_category`
resource. It allows one to pull in an external tag category by name,
using most of the same search call stack as importing.

### DIFF
--- a/website/docs/d/tag_category.html.markdown
+++ b/website/docs/d/tag_category.html.markdown
@@ -1,0 +1,40 @@
+---
+layout: "vsphere"
+page_title: "VMware vSphere: vsphere_tag_category"
+sidebar_current: "docs-vsphere-data-source-tag-category"
+description: |-
+  Provides a vSphere tag category data source. This can be used to reference tag categories not managed in Terraform.
+---
+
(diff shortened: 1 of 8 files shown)


website/docs/d/tag_category.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+
+The following arguments are supported:
+
+* `name` - (String, required) The name of the tag category.
+
+## Attribute Reference
+
+In addition to the `id` being exported, all of the fields that are available in
+the [`vsphere_tag_category` resource][resource-tag-category] are also
+populated. See that page for further details.
```
mbfrahry: Thoughts about linking that page?

vancluever: It is linked - just [above](https://github.com/terraform-providers/terraform-provider-vsphere/pull/167/files#diff-5a14dfcd6ae271d5c4361dadf90d0af5R17) as it's the same link :slightly_smiling_face:

vancluever:
Thanks @mbfrahry! Merging now

PR #169, opened 2017-09-20
provider: Fix version comparison

vancluever:
Got some feedback from early adopters that the version comparison code
was broken and now using the provider on an unsupported version of
vSphere (ie: pre 6.0) was not working.

Added tests and discovered some issues with the code that was indeed
causing things to break probably:

 * Product name was not being populated properly
 * Needed to break off version comparison from the rest of the
 older/newer logic so a reverse check on the version could be used to
 first confirm that a version was not newer/older before moving on to
 compare build versions.

Also updated the provider's version check to make sure it is also returning ineligible on if the products don't match. By design, `Older/Newer` return `false` for different products (it's not older or newer, it's different).

### DIFF
--- a/vsphere/tags_helper.go
+++ b/vsphere/tags_helper.go
@@ -36,7 +36,8 @@ func isEligibleTagEndpoint(client *govmomi.Client) bool {
 	if err := validateVirtualCenter(client); err != nil {
 		return false
 	}
-	if parseVersionFromClient(client).Older(tagsMinVersion) {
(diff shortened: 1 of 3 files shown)


kleinsasserm:
```./terraform plan``` is now working again with vSphere 5.5 :+1:

vancluever:
Ty @mbfrahry and @kleinsasserm!  :+1:

PR #171, opened 2017-09-20
New resource and data source: vsphere_tag

vancluever:
This adds the `vsphere_tag` resource and data source, allowing the ability to manage and pull in tags.

Next step is adding the ability to tag all of the objects in the provider that can be tagged!

PS: This also includes an update to the vic package which was also missing tag update support.

### DIFF
--- a/website/docs/r/tag.html.markdown
+++ b/website/docs/r/tag.html.markdown
@@ -1,0 +1,78 @@
+---
+layout: "vsphere"
+page_title: "VMware vSphere: vsphere_tag"
+sidebar_current: "docs-vsphere-resource-inventory-tag-resource"
+description: |-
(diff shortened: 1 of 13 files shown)


website/docs/r/tag.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (String, required) The display name of the tag. The name must be
+  unique within its category.
+* `description` - (String, optional) A description for the tag.
+* `category_id` - (String, required, forces new resource) The unique identifier
```
mbfrahry: Thoughts on putting the required arguments next to each other?

vancluever:
Applied doc change and pushed - thanks @mbfrahry! Merging now

Issue #172, opened 2017-09-21
vsphere_virtual_machine - support for guest_id ?


ze42:
Hello,

### Terraform Version
Terraform, latest: v0.10.6
vpshere provider: latest, and master branch

### Affected Resource(s)
vsphere_virtual_machine

### Current Behavior

It does not seem to support setting/changing the GuestId.
From what I read in the code, unless based on a template, GuestId is set to `otherLinux64Guest`, and no other value can be specified.

### Expected Behavior

A way to specify guest_id and override default behaviour with specific value.

Did I miss anything, or is it a feature that could be added?

sofixa:
There isn't a way to set it, for now.
At work we're currently working on that and a few other improvements(like the ability to modify VMs), and we'll submit a PR soon(end of week).

Cheers, 
Adrian

ze42:
@sofixa Good to hear.

Was probably about to start working on it in the next days, but if you have a PR coming this way with much more features, I'm looking forward to see this PR :)

vancluever:
Hey @ze42, just a FYI that I have just closed #174. At this point in time changing of the guest ID is not something that we are planning to support.

There is a rewrite of `vsphere_virtual_machine` planned in the next couple of months that will be addressing the current complications of the existing resource, with the objective of simplifying the resource overall and being very specific on what you can do in-band with the resource. Creating VMs from scratch will be reviewed, but more than likely, templates will be the major method of bringing VMs into Terraform, with the recommendation to use tools like [Packer](https://www.packer.io/) to configure the VM hardware the way you want it to look.

Thanks!

ze42:
Ok... I'll consider Packer to build images once it will be able to deploy directly on a vSphere cluster using the API (did not find anything in its documentation so far).

Is there a way to currently build a VM from an existing template, reusing the hardware specs? Did not find any... so, if there isn't any, I'll have to use the #174 for now...

ze42:
@sofixa Thanks for the tip. Will probably help at some point.

But I meant build directly on vsphere. (are there some base official ova file for debian9 available somewhere? like EC2 templates available directly on maketplace by Debian...)

sofixa:
[...]

PR #174, opened 2017-09-21
Add the ability to modify VMs without recreating; add new parameters for cpu and memory hot add, guest_id, adapter_type

sofixa:
With this PR, you can modify virtual_machine resources, and use extra parameters - cpu and memory hot add, guest_id (ping @ze42 and #172  ) and adapter_type for network interfaces.

Open to reviews and comments from everyone :) The updates to the doc are incoming ASAP(tomorrow or early next week).

Adding my colleague who did most of the work to the conversation: @congiohj 

Cheers,
Adrian

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -1,6 +1,7 @@
 package vsphere
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"net"
@@ -8,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/hashicorp/terraform/helper/validation"
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
@@ -39,13 +39,14 @@ var DiskControllerTypes = []string{
 type networkInterface struct {
 	deviceName       string
 	label            string
+	key              string
 	ipv4Address      string
 	ipv4PrefixLength int
 	ipv4Gateway      string
 	ipv6Address      string
 	ipv6PrefixLength int
 	ipv6Gateway      string
-	adapterType      string // TODO: Make "adapter_type" argument
+	adapterType      string
 	macAddress       string
 }
 
@@ -80,15 +81,19 @@ type memoryAllocation struct {
 type virtualMachine struct {
 	name                  string
 	hostname              string
+	guestId               string
 	folder                string
+	rebootAllowed         bool
 	datacenter            string
 	cluster               string
 	resourcePool          string
 	datastore             string
 	vcpu                  int32
+	cpuHotAddEnabled      bool
 	memoryMb              int64
 	memoryAllocation      memoryAllocation
 	annotation            string
+	memoryHotAddEnabled   bool
 	template              string
 	networkInterfaces     []networkInterface
 	hardDisks             []hardDisk
@@ -132,61 +137,77 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
+				ForceNew: false,
 			},
-
 			"hostname": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
+				ForceNew: false,
 			},
-
+			"guest_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+			},
 			"folder": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
 			},
-
+			"reboot_allowed": &schema.Schema{
+				Type:     schema.TypeBool,
+				Required: false,
+				Optional: true,
+				Default:  false,
+				ForceNew: false,
+			},
 			"vcpu": &schema.Schema{
 				Type:     schema.TypeInt,
 				Required: true,
 			},
-
+			"cpu_hot_add_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Required: false,
+				Optional: true,
+				Default:  true,
+				ForceNew: false,
+			},
 			"memory": &schema.Schema{
 				Type:     schema.TypeInt,
 				Required: true,
 			},
-
 			"memory_reservation": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
 				Default:  0,
-				ForceNew: true,
+				ForceNew: false,
 			},
-
+			"memory_hot_add_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Required: false,
+				Optional: true,
+				Default:  true,
+				ForceNew: false,
+			},
 			"annotation": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-
 			"datacenter": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
 			},
-
 			"cluster": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
+				ForceNew: false,
 			},
-
 			"resource_pool": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
+				ForceNew: false,
 			},
-
 			"linked_clone": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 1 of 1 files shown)


vancluever:
Hey @sofixa, thanks for the pull request!

We are actually planning to re-do the `vsphere_virtual_machine` resource in the next couple of months - part of this work will be possibly making some opinionated design decisions on what you can and can't do in-band with a VM. These decisions will be made in line with core HC philosophies, such as immutable infrastructure, leaving more complex lifecycle workflows to the build stage of a virtual machine, using things like [Packer](https://www.packer.io/) and what not.

Given this, I unfortunately don't think this PR is a good fit for those plans. There are a lot of things in here that make the VM resource _less_ immutable, such as changing the guest ID, to address one point. These are the kinds of things that more than likely won't be making it into the rewrite as it implies ongoing support for situations that will more than likely become more complex over time, which is the opposite of what we will be trying to address with the rewrite.

Sorry, and thanks again!

sofixa:
@vancluever I respectfully, but wholeheartedly disagree.

I fully understand the advantages of immutable, but IMHO:

* the majority of people running VMware don't run immutable infrastructre(if they're mature enough to be able to do that, they're probably mature enough to realise VMware brings no advantages over basically anything else for this use case, *unless* they also have to support "legacy" VMs side-by-side with immutable ones, which brings me to the next point)
* migrating paradigms is hard to do. You can't just trash everything and start over, there will certainly be some overlap; having this in mind, don't you think it's a better idea to be able to do "legacy" mode with terraform on VMware so that people can start doing what they usually do, and then, already having the enabling tooling(terraform, packer) in place, migrate to more modern and flexible approaches(immutable).
* not all environments are created equal - where i work(we're a managed services/hosting/cloud/consulting provider, meaning we have hundreds of clients running on VMware, AWS, GCP, Azure, OVH) we already use Packer for AMIs, vagrant boxes, OVAs and what not;  we already use terraform for AWS, GCP, Kubernetes, Grafana. We've been looking at the terraform vsphere provider since ~2015(iirc), because it would be extremely nice to be able to use the same tooling regardless of the underlying infrastructure, but it was nowhere near usable for the way we used VMware(create VMs with a custom tool that uses .yml configuration files, deploy with Chef, and use it to manage the life cycle of the VM). Recently we decided that the benefit of using the same tools (which would also enable us to do much more interesting things with VMware than what we currently do, once we've migrated to them) across infrastructures outweighs the time we'll spent on development, and here was the resultant PR.
* various other terraform providers allow you do to life cycle management and modifications of your instances(AWS for instance); does what you say mean all of them will be disabled? If not, i find it quirky that it's the VMware one(one of the least likely platforms for immutable infrastructure) that will be opinionatedly handicapped. 
* `ongoing support for situations that will more than likely become more complex over time` IMHO there is nothing complex in modifying the number of vCPUs on a VMware VM, regardless of how many times you do it, and this PR doesn't add things much more complicated than that
* this provider has had plenty of interesting community contributions(especially recently), so i think maintaining support for moderately basic stuff like modifying VMs is certainly doable

TL;DR - making this provider unusable for non-immutable environments will severely limit its usability by the VMware community; it'd be much easier to move to immutable if you already have the tooling that permits you to do that AND your "legacy" workflows, rather than having to maintain two completely different tool sets and workflows; there are cases where immutable is not an option, yet, and my previous argument is a blessing.

Thanks in advance for reading the wall of text,
Kindly hoping you might reconsider,
Cheers
Adrian

ze42:
@sofixa Thanks alot. I'll have a look... will save me time.

From what I read in your initial post, it looks like *exactly* like what I would have had to do by myself in the comming weeks (guest_id, adapter_type... and optionaly allow hot cpu/ram change).

@vancluever 
From what I understand, some work on packer & terraform will allow to use some nice template as base for VM creation... Yeah, guess that could probably do the trick for guest_id.

Allowing hot cpu/memory change will probably be also an option to prepare in template... but can they be set/used yet?

How would that have any effect on selecting the adapter_type for your network card? No, even a good template doesn't necessarly have the same number of card, and changing/adding cards will probably need to specify the adapter_type anyway.

And as @sofixa mentionned, updating/changing your instances is already something you can do on other platform, why not allow it in vSphere with what it provides? You can change instance type on AWS (though it need a restart...)


Other changes I'll obviously need over template VM is obviously #99 too. A template doesn't necessarly have the same disks, and same sizes as the one you'll want on your different production servers.

bpoland:
Just wanted to chime in here.  It's been great to see all of the activity on this provider over the past few months and I really appreciate all of the work that @vancluever and others have been doing.

With that said, I have to agree with a lot of the stuff that @sofixa mentioned in his post.  I really hope that the rewrite of the provider is flexible enough that it doesn't limit the options for folks that can't go fully immutable.

And as was mentioned in https://github.com/terraform-providers/terraform-provider-vsphere/issues/172 the packer support for vsphere is pretty limited.  Your only option is to build a VM from source ISO every time, and you need direct hypervisor access.  It isn't possible to build a VM that has most of what you need, and use that as a base template for packer.

ze42:
@sofixa Testing some PR... and hitting a bug I think is related to this one...

Using master + a few PR, but this is the only one touching VM network cards.

I created the VM ("failed" due to not booting properly and retrieving an IP...), but planning again, I get an error:

```
$ terraform plan --out=tf.plan
[...]
vsphere_virtual_machine.innovation-vpn-1: Refreshing state... (ID: CERBERHOST/innovation/n4a-vpn/innovation-vpn-1)
Error refreshing state: 1 error(s) occurred:

* vsphere_virtual_machine.innovation-vpn-1: 1 error(s) occurred:

* vsphere_virtual_machine.innovation-vpn-1: vsphere_virtual_machine.innovation-vpn-1: unexpected EOF
panic: interface conversion: types.BaseVirtualDeviceBackingInfo is *types.VirtualEthernetCardDistributedVirtualPortBackingInfo, not *types.VirtualEthernetCardNetworkBackingInfo
2017-09-22T15:57:26.207Z [DEBUG] plugin.terraform-provider-vsphere: 
2017-09-22T15:57:26.207Z [DEBUG] plugin.terraform-provider-vsphere: goroutine 87 [running]:
2017-09-22T15:57:26.207Z [DEBUG] plugin.terraform-provider-vsphere: github.com/terraform-providers/terraform-provider-vsphere/vsphere.resourceVSphereVirtualMachineRead(0xc4201db340, 0x105c3e0, 0xc420430b50, 0x1, 0x1e0b620)
2017-09-22T15:57:26.207Z [DEBUG] plugin.terraform-provider-vsphere:     [...]/src/github.com/terraform-providers/terraform-provider-vsphere/vsphere/resource_vsphere_virtual_machine.go:1207 +0x35d6
```

Network interface looks like:

```
  network_interface = {
    adapter_type = "vmxnet3"
    label = "dvpg_xxx/dvpg-xxx-dvs-1-1603"
    mac_address = "00:1C:14:C1:03:02"
  }
```

(Connected to a virtual port group... to be directly on a specific VLAN...)

vancluever:
Hey all, thank you for your feedback!

First off, I just want to clarify: while I am saying I will be reviewing VM creation workflows, by no means will they be going away. Nor will VMs be _completely_ immutable. Sorry if I gave anyone that impression!

To address some of the specific points:

I think we need to be careful in assuming _why_ people are choosing to continue to use vSphere, or on-prem, over a cloud provider. A number of factors could be at play, such as migration costs, data sovereignty, or general application fitness. HashiCorp is supporting the vSphere provider as a first-class ecosystem provider because there is an overwhelming amount of people still using vSphere for whatever reason, and we don't see that going away anytime soon.

With that said, HC has a clear vision of _what_ we want our tools to do. [The Tao of HashiCorp](https://www.hashicorp.com/blog/the-tao-of-hashicorp/) gives us a mission to design simple workflows that involve immutable entities, which ultimately become modifiable through versioning. While nothing is black and white (also accounted for in the Tao), and indeed the compute resource in any provider is probably a shining example of how complicated things can be, our way forward needs to keep these things in mind, so that we can be sure we are helping people to use our tools according to the core philosophies they were designed with.

This is a balance that we will be considering when we work on re-designing the resource. There are three workflows that we will be looking to support: cloning of a VM from template, importing of a VM from an OVF/OVA template, and "bare metal" creation of a virtual machine. Each of these has their own challenges, and different API workflows to boot, so it's important to consider the consequences of say, how defining VM hardware in code will affect the expectations of everyone in all three workflows, and how such changes will behave in the lifecycle of a Terraform configuration itself.

As such, there more than likely will be things that we will not allow modification on in the resource specifically, or they may be immutable in the resource itself, with the ability to attach them externally through separate resources (such as how the AWS provider works with ENIs and EBS volumes).

Finally, considering all of the above - while this work is pending, I want to keep the change going on with the `vsphere_virtual_machine` resource down to very pointed quality-of-life updates and bug fixes, as evident with some of the recent work done to support Windows customization. Addressing the features in this PR specifically, I'd be open to reviewing and merging a PR that had the ability to do hot modifications of CPU and memory _without_ flagging (which is something we will be doing in the new version of this resource). Will keep an eye out for that if it gets sent along!

Thanks!

PS: @sofixa - on a personal note I worked 10+ years in hosting, with a good chunk of that working with vSphere, and would have loved to have had Terraform as a tool that I could have used to deploy clients. So I'm definitely keeping those use cases in mind as we move forward!

bpoland:
@vancluever thanks for the details, it all sounds great to me (particularly the separate attachment resources, I have had to do that with external calls to govc).  When you say "cloning of a VM from template" I hope that will still allow cloning from regular virtual machines (like the current provider) as well as vSphere templates?

Thanks again for all your work on the vsphere provider, looking forward to these changes!

(PS do you happen to know if anyone is/will be working on better packer support for vsphere?  As I mentioned above it's pretty limited right now.  As a workaround I actually ended up using terraform to create a VM, then the packer "null" builder to do the work inside the VM, and then govc to shut the VM down.  But would love to use more "native" packer functionality instead)

PR #177, opened 2017-09-24
tags: Add tagging support for datacenters

vancluever:
This update adds in the tagging support for datacenters. Just a few
minor modifications, nothing too serious, although this does now add the
`Read` function as tags on a datacenter (any object for that matter) are
modifiable in a non-destructive way.


vancluever:
Thanks @mbfrahry! Don't forget that there are a few other PRs before this one, I've just based the work on this one on them as they have functionality that I need, so once those are reviewed and merged this PR and subsequent ones will shrink quite a bit :slightly_smiling_face:

PR #179, opened 2017-09-26
r/folder: Resource rewrite

vancluever:
`vsphere_folder` has now been re-written, with the new version of the resource supporting all 5 major folder types, renaming, moving, and tagging. Documentation has also been fleshed out, and state migration is supported.

This work was undertaken as it was a little tricky to work with tags on the current incarnation of `vsphere_folder` resource, but also takes care of some work that has been pending for a long time to get `vsphere_folder` to support more than just virtual machines.


vancluever:
Thanks @mbfrahry!

I am actually adding a bit of code to `Delete` to prevent removal if child items exist in the folder, as vSphere will actually destroy all of these when `Destroy` is called if it can (including VMs :fearful:)

Will push in a little bit and might need eyes on it again!

PR #181, opened 2017-09-29
r/virtual_machine: Network fixes

### LINKED ISSUE #161: panic: reflect.Set: value of type mo.DistributedVirtualSwitch is not assignable to type mo.VmwareDistributedVirtualSwitch
terraform crashes with the message above if using a specific network label. Using the standard "VM Network"  label works fine.

### Terraform Version
Terraform v0.10.5

Vsphere plugin version:
 terraform providers
.
└── provider.vsphere ~> 0.3

### Affected Resource(s)

- vsphere_virtual_machine

```hcl
provider "vsphere" {
  user           = "myuser"
  password       = "mypassword"
  vsphere_server = "myurl.com"
  version = "~> 0.3"
  # if you have a self-signed cert
  allow_unverified_ssl = true
}

resource "vsphere_virtual_machine" "web" {
  name   = "terraform-web"
  vcpu   = 2
  memory = 4096
  cluster = "mycluster"
  network_interface {
    label = "mynet"                   
    ipv4_address = "192.168.11.160"
    ipv4_prefix_length="24"
    ipv4_gateway = "192.168.11.1"
  }

  disk {
    datastore = "mydatastore"
    template = "mytemplate"
  }

}

```

### Debug Output
https://gist.github.com/hlima/b6931ad3347fcf1e313dc4ec8b0bd25c#file-debug-output

### Panic Output
https://gist.github.com/hlima/0edaca6f7b7774fcafcd4a3c24a696ce
### Expected Behavior

Specified a network label. Expected the VM to be created and assign to that network.

### Actual Behavior

terraform crashed.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids
Using a  network label that maps into a  VmwaRE Distributed Virtual switch ( vcenter tags as VmwareDistributedVirtualSwitch) casuses the crash. Using the standard "VM Network" label things work fine
Trying to create a VM with the same network label using govc also crashes with similar  output

### LINKED ISSUE #165: * vsphere_virtual_machine.vm: network not found
_This issue was originally opened by @sidhurana as hashicorp/terraform#16111. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,

Terraform v0.10.5

Issue: While trying to create VM, network-label seems creating problems. And issue is that my network label contains "space and forward slash as well" i.e. v910 10.33.156.0/24

 #govc ls network
/testing/network/v910 10.33.156.0/24

* vsphere_virtual_machine.vm: network '*v910 10.33.156.0/24' not found. 
Let me know if you need more info.

Is it possible to get it working with special characters within network label ? 

Regards,
Sidh

vancluever:
An update of govmomi corrects a couple of recent issues with handling networks in `vsphere_virtual_machine`:

Fixes #165, which was an issue where network searches in govmomi were
not working for networks with unescaped slashes in their names. We also
have a test for this now as well where we reproduce by adding a
host-level portgroup, which the API does not escape slashes for.

Fixes #161, which was an issue with network searches where panics could
happen on trying to load a network backed by a 3rd party hardware DVS.
The latest govmomi has commits that properly use base types for the
respective items.


vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -2013,3 +2051,129 @@ resource "vsphere_virtual_machine" "vm" {
...
+		os.Getenv("VSPHERE_HOST_NIC0"),
+		os.Getenv("VSPHERE_CLUSTER"),
+		os.Getenv("VSPHERE_RESOURCE_POOL"),
+		testAccResourceVSphereVirtualMachineSlashNetLabel,
+		os.Getenv("VSPHERE_IPV4_ADDRESS"),
+		os.Getenv("VSPHERE_IPV4_PREFIX"),
+		os.Getenv("VSPHERE_IPV4_GATEWAY"),
+		os.Getenv("VSPHERE_DATASTORE"),
+		os.Getenv("VSPHERE_TEMPLATE"),
+		os.Getenv("VSPHERE_USE_LINKED_CLONE"),
```
tombuildsstuff: _minor/not a blocker_ feels like this lot would make a nice test helper object

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -2013,3 +2051,129 @@ resource "vsphere_virtual_machine" "vm" {
...
+    ipv4_gateway       = "${var.ipv4_gateway}"
+  }
+
+  disk {
+    datastore = "${var.datastore}"
+    template  = "${var.template}"
+    iops      = 500
+  }
+
+  depends_on = ["vsphere_host_port_group.pg"]
```
tombuildsstuff: I'd have thought this should be implied by `network_interface.0.label` being set to `"${vsphere_host_port_group.pg.0.name}"`?

vsphere/resource_vsphere_virtual_machine_test.go
vancluever: Yeah, things are getting a little ridiculous here :wink:

Definitely could be something we could assemble into an object, although I'm a little resistant to the idea of breaking the variable configuration off, which is kind of what this would do, as right now all config fixtures are pretty easy to read and look like real Terraform config rather than chunks, which was a problem with this test code before (in a different way, though, one that made a lot less sense than what would be accomplished with a proper test fixture generator. :slightly_smiling_face: )

vsphere/resource_vsphere_virtual_machine_test.go
vancluever: We wait on all resources here because while we are only referencing the first resource for the network name, the VM could be created on any of the three hosts, hence it's important to ensure that the network is available on all three before proceeding with creation of the VM.

vancluever:
Thanks for the review @tombuildsstuff! Merging now

PR #184, opened 2017-09-30
add initial support for importing compute clusters and modifying DRS

jgensler8:


jgensler8:
I am not sure I can find time to finish this soon. I figured I would open a PR to gain visibility if someone was thinking of implementing this. Unfortunately, I am not able to create or delete clusters on the vSphere server I have access to. I have written an import function for others that might have a similar limitation.

I started writing another resource to apply DRS suggestions but I am not sure what the interface looks like other than an "assertion" resource (there ought to be zero recommendations). The goal here was a Partially Automated DRS cluster where I control when the recommendations are run.

vancluever:
Hey @jgensler8 - thanks for the PR!

Compute clusters are on our roadmap, probably after v1.0.0 goes out sometime in November.

I will keep this PR open here just so that we have the code as a reference or in case someone else wants to pick up the work. :+1:  on supporting import too! Documentation would be great :wink:

Applying DRS recommendations sounds like an interesting thing, and possibly something that I might be keeping in mind when we start work on the new `vsphere_virtual_machine` resource, as this sounds like a natural fit for `Refresh` behaviour that would change the host that a specific system is currently homed to so that it can be vMotioned on update. More specifically I think it's a great candidate for hashicorp/terraform#14887, as this behaviour is definitely more diff-related than it is refresh related. Regardless, I think that `vsphere_virtual_machine` is the right place for this as we want to make sure DRS recommendations are things that we process for resources that Terraform actually manages, and a resource that manages this OOB of the lifecycle of the virtual machine would ultimately conflict once the new resource is live in a few months. Hence, I would hold off on doing any major work on such a resource.

PS: If you need a lab to develop against, you can set up a lab pretty "easy" if you can get your hands on vSphere doing nested ESXi on a single server node (this is actually what our lab is here). There is also the govmomi [simulator](https://github.com/vmware/govmomi/tree/master/simulator), but YMMV on that one.

Cheers!

vancluever:
Hey @jgensler8 - just noticed that you have been updating this PR.

Just FYI - now that #244 is completed, we will be setting eyes on this resource and a couple of others after getting v1.0.0 out. If you are still planning on working on this, I have a few requests:

* Rename the resource to `vsphere_compute_cluster`. This is what I have envisioned the name to look like internally, which will match with a planned `vsphere_storage_cluster` resource as well.
* Have you read the [cluster config spec](https://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/vim.cluster.ConfigSpec.html) yet and aggregated the options yet? I would recommend doing this, and also taking in the UI into account, so the implementation is the best of both worlds.

If you could let me know how much you will be able to work on this as well going forward over the next month or so, that would be great. v1.0.0 is a few weeks out still, but after that, I will be looking at this resource a bit more in depth.

Thanks!

vancluever:
Hey @jgensler8, just FYI that we will be focusing on this for 1.5.0. More than likely we will take over development of this resource to get it completed at a standard we are comfortable with (but will build off of this work so contributions are preserved).

Apologies for the radio silence and thanks for the work here!

jgensler8:
No worries, @vancluever. This has totally dropped off my radar.

I don't think I'll find time to work on this resource and I think I'll close out this PR and wait for the feature to be added later.

Thank you again for your communication along the way! You are doing great work with this provider!

vancluever:
No problem @jgensler8 and thank you for the kind words! :slightly_smiling_face:

Issue #24, opened 2017-06-13
terraform 0.6.16 - Vmware Vsphere Provider - domain saved as "vsphere.local" in some network configurations


hashibot:
_This issue was originally opened by @totojack as hashicorp/terraform#6658. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

terraform 0.6.16
### Affected Resource(s)
- vsphere_virtual_machine
### Terraform Configuration Files

```
variable "vsphere_user" {}
variable "vsphere_password" {}
variable "vsphere_server" {}
variable "vsphere_cluster" {}
variable "vsphere_datacenter" {}
variable "vsphere_domain" {}
variable "vsphere_dns" {}
variable "vsphere_timezone" {}
provider "vsphere" {
user = "${var.vsphere_user}"
password = "${var.vsphere_password}"
vsphere_server = "${var.vsphere_server}"
allow_unverified_ssl = true
}
resource "vsphere_virtual_machine" "web" {
name = "terraform-web"
folder = "${vsphere_folder.testfolder.path}"
vcpu = 2
memory = 2048
datacenter = "${var.vsphere_datacenter}"
cluster = "${var.vsphere_cluster}"
domain = "my.custom.domain.com"
dns_servers = ["${var.vsphere_dns}"]
#resource_pool = ""
time_zone = "${var.vsphere_timezone}"

network_interface {
label = "virtualwire_Template"
ipv4_address = "172.16.1.19"
ipv4_prefix_length = "28"
ipv4_gateway = "172.16.1.17"
}

disk {
template = "TEMPLATE/Centos7_tmpl_20160422_01"
datastore = "VPLEX-SAS"
}
disk {
size = "20"
datastore = "VPLEX-SAS"
}
disk {
size = "10"
datastore = "VPLEX-SAS"
}
}
```
### Debug Output

https://gist.github.com/totojack/4db892bf6d9b03b3b982401fedd8c2f6
### Expected Behavior

domain value saved in files:
- /etc/resolv.conf , with line "search my.custom.domain.com"
- /etc/sysconfig/network-scripts/ifcfg-eth0 , with line "domain my.custom.domain.com"
### Actual Behavior

domain value saved as default value 'vsphere.local' in files:
- /etc/resolv.conf, with line "search vsphere.local"
[...]

PR #185, opened 2017-10-02
WIP: add domain as dns suffix option

mjrider:
default virtual machines get customized with vsphere.local as dns lookup suffix. While most of the time there is a better alternative, being the domain name of the VM. This adds a check for this alternative

This was suggested a while ago and lived as a stash in my git folder.
rebased it, and pushed it.

I can test it tomorrow at work.


vancluever:
This looks like the related issue: #24

vancluever:
Hey @mjrider - smoke tested this, all looks good! Just adding a doc update to it as well and will be merging shortly.

Thanks for the contribution!

PR #188, opened 2017-10-08
New resource: vsphere_distributed_virtual_switch (refactor)

vancluever:
This commit is a major refactor of the DVS resource provided in PR #135.

More options have been added, the DVS has been fixed so that it uses the
VMware DVS object instead of the generic DVS object, which is more or
less just used for 3rd party DVS switches. This meant that the DVS was
missing a number of features that more than likely would have been
expected in the DVS.

The breadth of the features added can be seen in the documentation. It
should be noted that this does not include the ability to set private
VLANs, filtering policies, or port mirroring sessions, which will be
added at a later date and pending business case.

This work also contains a bunch of up front work that will be necessary
for the `vsphere_distributed_port_group` resource, which will be following
not too long after this resource.


vancluever:
Thanks @mbfrahry!

PR #191, opened 2017-10-10
Updated docs to show permissions needed to perform tagging operations.

deberon:
It appears that starting with the 0.3.0 release tagging permissions are required when creating (at least) a folder, whether you set tags or not.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -128,6 +128,9 @@ vCenter must be assigned the following privileges:
    - Inventory (all)
    - Provisioning (all)
 
+* vCenter Inventory Service
+   - vCenter Inventory Service Tagging (all)
+
 * License 
    - Add License
    - Remove License


vancluever:
Hey @deberon, can you send along the error that you are getting?

This does not _seem_ correct to me, as the tagging endpoint should only be loaded if tags are required. It could be something that needs to be fixed in the tagging code.

PS: That document will be undergoing a rewrite in a few weeks ahead of the v1.0.0 release, at which point the "required permissions part" may be going away, or at least will be consistent with the rest of the providers (ie: we don't necessarily indicate every IAM permission needed in the AWS provider, so we won't be including every permission needed for the vSphere provider).

Thanks!

deberon:
You bet. Here's the main.tf I used to troubleshooting this:

```hcl
provider "vsphere" {
    user = "${var.vsphere_username}"
    password = "${var.vsphere_password}"
    vsphere_server = "${var.vsphere_server}"

    version = "0.4.0"

    allow_unverified_ssl = true
}

data "vsphere_datacenter" "datacenter" {
    name = "${var.vsphere_datacenter}"
}

resource "vsphere_folder" "folder" {
    path = "folder_test"
    // Start 0.4.0
    datacenter_id = "${data.vsphere_datacenter.datacenter.id}"
    type = "vm"
    // End 0.4.0

    // Start 0.3.0
    //datacenter = "${var.vsphere_datacenter}"
    // End 0.3.0
}
```

And here's what happens when I run an apply:

```
data.vsphere_datacenter.datacenter: Refreshing state...
vsphere_folder.folder: Creating...
  datacenter_id: "" => "** PARANOID REDACTION **"
  path:          "" => "folder_test"
  type:          "" => "vm"
Error applying plan:

1 error(s) occurred:

* vsphere_folder.folder: 1 error(s) occurred:

* vsphere_folder.folder: error reading tags: Get unexpected status code: 403: call failed: Error response from vCloud Suite API: {"value":{"messages":[]},"type":"com.vmware.vapi.std.errors.unauthorized"}

```

I'm running vCenter v6.0.0. Also I typo'd the version in my initial comment. It appears this starts happening in 0.4.0, not 0.3.0.

deberon:
Also it should be noted that the folder gets created correctly.

vancluever:
Thanks @deberon - will take a look soon!

vancluever:
Hey @deberon - I'm having a hard time reproducing this with just folder permissions alone. 

One thing - can you try with v0.4.1? There was a bug that was fixed in `vsphere_folder` with migration of the state for the folder resource that might be causing trouble for you too. I see that you have the version fully locked in your config.

vancluever:
Hey @deberon - I'm closing this one now as this seem to be more a bug than anything else right now. When you get around to responding, can you do so on the new issue?

Thanks!

vancluever:
Hey @deberon, as per #194 I am merging this one in after all and we will be making a note about needing tagging privileges when we make the update to the docs for v1.0.0.

Thanks again and sorry for the confusion!

PR #193, opened 2017-10-10
created adapter_type variable for network interfaces

shelson:
I needed this functionality and noticed that it was already commented with "TODO". It's my first foray into go so the following might be better worked out another way:

adapter_type_string := fmt.Sprintf("%T", device)
if adapter_type_string == "*types.VirtualE1000" {

Otherwise though it seems to work just fine.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -36,6 +36,11 @@ var DiskControllerTypes = []string{
 	"ide",
 }
 
+var NetworkAdapterTypes = []string{
+	"vmxnet3",
+	"e1000",
+}
+
 type networkInterface struct {
 	deviceName       string
 	label            string
@@ -45,7 +50,7 @@ type networkInterface struct {
 	ipv6Address      string
 	ipv6PrefixLength int
 	ipv6Gateway      string
-	adapterType      string // TODO: Make "adapter_type" argument
+	adapterType      string
 	macAddress       string
 }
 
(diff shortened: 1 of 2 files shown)


vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -382,6 +387,21 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							ForceNew: true,
+							Default: "vmxnet3",
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
```
vancluever: You can use [`StringInSlice`](https://github.com/hashicorp/terraform/blob/master/helper/validation/validation.go#L74) from the `validation` package in TF core to make this much easier on the eyes.

The syntax would be here:

`ValidateFunc: validation.StringInSlice(virtualMachineNetworkAdapterTypeAllowedValues, false)`

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -36,6 +36,11 @@ var DiskControllerTypes = []string{
 	"ide",
 }
 
+var NetworkAdapterTypes = []string{
```
vancluever: This variable should be unexported (starts with a lowercase letter, ensures that no package outside of `vsphere` can use it - basically a private variable), and be specific enough to avoid collisions.

Suggestion: `virtualMachineNetworkAdapterTypeAllowedValues`.

Also, you might want to check [here](https://godoc.org/github.com/vmware/govmomi/vim25/types) to see if there is any appropriate constants you can use.

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1144,6 +1167,12 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 	log.Printf("[DEBUG] Device list %+v", deviceList)
 	for _, device := range deviceList {
 		networkInterface := make(map[string]interface{})
+		adapter_type_string := fmt.Sprintf("%T", device)
```
vancluever: Here, rather than printing the type out, you should be using the more spec-native [type assertion](https://tour.golang.org/methods/15).

The code below would look like this:

```go
if _, ok := device.(*types.VirtualE1000); ok {
  networkInterface["adapter_type"] = "e1000"
} else {
  networkInterface["adapter_type"] = "vmxnet3"
}
``` 

You can also use the [`switch`](https://tour.golang.org/methods/16) type assertion format as well.

vancluever:
Hey @shelson, thanks for the update!

Next issue: the tests are failing because you haven't run the changes through `go fmt`.

Depending on the IDE/editor you are using this is set up different ways - if you can tell me I might be able to help.

Also see here: https://blog.golang.org/go-fmt-your-code

shelson:
Thanks for the awesome feedback - I really appreciate you taking the time to help me get this in a shippable state.

I'll work on the go-fmt thing in the morning as it's getting late here now.

Thanks again.

vancluever:
LGTM to me now @shelson! Merging now. Thanks for the contribution!

PR #196, opened 2017-10-10
tags: Ensure complete removal of tags is working

vancluever:
`tagsClientIfDefined` was checking to see if the tags client was defined
as a condition to loading up the tags client for `Create` and `Update`
operations. While this generally worked for adding the tags to new
resources and changing tags when necessary, it was not working for
complete tag removal due to its reliance on `GetOk`, which fails if a
field is removed from config completely.

This fixes `tagsClientIfDefined` so that it checks both the old and new
incarnations of the `tags` argument to make sure that both are empty
before passing on loading the tags client.

### DIFF
--- a/vsphere/helper_test.go
+++ b/vsphere/helper_test.go
@@ -243,6 +243,27 @@ func testObjectHasTags(s *terraform.State, client *tags.RestClient, obj object.R
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


vancluever:
Thanks @mbfrahry!

PR #198, opened 2017-10-11
r/virtual_machine: Don't attempt to discover gateway on power off

vancluever:
The new gateway discovery code inadvertently caused a regression in the
fixes that were made to prevent TF from crashing when a VM was powered
off. This fixes that by skipping the GW discovery altogether when there
is no guest device information to use for the association.


vancluever:
Thanks @mbfrahry!

Issue #183, opened 2017-09-30
vsphere terraform 0.10.6 issue


rismoney:
Using the vsphere module to create a simple vm, 
I am on v.0.10.6

This looks very similar to an issue (I'm on windows) but it was reported a few hours ago, and was referred to digitalocean provider-
https://github.com/hashicorp/terraform/issues/16220

```
Error applying plan:

1 error(s) occurred:

* module.igen-nj-wks01.vsphere_virtual_machine.wks: 1 error(s) occurred:

* vsphere_virtual_machine.wks: Post https://tp-nj-vc-01.mydom.corp/sdk: EOF

```

Trace log showS:
```
2017/09/29 17:35:05 [TRACE] dag/walk: vertex "root", waiting for: "meta.count-boundary (count boundary fixup)"
2017/09/29 17:35:05 [TRACE] dag/walk: vertex "provider.vsphere (close)", waiting for: "module.igen-nj-wks01.vsphere_virtual_machine.wks"
2017/09/29 17:35:07 [TRACE] root.igen-nj-wks01: eval: *terraform.EvalWriteState
2017/09/29 17:35:07 [TRACE] root.igen-nj-wks01: eval: *terraform.EvalApplyProvisioners
2017/09/29 17:35:07 [TRACE] root.igen-nj-wks01: eval: *terraform.EvalIf
2017/09/29 17:35:07 [TRACE] root.igen-nj-wks01: eval: *terraform.EvalWriteState
2017/09/29 17:35:07 [TRACE] root.igen-nj-wks01: eval: *terraform.EvalWriteDiff
2017/09/29 17:35:07 [TRACE] root.igen-nj-wks01: eval: *terraform.EvalApplyPost
2017/09/29 17:35:07 [ERROR] root.igen-nj-wks01: eval: *terraform.EvalApplyPost, err: 1 error(s) occurred:

* vsphere_virtual_machine.wks: Post https://tp-nj-vc-01.mydom.corp/sdk: EOF
2017/09/29 17:35:07 [ERROR] root.igen-nj-wks01: eval: *terraform.EvalSequence, err: 1 error(s) occurred:

* vsphere_virtual_machine.wks: Post https://tp-nj-vc-01.mydom.corp/sdk: EOF
2017/09/29 17:35:07 [TRACE] [walkApply] Exiting eval tree: module.igen-nj-wks01.vsphere_virtual_machine.wks
[...]

PR #199, opened 2017-10-11
r/virtual_machine: Allow tuning of customization waiter

vancluever:
The current value is hardcoded at 10 minutes, which has been possibly
causing some issues. This allows for the modification of the timeout -
so that it can be change to be a later value if customization takes a
long time (possible on systems with older hardware or non-SSD disk), or
disabled altogether to eliminate it as a source of problems.

Note that disabling the waiter pushes waiting over to the guest
networking waiter, which is currently also hardcoded at 5 minutes.
Modification of this attribute to a different value will require a state
migration (to translate `true` => `5min` and `false` => `<disabled>`) which is a
bit more of an involved change, so I want to take a wait-and-see
approach to see if it is required, especially considering the resource
will be undergoing a larger refactor very soon.

Ref: #183

### DIFF
--- a/vsphere/event_helper.go
+++ b/vsphere/event_helper.go
@@ -17,10 +17,6 @@ const (
 	eventTypeCustomizationSucceeded = "CustomizationSucceeded"
 )
 
-// virtualMachineCustomizationWaiterTimeout is the default time that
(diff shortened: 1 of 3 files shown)


vancluever:
Thanks @mbfrahry!

PR #201, opened 2017-10-12
New data source: vsphere_network

vancluever:
This introduces the `vsphere_network` data source, which allows one to
look up a general network in vSphere via a path.

This will be useful in upcoming resources that will require a network
managed object ID, versus a network name or path, effectively moving all
of the "heavy lifting" in translating a network from a human-readable
name to an object over to this data source, versus being handled in the
resource itself (which will ultimately just fetch the network by its
MOID, without worrying about the details underneath).

### DIFF
--- a/vsphere/data_source_vsphere_network.go
+++ b/vsphere/data_source_vsphere_network.go
@@ -1,0 +1,52 @@
+package vsphere
+
+import (
+	"fmt"
+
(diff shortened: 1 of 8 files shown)


vancluever:
Thanks @catsby! :slightly_smiling_face:

PR #202, opened 2017-10-12
r/distributed_port_group: Use MOID as resource ID

vancluever:
The additions that are in #201 dictate the basis for how we will be
working with vSphere networks in the future - any resource that requires
a network for a backing (ie: virtual ethernet cards, VMkernel nics, etc)
will require a MOID to some sort of network, versus the name itself,
allowing us to circumvent actually searching for the network in the
downstream resource itself. Searching would be circumvented altogether
in the case of DVS portgroups being managed by TF completely as we would
just pass the MOID to the downstream resource.

The current behaviour of the `vsphere_distributed_port_group` resource is
to use the `key` attribute in the MO as the ID of the resource, which is
documented in the API as the UUID of the DVS port group, although in
reality it actually seems to be the MOID, which allows #201 to work
without issue right now. However, in order to guarantee that things will
be stable in the long run, we need to be sure we are using the right
values, so this update changes the ID to be the MOID of the object
itself. `key` is now being exported as a computed attribute of the same
name.

### DIFF
--- a/vsphere/resource_vsphere_distributed_port_group.go
+++ b/vsphere/resource_vsphere_distributed_port_group.go
@@ -12,9 +12,15 @@ import (
 func resourceVSphereDistributedPortGroup() *schema.Resource {
 	s := map[string]*schema.Schema{
 		"distributed_virtual_switch_uuid": {
-			Type:     schema.TypeString,
-			Required: true,
-			ForceNew: true,
+			Type:        schema.TypeString,
+			Description: "The UUID of the DVS to attach this port group to.",
+			Required:    true,
+			ForceNew:    true,
+		},
+		"key": {
+			Type:        schema.TypeString,
+			Description: "The generated UUID of the portgroup.",
+			Computed:    true,
 		},
 		// Tagging
 		vSphereTagAttributeKey: tagsSchema(),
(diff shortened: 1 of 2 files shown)


vsphere/resource_vsphere_distributed_port_group.go
```diff
@@ -85,17 +92,18 @@ func resourceVSphereDistributedPortGroupRead(d *schema.ResourceData, meta interf
 	if err := validateVirtualCenter(client); err != nil {
 		return err
 	}
-	dvsID := d.Get("distributed_virtual_switch_uuid").(string)
 	pgID := d.Id()
-	pg, err := dvPortgroupFromUUID(client, dvsID, pgID)
+	pg, err := dvPortgroupFromMOID(client, pgID)
```
catsby: The method signature of `dvPortgroupFromMOID` changes here, but I don't see that in the changes here. Was it changed elsewhere, and this resource currently doesn't work without this change?

vancluever: Not just the signature, but the method as well :slightly_smiling_face:

`dvPortgroupFromUUID` vs `dvPortgroupFromMOID`

catsby: ohisee

vancluever:
Thanks @catsby!

PR #204, opened 2017-10-13
Add vsphere_folder required param in index example

Ketouem:


vancluever:
Hey @Ketouem, thanks for the edit!

We will be reviewing the examples in the index (and the index in general) ahead of the v1.0.0 release, but for now this is a good small update. :+1:

PR #212, opened 2017-10-18
Add Windows VM configuration options to resource_vsphere_virtual_machine

mahcsig:
I have added options for the AutoLogon and FirstLogonCommands for Windows guest customization. 
I also made the registered user and organization configurable.
I was going to add in the ChangeSID option, but it seems to always change the SID whether the option is passed or not, and whether a True or False are passed, so I left it out.

I would like to clean up the code at 2340 of `vsphere/resource_vsphere_virtual_machine.go` but I couldn't find a way to not pass in the `guiRunOnce` with no passed in `run_once` commands and not have it fail.  Passing an empty array to `CustomizationSysprep.GuiRunOnce`  fails on the vSphere side. Any tips to fix that workaround would be appreciated.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -75,6 +75,7 @@ type windowsOptConfig struct {
 	domainUserPassword string
 	fullName           string
 	orgName            string
+	runOnce            []string
 }
 
 type cdrom struct {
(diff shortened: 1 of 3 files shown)


dhekimian:
@vancluever Pinging you to check on the process to get this merged.

Please Advise.

vancluever:
Hey @dhekimian and @mahcsig,

Sorry for not getting to this when it was first submitted. As a part of the coming changes to the resource in general, we have added a more comprehensive list of customization options. Included with this is the above windows configuration options (we even expose the field in the customization spec that allows you to supply your own answer file).

Since the new feature set covers these options, for clarification sake I'm going to close out this PR. The new resource is nearly complete and will be in review soon, and should be out in the next few weeks.

Thanks, sorry for the delay, and thanks for your contribution @mahcsig!

Issue #120, opened 2017-08-08
Feature Request: Manage Custom Attributes for VMS (annotations)


rismoney:
Recently added was an annotation capability https://github.com/terraform-providers/terraform-provider-vsphere/pull/111.  

It would be great if we can create Custom attributes.  Custom attributes apply to all machines in the inventory.  

I would think the proper behavior here is:
1. if the attribute doesn't exist create it.  (On VMs this seems to be global)
2. if the attribute exists (name), and it's value is different update it
3. if the attribute exists (name), and it's value is the same noop it.

Welcome thoughts.

vancluever:
Hey @rismoney, this is a great suggestion! Will add it to the TODO.

Just did a bit of research on this one and I think that we could just implement most of the functionality that is in the [fields](https://github.com/vmware/govmomi/tree/master/govc/fields) module in the govc reference implementation. 

Your idea on how the CRUD would work for this pretty much matches what the normal expected operation of TF, so we are on the same page here.

rismoney:
I think this can be safely closed, as there is solid attribute support in the provider at this time.

vancluever:
Thanks @rismoney! This is definitely dated now that we have #229. Glad it's working for you now!

PR #229, opened 2017-10-31
Add Custom Attribute resource and data source

rmbrad:
This allows management of Custom Attributes and allows their values to be specified on VMs satisfying #120. Specifically it adds a `vsphere_custom_attribute` data source for using predefined attributes, a `vsphere_custom_attribute` resource for managing new custom attributes using Terraform, and a `custom_attributes` argument on VM resources for specifying custom attribute values.

Still need to add documentation and tests, which I can work on if you're interested in merging.

### DIFF
--- a/vsphere/internal/helper/customattribute/custom_attributes_helper.go
+++ b/vsphere/internal/helper/customattribute/custom_attributes_helper.go
@@ -1,0 +1,154 @@
+package customattribute
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/viapi"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// ConfigKey is the string that should always be used as the
+// key for specifying custom attribute values for a resource. Various resource
+// custom attribute helpers will depend on this value being consistent across
+// resources.
+//
+// When adding custom attributes to a resource schema, the easiest way to do
+// that (for now) will be to use the following line:
+//
+//   customattribute.ConfigKey: customattribute.ConfigSchema(),
+//
+// This will ensure that the correct key and schema is used across all
+// resources.
+const ConfigKey = "custom_attributes"
+
+// ConfigSchema returns the schema for custom attribute configuration
+// for each resource that needs it.
+//
+// The key should be set to the ConfigKey constant and should be a
+// map of custom attribute ids to values.
+func ConfigSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:        schema.TypeMap,
+		Description: "A list of custom attributes to set on this resource.",
+		Optional:    true,
+	}
+}
+
+func VerifySupport(client *govmomi.Client) error {
+	if err := viapi.ValidateVirtualCenter(client); err != nil {
+		return errors.New("Custom attributes are only supported on vCenter")
+	}
+	return nil
+}
+
+func IsSupported(client *govmomi.Client) bool {
+	return VerifySupport(client) == nil
+}
+
+func ReadFromResource(client *govmomi.Client, entity *mo.ManagedEntity, d *schema.ResourceData) {
+	customAttrs := make(map[string]interface{})
+	if len(entity.CustomValue) > 0 {
+		for _, fv := range entity.CustomValue {
+			value := fv.(*types.CustomFieldStringValue).Value
+			if value != "" {
+				customAttrs[fmt.Sprint(fv.GetCustomFieldValue().Key)] = value
+			}
+		}
+	}
+	d.Set(ConfigKey, customAttrs)
+}
+
+type CustomAttributeDiffProcessor struct {
+	// The field manager
+	fm *object.CustomFieldsManager
+
+	// Old map of custom attribute key to values
+	oldAttributes map[string]interface{}
+
+	// New map of custom attribute key to values
+	newAttributes map[string]interface{}
+}
+
+func (p *CustomAttributeDiffProcessor) clearRemovedAttributes(subject object.Reference) error {
+	for k := range p.oldAttributes {
+		_, ok := p.newAttributes[k]
+		if !ok {
+			key, err := strconv.ParseInt(k, 10, 32)
+			if err != nil {
+				return err
+			}
+			err = p.fm.Set(context.TODO(), subject.Reference(), int32(key), "")
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (p *CustomAttributeDiffProcessor) setNewAttributes(subject object.Reference) error {
+	for k, v := range p.newAttributes {
+		key, err := strconv.ParseInt(k, 10, 32)
+		if err != nil {
+			return err
+		}
+		err = p.fm.Set(context.TODO(), subject.Reference(), int32(key), v.(string))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (p *CustomAttributeDiffProcessor) ProcessDiff(subject object.Reference) error {
+	if err := p.clearRemovedAttributes(subject); err != nil {
+		return fmt.Errorf("error clearing removed attributes for object ID %q: %s", subject.Reference().Value, err)
+	}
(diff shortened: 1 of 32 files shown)


vancluever:
Wow @rmbrad, this would be definitely something that I think we would be interested in - it would be a great compliment to the current tagging support! :+1:

I will hold off on my review until you have time to finish the tests and the docs. :slightly_smiling_face:

rmbrad:
@vancluever Finally got around to finishing off the documentation and tests. Let me know if there's anything I need to update after you get a chance to review.

vancluever:
Hey @rmbrad - just had a minute to go over the docs now. I noticed that the only resource to have custom attribute added to it so far has been virtual machines. Can you work to add the support to other resources?

It might help to spin off the common functionality like we do with tags - see [`tags_helper.go`](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/tags_helper.go).

Once #244 is in we will need to re-visit adding this to the new VM resource, so this will probably go in sometime after that gets merged.

Thanks!

rmbrad:
Hi @vancluever. No problem, I'll start adding support for other resources this week.

vancluever:
@rmbrad I just did a quick once-over on your progress and I just wanted to say it's looking awesome! :+1:

Looks like you are making some pushes still so I'll wait till it's all settled before I start the review - let me know when you are ready!

vancluever:
PS: One thing I did want to say/ask - have you tested this at all on ESXi? I have to double check, but if custom attributes are not supported on ESXi (as in the same fashion as tags), we need to make sure we skip this functionality and prohibit its use when connected to ESXi directly. There is a small helper in `internal/helper/viapi` that can be used to detect non-vCenter connections and send a generic error message.

Looking forward to getting this in!

rmbrad:
@vancluever Thanks! I just need to add support for the datastore resources to achieve parity with the tag support. I've been running all tests against vCenter and not ESXi directly, but will be setting up a new ESXi instance for testing the VMFS datastore support; so that will allow me to verify whether ESXi supports custom attributes or not.

I'm hoping to have everything ready for review by early next week, at this point it's only the testing that takes time as all logic is now in the helper.

rmbrad:
Hey @vancluever - I believe this PR should now be ready for review. I've added support for custom attributes to all resources that supported tagging. Custom attributes are unsupported on direct ESXi connections, so I added the necessary checks similar to what is done for tagging support.

Let me know if there's anything else needed.

vsphere/internal/helper/customattribute/custom_attributes_helper.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/viapi"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// vSphereCustomAttributeKey is the string that should always be used as the
```
vancluever: Looks like this needs to be changed to match the name of the identifier:

```
// ConfigKey is the string...
```

vsphere/internal/helper/customattribute/custom_attributes_helper.go
```diff
@@ -0,0 +1,153 @@
...
+// When adding custom attributes to a resource schema, the easiest way to do
+// that (for now) will be to use the following line:
+//
+//   vSphereCustomAttributeKey: customAttributesSchema(),
+//
+// This will ensure that the correct key and schema is used across all
+// resources.
+const ConfigKey = "custom_attributes"
+
+// customAttributesSchema returns the schema for custom attribute configuration
```
vancluever: Same here

```
// ConfigSchema returns the schema...
```

vsphere/resource_vsphere_datacenter_test.go
```diff
@@ -81,6 +81,35 @@ resource "vsphere_datacenter" "testDC" {
...
 
+const testAccCheckVSphereDatacenterConfigAttributes = `
+resource "vsphere_custom_attribute" "terraform-test-attribute" {
+  name                = "terraform-test-attribute"
+  managed_object_type = "Datacenter"
+}
+
+resource "vsphere_datacenter" "testDC" {
+  name = "testDC"
+  custom_attributes = "${map(vsphere_custom_attribute.terraform-test-attribute.id, "value")}"
```
vancluever: I notice the pattern of using the `map` interpolation function is present in tests quite a bit. I'm not too sure if this works, but have you just tried:

```hcl
resource "vsphere_datacenter" "testDC" {
  name = "testDC"
  custom_attributes = {
    "${vsphere_custom_attribute.terraform-test-attribute.id}" = "value"
  }
}
```

I would prefer that if it works - leaving more complicated scenarios using interpolation to any individual practitioner's use case :slightly_smiling_face:

vsphere/resource_vsphere_distributed_port_group_test.go
```diff
@@ -171,6 +171,53 @@ func TestAccResourceVSphereDistributedPortGroup(t *testing.T) {
 				},
 			},
 		},
+		{
+			"single_attribute",
```
vancluever: I think we need to be more descriptive in the sub-tests here - _tag_ is descriptive enough, but _attribute_ has meaning in TF itself - ie: a _resource attribute_. Maybe something like "single custom attribute", etc.

website/docs/r/datacenter.html.markdown
```diff
@@ -46,6 +46,12 @@ The following arguments are supported:
 ~> **NOTE:** Tagging support is unsupported on direct ESXi connections and
 requires vCenter 6.0 or higher.
 
+* `custom_attributes` - (Optional) Map of custom attribute ids to value
```
vancluever: Can we have a link to directions on how to apply custom attributes similar to what we have for tags [here](https://www.terraform.io/docs/providers/vsphere/r/tag.html#using-tags-in-a-supported-resource)?

website/docs/d/custom_attribute.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+
+```hcl
+data "vsphere_custom_attribute" "attribute" {
+  name = "terraform-test-attribute"
+}
+```
+
+## Argument Reference
+
+* `name` - (String, required) The name of the custom attribute.
```
vancluever: We recently standardized on a new naming convention for attributes that I want to make sure is followed - the [VM resoruce](https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#argument-reference) has a good layout of the format. An example is below:

> `some_attr` - (Required or Optional) Does some thing. Forces a new resource if changed (if this actually is a `ForceNew`) argument.

vsphere/resource_vsphere_datacenter_test.go
rmbrad: Yeah, use of the `map` function bothers me as well; but unfortunately interpolation doesn't work for map keys in resources as mentioned here https://github.com/hashicorp/terraform/issues/14516#issuecomment-301630345 :disappointed:.

One work around I often use is to place the definition of the custom attribute map in a locals block, where map keys are interpolated (keys will also be interpolated when passing to a module) as follows:
```hcl
locals {
  dc_attrs = {
    "${vsphere_custom_attribute.terraform-test-attribute.id}" = "value"
  }
}

resource "vsphere_datacenter" "testDC" {
  name = "testDC"
  custom_attributes = ${local.dc_attrs}
}
```
If you would prefer I can update the tests to work that way.

rmbrad:
Hi @vancluever, I've rebased the PR on master and believe I have addressed all your comments, with exception of the use of the `map` function I commented on above. I also added a test to the `vsphere_folder` resource that removes all custom attribute values. Let me know if find any other issues. Thanks!

vsphere/resource_vsphere_datacenter_test.go
vancluever: I think this looks a lot cleaner :slightly_smiling_face: and we also use the map assignment pattern in other places like you type there (not necessarily in the vSphere provider, but elsewhere in TF-land).

vancluever:
@rmbrad I've done my final review and just make a couple of documentation corrections, but it largely looked good!

I had a second thought about the tests regarding keeping the `map` interpolation function as I actually think that looks better after all - but I noticed you changed the tests already. That's fine - we can change it on a refactor if it rubs anyone the wrong way later on. I want to keep the convention of using `map` for examples and stuff though, and it looks like you've kept that, so we are good.

Anyway, thank you very much for this - this is a great PR and we appreciate the contribution! Well done!

PR #236, opened 2017-11-10
Update directions to network interface block

genevievelesperance:


vancluever:
Hey @genevievelesperance, thanks (again) for keeping eyes on the docs!

As mentioned, #244 is making big changes to the resource. This includes the `network_interface` section as well. If you are curious about how the new resource looks or how it's documented, check the PR - the docs are linked there as well if you want to provide feedback on them.

Thanks for looking at this and sorry for the confusion!

PR #237, opened 2017-11-10
bootable_vmdk_path renamed to vmdk in vm docs

genevievelesperance:


vancluever:
Hey @genevievelesperance, thanks for keeping eyes on the docs!

We have just put in #244 which includes significant changes to the `disk` sub-resource and its docs, so this information will not apply anymore. Sorry for the confusion, but thank you very much again for looking at this!

PR #241, opened 2017-11-16
Disk config name and vmdk path are not compatible

christianang:


vancluever:
Hey @christianang, with #244 going in soon, We are actually removing the double options in favor of a single `name` field. The docs are changing quite drastically, so this will not apply.

Thanks for having eyes on the documentation though!

PR #242, opened 2017-11-17
Fixes #213 by allowing spaces in the datastore name using regexp.

### LINKED ISSUE #213: Error on terraform apply with datastores with spaces in the name
### Terraform Version

Run `terraform -v` to show the version. If you are not running the latest
version of Terraform, you can try upgrading, depending on if your problem is
with the provider or with Terraform core itself.

Terraform v0.10.7

### Affected Resource(s)
- vsphere provider

### Description
I recall this being an issue with an older version of the vsphere provider; but it had been fixed. Similar to https://github.com/terraform-providers/terraform-provider-vsphere/issues/52 except the VM is actually created correctly, terraform just spits out an error at the end of a terraform apply operation.

`[ERROR] Failed trying to parse disk path: [Compellent VMWare 5] centos-7-test/centos-7-test.vmdk`

### Expected Behavior

`terraform apply` should have created a single VM.

### Actual Behavior

`terraform apply` creates the VM but has an error at the end of the operation regarding being unable to parse the disk path.

### Steps to Reproduce
1. Create a VMWare datastore with spaces in the name and reference it in your config file
2. `terraform apply`

tigelane:


vancluever:
Hey @tigelane, thanks for the PR.

We have been working on a new version of the `vsphere_virtual_machine` resource that moves the functionality of parsing the composite datastore/path file names into the [DataStorePath](https://godoc.org/github.com/vmware/govmomi/object#DatastorePath) object in govmomi.

Looking at the code for [`FromString`](https://github.com/vmware/govmomi/blob/master/object/datastore_path.go#L33), this should ultimately fix the issue by moving to upstream behaviour.

Going to close this one considering. Thanks for taking the time to look into it however!

PR #248, opened 2017-11-22
d/network: Remove vCenter requirement

vancluever:
This resource has always functioned correctly on ESXi, and will be
required to work with virtual machines going forward, so it's important
it works on both ESXi and vCenter.

Specifically tested this on ESXi to make sure. Also added skips for all
the DVS related tests in this data source so that the host/standard port
group example is the only one that runs when conducting ESXi tests.


vancluever:
Just merging this in now. All this does is remove a minor restriction and does not change functionality.

Issue #243, opened 2017-11-17
Support VAppConfig for vsphere_virtual_machine resource


rowanjacobs:
Hi,

We'd like to automate deploying a template that has some VAppConfig values, including IP, DNS server, gateway, etc. set, but it looks like the `vsphere_virtual_machine` resource currently doesn't support configuring those values.

We'd like to PR this support into the provider. We were thinking the config in the resource might look like this:
```hcl
resource "vsphere_virtual_machine" "some-virtual-machine" {
  name     = "some-virtual-machine"
  hostname = "some-host"
  folder   = "some-folder"
  vcpu     = 2
  memory   = 4096

  app_config {
    ip = "10.10.10.10"
    dns = "8.8.8.8"
    admin_password = "some-password"
  }

...
}
```

We wanted to get some feedback on this before we start working on the PR; is this UI acceptable?  any concerns about this that we might not be aware of?

cc @christianang

vancluever:
Hey @rowanjacobs - thanks for checking in before starting work on this!

I have an incoming PR for the new `vsphere_virtual_machine` resource. Part of this work is a large re-organization of higher-level VM lifecycle config stuff like cloning and customization to dedicated blocks like you have have laid out above, so I think the config structure is on the right track.

However, I'm curious to the value of these would have when there is no current support for vApps within Terraform. This sounds like it might be setting some vApp options but not really doing anything outside of that. Just quickly looking at the docs [here](https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.vm_admin.doc/GUID-33840994-E5D4-4746-AC7C-359411239FD3.html), it looks like editing vApp config options just simply assists with when you want to export the VM, which is a process that might be better suited for something like [Packer](https://www.packer.io/) or some other tool that is designed to build applications based on virtual machines, versus their deployment.

If you could reply with more of your use case, or if I'm missing something, we can see if this would be a good fit for the VM resource or not.

Thanks!

rowanjacobs:
Our main use case is to deploy an existing template with configurable IP address, gateway, DNS, and NTP options. 

Having spiked out some changes to the provider to add values specific to our vSphere cluster for those vAppConfig options, we were able to get the template working after `terraform apply` without any additional configuration. I admit I don't understand that much about vApps but that seems to me like it's doing more than just assisting export.

vancluever:
[...]

PR #257, opened 2017-11-30
helper/envbrowse: QueryConfigOptionEx -> QueryConfigOption

vancluever:
QueryConfigOptionEx is not available on vSphere 5.5, hence releasing
with our environment browser implementation using this call would have
effectively deprecated/terminated support for this version of vSphere.

* For default devices, using guest ID does nothing - it was wrongly
assumed that this helped give back a hardware list specific to a guest
OS type, but it just restricts the guest OS descriptors returned. The
Key parameter has been retained as it's a part of the QueryConfigOption
parameter list as an optional field.
* For looking up Guest OS families by Guest ID (example: ubuntu64Guest ->
Linux, windows9Server64Guest -> Windows), rather than using
QueryConfigOptionEx to filter the results, we just iterate through the
whole descriptor list returned by QueryConfigOption until we find our
guest ID. This helps get rid of some length assertion code as well.

Reference: #243 

Will paste acceptance tests when done.

PR #258, opened 2017-11-30
r/virtual_machine: API backwards compatability

vancluever:
Ref: #243 

This PR is focusing on addressing compatibility issues, namely with vSphere 5.5.

Thank you very much to @rowanjacobs and @genevievelesperance for their help with testing on older versions!

### DIFF
--- a/vsphere/internal/helper/computeresource/compute_resource_helper.go
+++ b/vsphere/internal/helper/computeresource/compute_resource_helper.go
@@ -137,11 +137,6 @@ func BasePropertiesFromReference(client *govmomi.Client, ref types.ManagedObject
 
 // DefaultDevicesFromReference fetches the default virtual device list for a
 // specific compute resource from a supplied managed object reference.
-//
-// The current implementation of this method takes a single guest ID, which
-// ultimately filters down to a QueryConfigOptionEx call with only the single
-// guest ID set. This ensures that the VirtaulDeviceList returned matches a
-// default set best suited for the guest type that is being created.
 func DefaultDevicesFromReference(client *govmomi.Client, ref types.ManagedObjectReference, guest string) (object.VirtualDeviceList, error) {
 	log.Printf("[DEBUG] Fetching default device list for object reference %q for OS type %q", ref.Value, guest)
 	props, err := BasePropertiesFromReference(client, ref)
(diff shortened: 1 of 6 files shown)


rowanjacobs:
The "efiSecureBootEnabled" tag seems to also be a problem.

We're currently running with the `BootOptions` schema property commented out and it's looking promising so far; will update when the test finishes running.

vancluever:
:+1: awesome!

In the meantime I've added a gate for that. Looks like this affects vSphere 6.0 as well as it's only available on 6.5 and higher. I am going to smoke test this on our lab, which is 6.5.

vancluever:
OK, smoke test now works against the latest commit - EFI secure boot gets set properly (also discovered it needed to have `restart_required` flagged to modify so made sure that change is in there).

@rowanjacobs @genevievelesperance would love to know if this works for you on your 5.5 installation!

genevievelesperance:
Hey! We have gotten to a different error but probably something in our own configuration. We will investigate more in the morning:

```
--- FAIL: TestAccResourceVSphereVirtualMachine_Basic (690.39s)
    --- FAIL: TestAccResourceVSphereVirtualMachine_Basic/basic (331.28s)
    	testing.go:503: Step 0 error: Error applying: 1 error(s) occurred:

    		* vsphere_virtual_machine.vm: timeout waiting for a routeable interface
    --- FAIL: TestAccResourceVSphereVirtualMachine_Basic/shutdown_OK (333.05s)
    	testing.go:503: Step 0 error: Error applying: 1 error(s) occurred:

    		* vsphere_virtual_machine.vm: timeout waiting for a routeable interface
    --- FAIL: TestAccResourceVSphereVirtualMachine_Basic/clone_from_template (26.06s)
    	testing.go:503: Step 0 error: Error applying: 1 error(s) occurred:

    		* vsphere_virtual_machine.vm: error sending customization spec: Customization of the guest operating system 'debian7_64Guest' is not supported in this configuration. Microsoft Vista (TM) and Linux guests with Logical Volume Manager are supported only for recent ESX host and VMware Tools versions. Refer to vCenter documentation for supported configurations.
    	testing.go:563: Error destroying resource! WARNING: Dangling resources
    		may exist. The full state and error is shown below.

    		Error: Check failed: vsphere_virtual_machine.vm not found in state
```

vancluever:
Hey @genevievelesperance yeah a couple of things you want to keep in mind, or some items from our own setup:

* We use Ubuntu templates for cloning and customization testing (might be an issue there with Debian guests specifically)
* For the non-clone tests, you want to make sure that you are booting those on a PXE network that can boot a VM into a full OS environment that has VMware tools installed. We use RancherOS using a custom build of the initrd from my PR here: https://github.com/rancher/os/pull/2138

PS: I'm closing this in favour of #259, which contains all of the commits from #257 and #258 and then a couple of more changes (namely folder support fixes and also a bit of a refactor of the version check for disk sharing just to be consistent). Can you try your tests against that branch, just so that we are all on the same page?

Thanks!

PR #261, opened 2017-12-01
vendor: Update validation to 5da62a44f

vancluever:
This fixes the documentation of StringMatch so that it's consistent with
its implementation.

Very minor, but it's just a loose end I'd like to be in 1.0.0. This will
be updated along with all of the TF deps on the next release this is
merged in.


vancluever:
Thanks all!

Issue #268, opened 2017-12-04
Create VM ESXi 6.0.0 and 6.5.0 not possible on standalone license!


tommymunter:
Hi there,
Great work with new provider.vsphere v1.0.1 but it's not working on ESXi standalone :(
Only working with vcenter, Essentials Plus license. 
Or temporary in Evaluation Mode.

### Terraform Version
Terraform v0.11.1

### vSphere Provider Version
+ provider.vsphere v1.0.1

### Affected Resource(s)
- `vsphere_virtual_machine`

### Terraform Configuration Files

```hcl
data "vsphere_datacenter" "dc" {
  name = "ha-datacenter"
}

data "vsphere_datastore" "datastore" {
  name          = "datastore1"
  datacenter_id = "${data.vsphere_datacenter.dc.id}"
}

data "vsphere_resource_pool" "pool" {}

data "vsphere_network" "mgmt_lan" {
  name          = "MGMT LAN"
  datacenter_id = "${data.vsphere_datacenter.dc.id}"
}

resource "vsphere_virtual_machine" "test2" {
  name             = "test2"
  resource_pool_id = "${data.vsphere_resource_pool.pool.id}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"

  num_cpus   = 1
  memory     = 2048
  wait_for_guest_net_timeout = 0
  guest_id = "centos7_64Guest"
  nested_hv_enabled =true
  network_interface {
   network_id     = "${data.vsphere_network.mgmt_lan.id}"
   adapter_type   = "vmxnet3"
   use_static_mac = true
   mac_address    = "00:50:56:17:01:01"
  }

  disk {
   size             = 16
   name             = "test2.vmdk"
   eagerly_scrub    = false
   thin_provisioned = true
  }
}

```
### Expected Behavior

Create a new VM

### Actual Behavior

* vsphere_virtual_machine.test2: error shutting down virtual machine: ServerFaultCode: Current license or ESXi version prohibits execution of the requested operation.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

We have Standalone ESXi 6.0.0 u2 and ESXi 6.5.0, same behavior on both

Workaround:
remove license key, then it's possible to apply vm

Is the "Essentials Plus" license required????
I thought it should be possible to use the standalone license!

vancluever:
Hey @tommymunter, thanks for the report.

[...]

PR #270, opened 2017-12-04
docs: Free ESXi is not supported

vancluever:
This is something that has potentially come up, had to do a bit of
looking into it myself as it's been a while since I've used free ESXi
and sure enough API "write" operations are not supported, hence the
provider is not really usable on such environments.


vancluever:
Ref: #268 and also https://www.virtuallyghetto.com/2011/06/dreaded-faultrestrictedversionsummary.html for more info.

PR #273, opened 2017-12-05
d/resource_pool: Make standalone ESXi section a bit clearer

vancluever:
Just add a little language so that it's easier to notice.

### DIFF
--- a/website/docs/d/resource_pool.html.markdown
+++ b/website/docs/d/resource_pool.html.markdown
@@ -41,10 +41,10 @@ data "vsphere_resource_pool" "pool" {
 }
 ```
 
-### Using with ESXi
+### Using with Standalone ESXi
 
-On ESXi, you don't have to specify either attribute to use this data source. An
-empty declaration will load the default resource pool.
+When using ESXi without vCenter, you don't have to specify either attribute to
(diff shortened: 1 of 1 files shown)


website/docs/d/resource_pool.html.markdown
```diff
@@ -41,10 +41,10 @@ data "vsphere_resource_pool" "pool" {
 }
 ```
 
-### Using with ESXi
+### Using with Standalone ESXi
 
-On ESXi, you don't have to specify either attribute to use this data source. An
-empty declaration will load the default resource pool.
+When using ESXi without vCenter, you don't have to specify either attribute to
+use this data source. An empty declaration will load the default resource pool.
```
tombuildsstuff: from a UI / consistency point of view for this kind of information - this may probably make more sense as a `NOTE` i.e.:

```
~> **Note: Usage with Standalone ESXi** When using ESXi without vCenter, you don't have to specify either attribute to use this data source. An empty declaration will load the default resource pool.
```

vancluever: I like this - will make the change!

vancluever:
Thanks for the review @tombuildsstuff! Made the change and am merging now

PR #276, opened 2017-12-06
Use subresourceTypeCdrom for cdrom schema update

palesiak:
While creating a virtual CDROM and attaching it, I was getting a handful of errors that seemed unrelated to the network_interface:

```
* vsphere_virtual_machine.orchestrator-2: 1 error(s) occurred:

* vsphere_virtual_machine.orchestrator-2: Invalid address to set: []string{"network_interface", "0", "path"}

```

Digging through the code, I found a schema update that was using the `subresourceTypeNetworkInterface` type instead of `subresourceTypeCdrom`. After using this newly built provider plugin, I'm able to create the instance without issue


vancluever:
Ouch! Thank you very much @palesiak! Will merge this and release shortly

PR #277, opened 2017-12-06
r/virtual_machine: Validate exact disk type when using linked_clone

### LINKED ISSUE #275: r/virtual_machine: Enforce disk provisioning options on linked clones
Looks like we have an issue in which if a VM is cloned via linked_clone, and the disk options are set to something incorrect of what the source template is, we end up with an inconsistent state.

The first apply fails with something like:

```
* vsphere_virtual_machine.foo: disk.0: cannot change the value of "thin_provisioned" - (old: false new: true)
```

Trying to run any plans after this will give you:

```
* vsphere_virtual_machine.foo: vsphere_virtual_machine.foo: network_interface.0: cannot find network device: invalid ID ""
```

The VM is not able to be deleted in this state and will need to be removed from vSphere and Terraform state manually.

What we need to do is:

* Validate the template's `disk` options match the options of the source disk
* Possibly expose these options via the `vsphere_virtual_machine` data source, so that they can be accessed in the same way we expose the disk sizes.

vancluever:
A VM that's cloned with `linked_clone` and different disk mode options
that what is set on the source template will end up with inconsistent
settings post-clone, causing a read-update failure that will leave the
VM in a state that is irrecoverable to Terraform - the only way to
rectify the situation is to delete the VM and start over with a new
state.

This adds extra validation to make sure that `thin_provisioned` and
`eagerly_scrub` are set to the same settings as the source template's
counterpart disk, along with validating that the disk is the exact same
size (something that wasn't done yet).

Fixes #275.

-----

PS: There will be a larger-ish PR (mainly because of adjustments to tests) that will follow this one to add `thin_provisioned` and `eagerly_scrub` attributes to the `vsphere_virtual_machine` data source. This will help ensure that these can be fetched in the same way we have the data source fetching `disk_sizes` currently.

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
@@ -555,7 +555,7 @@ nextNew:
 // This function is meant to be called during diff customization. It is a
 // subset of the normal refresh behaviour as we don't worry about checking
 // existing state.
-func DiskCloneValidateOperation(d *schema.ResourceDiff, c *govmomi.Client, l object.VirtualDeviceList) error {
+func DiskCloneValidateOperation(d *schema.ResourceDiff, c *govmomi.Client, l object.VirtualDeviceList, linked bool) error {
 	log.Printf("[DEBUG] DiskCloneValidateOperation: Checking existing virtual disk configuration")
 	devices := selectDisks(l, d.Get("scsi_controller_count").(int))
 	// Sort the device list, in case it's not sorted already.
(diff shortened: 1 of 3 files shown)


vancluever:
Thanks @tombuildsstuff! No regressions in a number of existing clone tests

```
=== RUN   TestAccResourceVSphereVirtualMachine
=== RUN   TestAccResourceVSphereVirtualMachine/clone_from_template
=== RUN   TestAccResourceVSphereVirtualMachine/clone,_multi-nic_(template_should_have_one)
=== RUN   TestAccResourceVSphereVirtualMachine/clone_with_different_timezone
=== RUN   TestAccResourceVSphereVirtualMachine/clone_with_bad_timezone
=== RUN   TestAccResourceVSphereVirtualMachine/clone_with_different_hostname
=== RUN   TestAccResourceVSphereVirtualMachine/clone_with_extra_disks
=== RUN   TestAccResourceVSphereVirtualMachine/clone_with_cdrom
--- PASS: TestAccResourceVSphereVirtualMachine (908.33s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone_from_template (124.53s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone,_multi-nic_(template_should_have_one) (122.93s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone_with_different_timezone (120.07s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone_with_bad_timezone (0.01s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone_with_different_hostname (120.03s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone_with_extra_disks (294.76s)
    --- PASS: TestAccResourceVSphereVirtualMachine/clone_with_cdrom (126.01s)
```

PR #303, opened 2017-12-08
r/virtual_machine: add vApp properties

rowanjacobs:
In order to support configurable vApp properties on VMs cloned from an OVF (#243), we have added a `vapp` subresource with a `properties` key-value map. In order to run acceptance tests, the template needs to have vApp properties with the keys `example_key` and `another_key`, whose initial and default values should be blank. 

Note that we haven't been able to successfully implement zeroing out a property.  This seems to likely be an issue with the vSphere API or client.  But we think this should not block adding this feature in the mean time.

cc @mcwumbly @desmondrawls @genevievelesperance @evanfarrar


charandas:
@rowanjacobs I am working with a coreos ova with Terraform using your fork.

Pertinent excerpt from my ova:

```xml
<Property ovf:userConfigurable="true" ovf:type="string"
          ovf:key="guestinfo.coreos.config.data" ovf:value="">
  <Label>CoreOS config data</Label>
  <Description>Inline cloud-config data</Description>
</Property>
<Property ovf:userConfigurable="true" ovf:type="string"
          ovf:key="guestinfo.coreos.config.data.encoding" ovf:value="">
  <Label>CoreOS config data encoding</Label>
  <Description>Encoding for cloud-config data</Description>
</Property>
```

Pertinent Terraform configuration:

```
data "ct_config" "worker" {
  content      = "${data.template_file.container_config.rendered}"
  pretty_print = false
}

  name             = "terraform-test"
  resource_pool_id = "${data.vsphere_resource_pool.pool.id}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"
  vapp {
    properties {
      "guestinfo.coreos.config.data" = "${base64encode(data.ct_config.worker.rendered)}"
      "guestinfo.coreos.config.data.encoding" = "base64"
    }
  }
}
```

The Terraform output on `apply` contains below, so the values do make it through:
```
  vapp.0.properties.%: "" => "2"
  vapp.0.properties.guestinfo.coreos.config.data:  "" => "some_legal_base64"
  vapp.0.properties.guestinfo.coreos.config.data.encoding: "" => "base64"
```

The VM is created, but it never ends up getting those vApp properties set. I verify it by failing to ssh into it, as my `coreos.config.data` contains my ssh public key. I can also verify in vSphere client, that the generated VM does not have any vApp properties set.

Am I doing something wrong?  Thought I could benefit from a second pair of eyes.

rowanjacobs:
I've verified that setting `vapp.properties.guestinfo.coreos.config.data` in  `coreos.config.data` when cloning a VM sets the vApp property on the VM cloned from the CoreOS OVA. Are you using the `clone` setting? If not, how are you building the VM from the CoreOS OVA?

Here's what the VM configuration I'm using looks like:
```
  name             = "terraform-test"
  resource_pool_id = "${data.vsphere_resource_pool.pool.id}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"

  num_cpus = 2
  memory   = 2048
  guest_id = "${data.vsphere_virtual_machine.template.guest_id}"

  network_interface {
    network_id   = "${data.vsphere_network.network.id}"
    adapter_type = "${data.vsphere_virtual_machine.template.network_interface_types[0]}"
  }

  disk {
    name = "terraform-test.vmdk"
    size = "${data.vsphere_virtual_machine.template.disks.0.size}"
  }

  vapp {
    properties {
      "guestinfo.hostname" = "khaleesi-terraform-test"
      "guestinfo.coreos.config.data" = "some guest data"
    }
  }

  clone {
    template_uuid = "${data.vsphere_virtual_machine.template.id}"
  }
}
```

charandas:
I am on vSphere ESXi 6.0 though. Are you on 6.5?

rowanjacobs:
I'm testing against vSphere 5.5.

charandas:
screen shot 2017-12-13 at 3 03 35 pm
screen shot 2017-12-13 at 3 04 34 pm

charandas:
I had missed the file changes in vsphere/virtual_machine_config_structure.go when building last night. Sorry for the false alarm. It works now. The reason I reverted to manually copying changes from your fork, is because of the following:

```
# would not build for me, with go 1.9.2
❯ go get -u github.com/rowanjacobs/terraform-provider-vsphere
# github.com/rowanjacobs/terraform-provider-vsphere
serve/gopath/src/github.com/rowanjacobs/terraform-provider-vsphere/main.go:10:15: cannot use vsphere.Provider (type func() "github.com/terraform-providers/terraform-provider-vsphere/vendor/github.com/hashicorp/terraform/terraform".ResourceProvider) as type "github.com/rowanjacobs/terraform-provider-vsphere/vendor/github.com/hashicorp/terraform/plugin".ProviderFunc in field value

# I basically tried to get it to build by copying over these changes
# into  github.com/terraform-providers/terraform-provider-vsphere in my gopath
```

vancluever:
Hey @rowanjacobs, just wanted to say I have been watching this and it looks promising - I will give this a much closer look later this week!

One thing that I did notice when I looked earlier - does this support _removal_ of the vApp properties? It looks like another attribute needs to be set to make this happen (`vAppConfigRemoved`). I'm kind of on the fence of how I would prefer this to be implemented as from what I remember, wiping the config removes them for good, and I don't know how easy they would be to re-construct after the fact. I'm wondering if maybe we can support that and source the proper settings from somewhere else, like a template?

In any case - will look later this week as mentioned.

tevaum:
Just want to thank you guys for working on this stuff... I use terraform with vsphere to deploy some images that doesn't support guest customization and this is the only way to go for me, so I'm very excited  seing you guys working on it!

vancluever:
Hey @rowanjacobs, I'm about to give this a spin to see how it goes. I'll also see if I can solve the config zeroing problem as well.

Re: What I mentioned about removing vApp config, I don't think we need to concern ourselves with this right now 100%. A VM's OVF configuration is a complex thing that I think that we might not even want to touch with TF, treating vApp and OVF template configuration as kind of a template that that is out of our scope to manage, but rather something that would be up to something like Packer to configure as part of the image building process.

Anyway, will update soon!

vancluever:
Hey @rowanjacobs, alright, finished checking, looks like there is an issue here.

Here is the `vapp` snippet of my config using the CoreOS OVA:

```
  vapp {
    properties {
      "guestinfo.hostname"                        = "${var.vm_name}.foobar.local"
      "guestinfo.interface.0.name"                = "ens192"
      "guestinfo.interface.0.ip.0.address"        = "172.16.1.150/24"
      "guestinfo.interface.0.route.0.gateway"     = "172.16.1.1"
      "guestinfo.interface.0.route.0.destination" = "0.0.0.0/0"
      "guestinfo.dns.server.0"                    = "172.16.1.1"
    }
  }
```
This is not writing a correct configuration. I think the best way to illustrate it is via how it seems to be overwriting some of the config parameters. Here is an example of the first three from the MOB.

The first parameter - `guestinfo.hostname` - seems to be writing correctly:

**Before**
hostname-before

**After**
hostname-after

However, the next couple get odd looking. This is the next parameter, the full cloud-config data, which should have not been touched. However if you look at the second picture here closely, you'll see that the key has been overwritten with the `guestinfo.dns.server.0` key, which is empty:

**Before**
cloud-config-before

**After**
cloud-config-after

Finally, the last example overwrites the cloud-config URL property with the key and value for `guestinfo.hostname`.

**Before**
url-before

**After**
url-after


Finally, there is a non-empty diff on the next plan:

```
  ~ vsphere_virtual_machine.foo
      vapp.0.properties.guestinfo.dns.server.0:                    "" => "172.16.1.1"
      vapp.0.properties.guestinfo.interface.0.ip.0.address:        "" => "172.16.1.150/24"
      vapp.0.properties.guestinfo.interface.0.route.0.destination: "" => "0.0.0.0/0"
      vapp.0.properties.guestinfo.interface.0.route.0.gateway:     "" => "172.16.1.1"

```

Looking deeper, I'm guessing there has to be something up with the way the `VAppPropertySpec`/`ArrayUpdateSpec` is being built to have these kinds of conflicts in them and it needs to be fixed. 

Also thinking about this a bit more, and given the potential for corruption here: I think we should be a bit more absolute when working with the vApp settings - we should only allow keys that are _already_ in the vApp configuration (ie: are already valid parameters in the template). Also, we should make sure that we are saving non-zero values for any keys that are a part of the vApp configuration but are _not_ defined in Terraform config (this way they can be properly zeroed out on the next run).

I'll add some notes in the code in the next couple of hours.

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -192,6 +199,21 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 	}
 }
 
+// VAppSchema represents the schema for the vApp sub-resource.
+//
+// This sub-resource allows the customization of vApp properties
+// on cloned VMs.
+func VAppSubresourceSchema() map[string]*schema.Schema {
```
vancluever: I think we could probably just put this directly into `schemaVirtualMachineConfigSpec` as it's so small, along with the `vapp` key. I'm almost tempted to also just say we flatten it and name it `vapp_properties`, but we might need to add more keys at a later time.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -836,6 +836,45 @@ func TestAccResourceVSphereVirtualMachine(t *testing.T) {
 				},
 			},
 		},
+		{
+			"clone with vapp properties",
```
vancluever: Some thoughts on testing this while enforcing present keys only - we might need to test this against a CoreOS template or something else well known, and work on setting well-known cloud-config keys.

vsphere/virtual_machine_config_structure.go
```diff
@@ -508,6 +509,117 @@ func flattenExtraConfig(d *schema.ResourceData, opts []types.BaseOptionValue) er
...
+			props = append(props, prop)
+		} else if isOld {
+			prop := types.VAppPropertySpec{
+				ArrayUpdateSpec: types.ArrayUpdateSpec{
+					Operation: types.ArrayUpdateOperationEdit,
+				},
+				Info: &types.VAppPropertyInfo{
+					Key:   int32(i) + 1,
+					Id:    k,
+					Value: "",
```
vancluever: Just noticed this here - the original comment said that there was problems zeroing out a property - is this code non-functional, or were you referring to something else (maybe deleting a key completely)?

vsphere/virtual_machine_config_structure.go
```diff
@@ -508,6 +509,117 @@ func flattenExtraConfig(d *schema.ResourceData, opts []types.BaseOptionValue) er
...
+		return nil
+	}
+	props := config.GetVmConfigInfo().Property
+	if len(props) < 1 {
+		// No props to read is a no-op
+		return nil
+	}
+	vac := make(map[string]interface{})
+	vApp := d.Get("vapp").([]interface{})
+	if len(vApp) > 0 && vApp[0] != nil {
```
vancluever: As mentioned in the comments before the change request, I think we need to add config values that have non-zero values set on them as well, that way, on the next apply, they get zeroed out and any defaults apply. That way, removing a key always removes the corresponding value from config, even if someone adds OOB changes.

vsphere/virtual_machine_config_structure.go
```diff
@@ -508,6 +509,117 @@ func flattenExtraConfig(d *schema.ResourceData, opts []types.BaseOptionValue) er
...
+			allIds = append(allIds, k)
+		}
+	}
+	sort.Sort(allIds)
+
+	for i, k := range allIds {
+		_, isOld := oldMap[k]
+		_, isNew := newMap[k]
+
+		if isNew {
```
vancluever: Annotating as I'm guessing it would be around here - just noting that we will want to somehow check to ensure the key is actually present in the vApp config before working with it. This may be here, or in another part of the function, or in another function altogether (maybe it actually is part of diff customization so that we can keep the method signature clean here and as consistent as possible with the rest of the expander/flattener logic).

vsphere/virtual_machine_config_structure.go
```diff
@@ -508,6 +509,117 @@ func flattenExtraConfig(d *schema.ResourceData, opts []types.BaseOptionValue) er
 	return d.Set("extra_config", ec)
 }
 
+// expandVAppConfig reads in all the vapp key/value pairs and returns
+// the appropriate VmConfigSpec.
+//
+// We track changes to keys to determine if any have been removed from
+// configuration - if they have, we add them with an empty value to ensure
+// they are removed from vAppConfig on the update.
+func expandVAppConfig(d *schema.ResourceData) types.BaseVmConfigSpec {
```
vancluever: `BaseVmConfigSpec` is an interface, but we aren't returning anything else other than `VmConfigSpec` so this is abstracting things in a way that they don't need to be abstracted. Can this be changed to return a `*types.VmConfigSpec`?

rowanjacobs:
Thanks for getting back to us with your feedback! Unfortunately I'm about to go on vacation and the rest of the team will be short-staffed next week, so it's unlikely we'll be able to make required changes until the beginning of January. We'll definitely get to it when we can!

vsphere/virtual_machine_config_structure.go
rowanjacobs: Done.

vsphere/resource_vsphere_virtual_machine_test.go
rowanjacobs: I've started testing against the CoreOS OVA instead of against a custom-built template.

vsphere/virtual_machine_config_structure.go
rowanjacobs: Although it changes the signature, I decided to check if the key is present in this function because it does clean up the logic of the function significantly, now that we no longer have to merge two different lists of vApp properties and can rely on the canonical list of properties from the template.

vsphere/virtual_machine_config_structure.go
rowanjacobs: The issues with zeroing out properties are due to the fact that at least in version 5.5, the vSphere API doesn't actually do anything when it receives an empty string for a vApp property value. We've gotten around this by changing the empty string to " ", and the corresponding portion of the new code should work properly now.

vancluever:
Hey @rowanjacobs - just FYI I will probably be reviewing this next week.

Thanks!

vancluever:
Hey all, looks like this is coming along great. A few things are still outstanding, save maybe documentation (still haven't looked at this part yet):

* Changing vApp properties should require a restart - this is to ensure that they will be sent along on the next boot (this will allow things like reconfiguration of IP addresses in CoreOS). You can do this by setting the `reboot_required` computed attribute when values have changed (there is a `getWithRestart` helper as well).
* While it looks like you can no longer add configuration keys to the vApp properties configuration, the operation still goes through in Terraform, and creates a perpetual diff as such. We need to ensure that these are blocked in the plan.

Do you know when you all will be able to work on these items? If it's going to be too much later, I can actually get these over the line for you all. We are actually doing some work here internally that could benefit from this and I'd be happy to finish it up. But let me know either way :slightly_smiling_face:

rowanjacobs:
I can definitely take care of the first change. I'll see if I can get to the second one this week, and if not then I'll pass it back to you since I'm going to be unavailable for a while starting next week.

vancluever:
Fair enough! Will keep an eye out :+1: 

This is looking very good by the way! I was excited to see a CoreOS VM come up purely from guestinfo!

rowanjacobs:
Thanks! Regarding the second change: should the plan fail and provide an error message indicating that you can't add new vApp property keys? Or should the new keys be silently removed from the plan? (Or, for my own reference, do you have an existing example somewhere in this or another provider that implements this in a preferred way?)

vancluever:
Hey @rowanjacobs sorry for missing this.

It should be set up so that TF explicitly fails with an error. We don't want silent behaviour as that gives the wrong impression to the user, and the bad keys risk ending up propagating to state, which would be bad as well.

rowanjacobs:
We were thinking of returning an error from `expandVAppConfig`, but none of the other expand functions, nor `expandVirtualMachineConfigSpec`, do this, so we'd have to change several function signatures. We can go ahead and add this to the PR; was this what you had in mind, or did you have another way you prefer of getting that error? We looked into adding a `ValidatorFunc` but it seems like that function wouldn't be able to access client or template information, and so wouldn't have access to the original list of vApp properties on the template.

rowanjacobs:
Here's an example of what the error looks like: ```
--- FAIL: TestAccResourceVSphereVirtualMachine_Basic (104.52s)
    --- FAIL: TestAccResourceVSphereVirtualMachine_Basic/clone_with_vapp_properties (104.52s)
    	testing.go:513: Step 2 error: Error applying: 1 error(s) occurred:

    		* vsphere_virtual_machine.vm: error expanding virtual machine config spec: Unsupported vApp properties in vapp.properties: [fake.property]
```

vancluever:
Hey @rowanjacobs, the error handling looks good for now. I want to maybe move this over to CustomizeDiff in the long run both so that the validation can be done at plan-time, and also so that we keep the expander/flattener interface clean, but that can be done later.

I'm reviewing this today - if it looks good, is it ready to merge?

Thanks!

vancluever:
Outstanding checklist:

- [ ] vApp property change requires a restart
- [ ] Block changes to unknown properties
- [ ] Check docs
- [ ] Check tests

vsphere/virtual_machine_config_structure.go
vancluever: Awesome! This is great info actually as I can now update some of my code on the work I was doing on govc. Good find! :+1:

vancluever:
FYI - this LGTM, I've made several updates and resolved the merge conflicts and I think we are in a good place now. I'm just waiting for the whole VM test suite to finish and then I'm merging!

Great job and hats off to everyone that made this happen!

PR #313, opened 2017-12-12
Update virtual_machine.html.markdown

jason-azze:
Fix a minor markdown typo where the `firmware` description was run-in with the description above it.


vancluever:
Good eye! Thanks @jason-azze!

PR #315, opened 2017-12-13
Change documentation of value for default clone timeout to 30 min.

jason-azze:
The optional timeout to wait for a clone to complete appears to be 30 minutes, not 10 minutes. I discovered this through direct observation.


vancluever:
Thanks again @jason-azze!

PR #318, opened 2017-12-13
r/virtual_machine: allow cloning VMs by template path rather than UUID

rowanjacobs:
Currently, resources and data sources which clone VMs or templates search for them by UUID. However, the vSphere `FindByUUID` API call does not return any results if the UUID is associated with a template rather than a VM.
(I haven't been able to find information on whether this behavior changed in vSphere 6.x, but it definitely occurs on vSphere 5.5, and I have seen reports of this behavior going back to 2006.)

In order to allow cloning templates rather than VMs, this PR introduces a helper function, `FromUUIDOrPath`, which first searches by UUID, then searches by path and datacenter. Two new optional properties in `clone`, `template_path` and `template_datacenter_id`, are added to support this workflow.


vancluever:
Hey @rowanjacobs, thanks for this.

Unfortunately, I don't think this is a good idea. The issue that you have described has only been reported on vSphere 5.5 - it is functional in 6.5, and we haven't received any reports of it happening on 6.0.

What I would recommend instead, if you want to work on a fix for this, is updating the cloning search process to allow for a fallback for a more general search of templates available in inventory in general and then comparing UUID from there. I was planning on looking to see about adopting an approach similar to how we do a general network search, using `ContainerView`. You can see the code [here](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/internal/helper/network/network_helper.go#L36).

Thanks and sorry for the inconvenience here!

rowanjacobs:
Thanks for the feedback! I'll go ahead and try to implement something based on `ContainerView` as you recommended.

genevievelesperance:
Hey Chris, since that code is from the govmomi examples, would it be okay to make the general templates search look like their vm search: https://github.com/vmware/govmomi/tree/master/examples/virtualmachines ?

vancluever:
@genevievelesperance I would almost go farther than that and just change over `FromUUID` to use that for everything, if it scales okay (remember that we don't necessarily require a "template" to be marked as such right now and that needs to stay that way).

I would be interested to see how each method acts at scale, so if there was a way we could set this up as an alternate method for now, spin up a sizeable amount of virtual machines (like 100-1000) and see how `go test -bench` behaves for each method, that would help us make a good decision on if we can just move it over wholesale or not.

_Or_, we just use the alternate form of `FromUUID` on ESXi 5.5 only:

```go
version := viapi.ParseVersionFromClient(client)
if version.Older(viapi.VSphereVersion{Product: version.Product, Major: 6}) {
  return fromUUIDContainerView(client, uuid)
}
```

genevievelesperance:
Yes, the latter was the approach we were going to take. We'll try it out and get back to you with something you can review.

nowhammies:
Just wanted to confirm I that I am unable to use a "marked as template" virtual machine in ESXI 6.0
Once I convert to a virtual machine things work as expected. I am using packer to provision these as templates with remote esxi and vsphere-template post processor.

Here is the govc about output

Name:         VMware vCenter Server
Vendor:       VMware, Inc.
Version:      6.0.0
Build:        5318203
OS type:      linux-x64
API type:     VirtualCenter
API version:  6.0
Product ID:   vpx
UUID:         15d87409-30cd-4865-9b65-437022dca8d7

PR #321, opened 2017-12-14
r/virtual_machine: Remove OOB deleted virtual machines from state

### LINKED ISSUE #299: Deleting resource manually breaks TF
Terraform 0.10.7
Provider:  Version 1.1.0

Reproducer:

1. run tf to clone windows vm.
1. manually Delete VM in Vsphere console.
1. run tf

Error message:
vsphere_virtual_machine.wks: Refreshing state... (ID: 422dbafe-5671-f3c6-132f-5e1ac75364e1)
Error refreshing state: 1 error(s) occurred:

* module.rismoney.vsphere_virtual_machine.wks: 1 error(s) occurred:

* module.rismoney.vsphere_virtual_machine.wks: vsphere_virtual_machine.wks: cannot locate virtual machine with UUID "422dbafe-5671-f3c6-132f-5e1ac75364e1": virtual machine with UUID "422dbafe-5671-f3c6-132f-5e1ac75364e1" not found

Expectation: TF should just recreate a new rismoney box, as actual doesn't match expected config.

Perhaps related to #271 ?

vancluever:
This was incorrectly changed from v0.4.2. While failing on read errors
is normal, it's a TF convention to check to see if the read error was
due to a resource being deleted outside of the TF workflow, and if it
has, mark that resource as gone by blanking out its ID.

This update restores that behavior to the VM resource. Other resources
in the provider will need to have this functionality added, although
most of those items are core infrastructure resources and are not as
ephemeral as a virtual machine is, so they are probably lower priority
to update. They will be done on an as needed basis.

Fixes #299.

### DIFF
--- a/vsphere/helper_test.go
+++ b/vsphere/helper_test.go
@@ -115,6 +115,17 @@ func copyStatePtr(t **terraform.State) resource.TestCheckFunc {
 	}
 }
 
+// copyState returns a TestCheckFunc that returns a deep copy of the state.
+// Unlike copyStatePtr, this state has de-coupled from the in-flight state, so
+// it will not be modified on subsequent steps and hence will possibly drift.
+// It can be used to access vales of the state at a certain step.
+func copyState(t **terraform.State) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		*t = s.DeepCopy()
+		return nil
+	}
+}
+
 // testGetPortGroup is a convenience method to fetch a static port group
 // resource for testing.
 func testGetPortGroup(s *terraform.State, resourceName string) (*types.HostPortGroup, error) {
@@ -342,6 +353,27 @@ func testDeleteVMDisk(s *terraform.State, resourceName string, name string) erro
 	return virtualdisk.Delete(tVars.client, p.String(), dc)
 }
 
+// testDeleteVM deletes the virtual machine. This is used to test resource
+// re-creation if TF cannot locate a VM that is in state any more.
+func testDeleteVM(s *terraform.State, resourceName string) error {
+	if err := testPowerOffVM(s, resourceName); err != nil {
+		return err
+	}
+	vm, err := testGetVirtualMachine(s, resourceName)
+	if err != nil {
+		return err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultAPITimeout)
+	defer cancel()
+	task, err := vm.Destroy(ctx)
+	if err != nil {
+		return fmt.Errorf("error destroying virtual machine: %s", err)
+	}
+	tctx, tcancel := context.WithTimeout(context.Background(), defaultAPITimeout)
+	defer tcancel()
+	return task.Wait(tctx)
+}
+
 // testGetTagCategory gets a tag category by name.
 func testGetTagCategory(s *terraform.State, resourceName string) (*tags.Category, error) {
 	tVars, err := testClientVariablesForResource(s, fmt.Sprintf("vsphere_tag_category.%s", resourceName))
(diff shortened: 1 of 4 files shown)


vancluever:
Test in question: 

```
=== RUN   TestAccResourceVSphereVirtualMachine
=== RUN   TestAccResourceVSphereVirtualMachine/re-create_on_deletion
--- PASS: TestAccResourceVSphereVirtualMachine (32.60s)
    --- PASS: TestAccResourceVSphereVirtualMachine/re-create_on_deletion (32.60s)
```

vsphere/helper_test.go
```diff
@@ -115,6 +115,17 @@ func copyStatePtr(t **terraform.State) resource.TestCheckFunc {
 	}
 }
 
+// copyState returns a TestCheckFunc that returns a deep copy of the state.
+// Unlike copyStatePtr, this state has de-coupled from the in-flight state, so
+// it will not be modified on subsequent steps and hence will possibly drift.
+// It can be used to access vales of the state at a certain step.
```
tombuildsstuff: `vales` -> `values`

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -21,6 +21,25 @@ import (
 
 var errGuestShutdownTimeout = errors.New("the VM did not power off within the specified amount of time")
 
+// UUIDNotFoundError is an error type that is returned when a
+// virtual machine could not be found by UUID.
+type UUIDNotFoundError struct {
+	s string
```
tombuildsstuff: worth making this `message` to be self-explanatory?

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -114,6 +114,41 @@ func TestAccResourceVSphereVirtualMachine(t *testing.T) {
...
+						Check: resource.ComposeTestCheckFunc(
+							copyState(&state),
+							testAccResourceVSphereVirtualMachineCheckExists(true),
+						),
+					},
+					{
+						PreConfig: func() {
+							if err := testDeleteVM(state, "vm"); err != nil {
+								panic(err)
+							}
```
tombuildsstuff: maybe it's just and Azure provider thing - but we'd generally make the deletion a Check function and assert it's been deleted, rather than crashing?

vancluever: `PreConfig` does not return error, so the only possible thing we can do is panic here.

vancluever: Update: The intention here is to make sure this gets handled _completely_ outside of the plan/apply testing cycle. `PreConifg` happens [very early on](https://github.com/hashicorp/terraform/blob/master/helper/resource/testing.go#L736), which is why I use it - `Check` on the other hand happens pretty much in the middle [after the apply](https://github.com/hashicorp/terraform/blob/master/helper/resource/testing_config.go#L75), but before the post-apply refresh to check for an empty plan. Having these OOB steps in `PreConfig` kind of gives them a more real-world workflow (ie: someone usually makes manual modifications _between_ a TF run, versus in the middle of one).

What would be nicer is if the signature of `PreConfig` was altered to send an error back, but that's kind of a breaking change.

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
vancluever: This is an internal field on a very small object - and this generally matches the pattern you see in something like the [`errors` package](https://golang.org/src/errors/errors.go), so I think it's fine :)

vsphere/helper_test.go
vancluever: Fixed!

PR #322, opened 2017-12-14
r/virtual_machine: Network allocation settings require vSphere >= 6.0

### LINKED ISSUE #286: Potential backwards compatibility issues with vSphere 5.5
While working on a PR for vApp config support (#243), we noticed a few problems in acceptance tests that may be related to backwards compatibility (#258) with vSphere version 5.5.

Since vSphere 5.5 does not support customization for Linux guests, we wrote an acceptance test that clones a simple vm template with Ubuntu pre-installed and skips customization. That test fails with the following error:

```
Error applying: 1 error(s) occurred:

    		* vsphere_virtual_machine.vm: timeout waiting for a routeable interface
```

However, looking at the vm in vcenter while the test is running we see that it receives an IP address. This makes us wonder whether:
1. We are actually failing to create a routeable network interface.
2. The network waiter is reporting false negative because it expects vSphere 6.x-specific information.
3. Our test setup is somehow wrong.

The relevant part of the terraform template for this test is:
```
  name             = "terraform-test"
  resource_pool_id = "${data.vsphere_resource_pool.pool.id}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"

  num_cpus = 2
  memory   = 2048
  guest_id = "${data.vsphere_virtual_machine.template.guest_id}"

  network_interface {
    network_id   = "${data.vsphere_network.network.id}"
    adapter_type = "${data.vsphere_virtual_machine.template.network_interface_types[0]}"
  }

  disk {
    name = "terraform-test.vmdk"
    size = "${data.vsphere_virtual_machine.template.disk_sizes[0]}"
  }

  disk {
    name = "terraform-test_1.vmdk"
	unit_number = 1
    size = "${data.vsphere_virtual_machine.template.disk_sizes[1]}"
  }

  clone {
    template_uuid = "${data.vsphere_virtual_machine.template.id}"
  }
}
```

The other issue we noticed comes from cloning a template with vApp settings. Before attempting to deploy the vm, our test fails at [checking for perpetual diffs](https://github.com/hashicorp/terraform/blob/master/helper/resource/testing_config.go#L94-L101) with the following error:
```
After applying this step, the plan was not empty:

    		DIFF:

    		UPDATE: vsphere_virtual_machine.vm
    		  network_interface.0.bandwidth_limit:       "0" => "-1"
    		  network_interface.0.bandwidth_share_level: "" => "normal"
```
If we explicitly set a `bandwidth_limit` value in the network interface, we get an API error from vSphere:
```
error reconfiguring virtual machine: A specified parameter was not correct.
    		spec.deviceChange.device.resourceAllocation.limit
```
Some research indicated that bandwidth limit and bandwidth share level for network interfaces are new features in vSphere 6.0.

cc @rowanjacobs

vancluever:
These are not default settings so this only comes up if you try to set
them, but they should be restricted anyway. Oddly they are settable on
vCenter 5.5 update 3f, and the settings take, but they do not seem to be
visible in the MOB. The docs say that the settings are not available
until vSphere 6 anyway.

Fixes #286.

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_network_interface_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_network_interface_subresource.go
@@ -15,6 +15,7 @@ import (
 	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/nsx"
 	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/provider"
 	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/structure"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/viapi"
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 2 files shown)


vancluever:
Smoke tested on our secondary vSphere 5.5 lab. We also have a test that does set resource allocation settings which still passes on 6.5:

```
=== RUN   TestAccResourceVSphereVirtualMachine/multi-device
    --- PASS: TestAccResourceVSphereVirtualMachine (174.59s)
    --- PASS: TestAccResourceVSphereVirtualMachine/multi-device (174.59s)
```

PR #332, opened 2017-12-19
docs: Clarify the use of cluster/host root resource pool

### LINKED ISSUE #326: resource_pool_id without DRS
When DRS is not enabled (i.e. due to licensing: Essentials, Essentials Plus, Standard) no resource pools can be created. But resource_pool_id is a mandatory parameter.

I've got mine by importing a VM and looking at the state file, but I think resource_pool_id should not be mandatory when there are cases where there is none (visible). Even with govc ls I could not find a pool in my installation.

vancluever:
There has been some confusion surrounding the fact that the
`vsphere_virtual_machine` resource now only takes a resource pool ID with
users that do not make use of resource pools, opting to just put virtual
machines in the root-level cluster or standalone host. This actually
places those virtual machines in a root resource pool that represents
all of the compute power of a particular host or set of hosts.  This
pool is normally invisible to the user and can only be seen generally by
using a tool that can browse the inventory off the CLI, like `govc`.

There are no plans to complicate the VM resource's implementation by
adding parameters like `compute_resource_id` - there is no point as they
would ultimately be looking up a resource pool anyway. Future plans may
see the development of data sources like `vsphere_cluster` and
`vsphere_standalone_host`, which could possibly have `resource_pool_id`
exported which would reference the root resource pool for that specific
resource. These may be developed along with their resource counterparts,
which are planned for a future release.

To back up the above, this commit adds a bit of clarification in the VM
resource on why only a resource pool is accepted - this can be expanded
on after we add the cluster and standalone host data sources. I've also
rephrased the explanation in the `vsphere_resource_pool` documentation so
that it's bit more clear what the root resource pool is and that this
logic also applies to standalone hosts. Also, the language has been
changed from using "default" to using "root", so that it's in line with
the vSphere official docs.

Fixes #326.

### DIFF
--- a/website/docs/d/resource_pool.html.markdown
+++ b/website/docs/d/resource_pool.html.markdown
@@ -23,16 +23,17 @@ data "vsphere_datacenter" "datacenter" {
 }
 
 data "vsphere_resource_pool" "pool" {
-  name          = "cluster1/Resources/resource-pool-1"
+  name          = "resource-pool-1"
   datacenter_id = "${data.vsphere_datacenter.dc.id}"
 }
 ```
 
-### Specifying the default resource pool for a cluster
+### Specifying the root resource pool for a cluster or standalone host
 
-If you don't have any resource pools in your cluster, or you want to use the
-parent resource pool for a cluster, you can just specify the `Resources` named
-default in your path for the resource pool:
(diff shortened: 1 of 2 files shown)


mbfrahry:
LGTM!

Issue #274, opened 2017-12-05
Provide a way to set DRS affinity rules


mmb:
I would like to create virtual machines with Terraform and tell vSphere to keep them on different hosts.

govc has recently added support for this:

https://github.com/vmware/govmomi/pull/928

vancluever:
Thanks for the info @mmb - super useful to know that govmomi has an example.

Cluster affinity is something we can probably touch when start the effort to add better cluster support ourselves - which is something we will be looking at soon, now that 1.0 is out!

Matty9191:
+1 for this.

d10sbt:
+2 for this

seanmalloy:
+1 for this feature

mjrider:
+4

virtrixke:
Yes, definitly a useful feature...

witchbutter:
+5 really need this.

movergan:
+1 please!

arri-cc:
Very nice

vancluever:
Everyone - 1.5.0 is out now with all of the new override and rule resources. Check them out and if you have any problems please submit any issues you need to!

Thanks!

PR #339, opened 2018-01-06
[WIP] r/cluster_rule resource

jnahelou:
Following #274, I started to add ```affinity```, ```antiaffinity```, ```vmhost``` (affine and antiaffine) rules as a generic resource called ```cluster_rule```.
To use it, I had to fetch vendor ```vmware/govmomi/object```
Waiting for feedback to continue.


pryorda:
:+1:

arri-cc:
Do you need someone to test this? @jnahelou

vancluever:
Hey all, apologies for the radio silence on this one.

I just wanted to point out that with the release of 1.4.0, our focus is now shifting towards resources of this kind and others in the clustering space (namely resource pools and compute clusters). We will probably be building on this PR but there may be changes that we may be making to the way these resources are structured - namely, more than likely these resources will be broken up into their individual rule types versus putting them all into one type of rule, as there are cluster rules that exist for both standard and Storage DRS, and we need to distinguish between the two (although it does look like these resources apply to just standard DRS).

As this is important for HashiCorp's longer-term strategy surrounding vSphere, we will probably be making changes to this PR to ensure that is at a standard that we are comfortable with. With that said, we appreciate the contributions, and the work here will be preserved and used as a base (possibly in a different branch though just in case anyone is depending on this work currently in a local copy).

Expect more details in the next couple of weeks.

Thanks!

vancluever:
Hey @jnahelou, thanks for this PR.

As I get further into working on all of the cluster resources it's becoming evident that we will be working on something that looks quite different enough from this implementation - namely, the fact that we want to have one resource per rule type is the particular part that will make the finished product different enough from this PR. Hence I'm just going to close this now in lieu of the `vsphere_compute_cluster_*_rule` and `vsphere_datastore_cluster_vm_anti_affinity_rule` resources, which will be following in the next couple of days.

Sorry for the delay in replying to this and thanks again!

PR #359, opened 2018-01-17
Integration with nsxt provider

annakhm:
Opaque network created by nsxt provider requires a specific backing.
This commit suggests providing nsxt uuid as network_id directly in
network interface subresource. To do so, new is_opaque argument needs
to be set.

### DIFF
--- a/vsphere/internal/helper/nsx/nsx_helper.go
+++ b/vsphere/internal/helper/nsx/nsx_helper.go
@@ -21,7 +21,8 @@ import (
 // vsphere_virtual_machine resource, as there is no direct path from an opaque
 // network backing to the managed object reference that represents the opaque
 // network in vCenter.
-func OpaqueNetworkFromNetworkID(client *govmomi.Client, id string) (*object.OpaqueNetwork, error) {
+// govmomi's Reference for OpaqueNetwork is Network, hence the return type
+func OpaqueNetworkFromNetworkID(client *govmomi.Client, id string) (*object.Network, error) {
 	// We use the same ContainerView logic that we use with networkFromID, but we
 	// go a step further and limit it to opaque networks only.
 	m := view.NewManager(client.Client)
@@ -58,7 +59,7 @@ func OpaqueNetworkFromNetworkID(client *govmomi.Client, id string) (*object.Opaq
 			// Should be safe to return here, as we have already asserted that this type
 			// should be a OpaqueNetwork by using ContainerView, along with relying
 			// on several fields that only an opaque network would have.
-			return nref.(*object.OpaqueNetwork), nil
+			return nref.(*object.Network), nil
 		}
 	}
 	return nil, fmt.Errorf("could not find opaque network with ID %q", id)
--- a/vsphere/internal/virtualdevice/virtual_machine_network_interface_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_network_interface_subresource.go
@@ -18,6 +18,7 @@ import (
 	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/viapi"
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
@@ -44,7 +45,7 @@ var networkInterfaceSubresourceMACAddressTypeAllowedValues = []string{
 	string(types.VirtualEthernetCardMacTypeManual),
 }
 
-// NetworkInterfaceSubresourceSchema returns the schema for the disk
+// NetworkInterfaceSubresourceSchema returns the schema for the network
 // sub-resource.
 func NetworkInterfaceSubresourceSchema() map[string]*schema.Schema {
 	s := map[string]*schema.Schema{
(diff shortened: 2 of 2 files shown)


vancluever:
Hey @annakhm, I tried contacting you you over the VMware {code} Slack about this but I haven't heard anything back, so I'm just putting it here instead:

My question is there still a way we can accomplish this while still inferring the opaque network from a MOID? The main reason - I want to avoid the `is_opaque` flag.  I've been trying to keep those kinds of flags out of the provider and for the most part have been successful. 

What I'm thinking is if govmomi's helpers aren't doing what we need to, then maybe we wrap it in another helper? From what I see here, it sounds like vSphere needs the extra NSX info that it's not getting in the base MO, which is fine, we just need to do a bit of extra work to get it.

I guess this is all contingent of if the NSX provider can export the ID for any particular opaque network.

Let me know and sorry for the delays on this!

annakhm:
Hi @vancluever, sorry about not responding on slack - I don't get notifications from there apparently..
Unfortunately NSX has no way to retrieve vsphere moid for the realized network.  
My main motivation was to avoid forcing the user to define a separate data source for each network, which would be tiresome for the user at scale - with nsx the network is a dynamic resource.
I guess we can avoid the is_opaque flag by looking at network_id format. Moid and uuid have different formats so these are not expected to collide. What do you think?

vancluever:
Hey @annakhm - I'm really sorry for not replying to this - I thought I did!

I think inferring network_id is a good compromise. We try to avoid it but I think it's acceptable in the face of ensuring we don't have to do this in a roundabout way with the vsphere_network data source

I'm actually just investigating this right now and will work off or your PR.

Sorry again for the radio silence!

vancluever:
@annakhm I'm also kind of wondering if we should offload a lot of this to govmomi (namely, fix EthernetCardBackingInfo so that it returns the correct backing for opaque networks). I'm actually planning on going that route and then will put in a PR against govmomi, in addition to allowing the inference - we can then add a helper to our network helpers that selects an opaque network based on the logical switch UUID.

vancluever:
Hey @annakhm, I actually want to revisit this approach as I just discovered that the root cause of the issues was a bug that I've just put in a PR to fix over at vmware/govmomi#1034. This actually corrects things so that the current code works without any issues.

I think this ultimately brings us back to the issue of that the MOID cannot be extracted from a logical switch directly in NSX-T. I'm on the fence of what is the lesser of the two evils - adding in inference that a UUID automatically and indefinitely means that you will be using an Opaque network, or just using the 
`vsphere_network` data source to pull down the opaque network from vSphere.

From an approach of which one is less invasive, the latter seems like the correct option, but from a UX perspective the former slightly wins. This, of course, is all dependent on the assumption that we are never ever going to use a UUID in the `network_id` field for anything else other than NSX logical switches.

I kind of want to let this one sit for a few more days to see what I think, of course let me know your thoughts as well!

If you want to test with the update, you should be able to vendor vmware/govmomi#1034 into your copy of the vSphere provider, or just make the code change manually - it's a very minor change.

More info in the next couple of days!

vancluever:
Hey @annakhm, after giving it some thought and doing some more testing with the NSX-T provider, my thoughts are that while longer term I think it might be a good idea to allow inference of opaque networks through supplying a UUID versus MOID, there are a couple of things that are getting in our way right now and I think are possibly things to be resolved at a later date:

## Logical Switch Realization/Opaque Network Delays

First off, there seems to be a bit of a delay in opaque networks showing up in vSphere that even waiting for full realization does not seem to correct. This manifests itself in errors like:

```
* data.vsphere_network.network: data.vsphere_network.network: error fetching network: network 'ffb3980e-b212-4d01-ad28-ebc5d03d8391' not found
```

(I left the LS name as the default UUID here)

If this error is not fixed, and we end up offloading the network lookup by ID to the VM, there is the possibility that a virtual machine could end up in a misconfigured state upon cloning, and is one of the reasons why we work with MOIDs almost strictly in the vSphere provider, as knowing we have an MOID guarantees as much as possible that we won't run into any errors during major operations that could be risky if incorrect data is supplied.

## Testing

While something that we could probably ultimately get around, right now we have a single path for network lookup and if we were to change that by adding another, we would need to more than likely test it. The real infrastructure required for testing right now is already pretty complicated, and adding a dependency to NSX-T to it is a concern I have.

That said, if we can take care of the first issue, I don't see why we can't work on getting this added, or at the very least, refactor things so that the appropriate code can be unit tested and not require NSX-T. One thing that I want to do is start testing more of the code using the govmomi simulator - under this framework, we could add opaque networks to the simulator inventory in the way that other types of network exist today. 

-----

Once we take care of this, I think we could possibly revisit this, in a different form of course as the issues with actually getting opaque networks has been resolved with vmware/govmomi#1034. For now though, I think the extra `vsphere_network` data source to bridge the translation of the LS to opaque network should be okay.

Thanks for submitting this - I'm excited to see how TF support for NSX-T will complement the vSphere provider!

annakhm:
Hi @vancluever,
Thanks for spending time on this! I'm glad we don't need to duplicate the govmomi code here. The realization issue is expected to be resolved in coming releases. I think this is an issue for the user regardless our approach in uuid, the workaround for now is to add sleep..
I will test nsxt provider with your fix. 
Do you think its a good idea to add nsx uuid to the network data source as a way to retrieve moid? I'm concerned about depending on network name which is not unique.
Thanks,
Anna

vancluever:
Hi @annakhm!

I think adding UUID to the vsphere_network data source would be a great compromise actually! This would bridge the gap between NSX and vSphere, which are in a way de-coupled now, and actually is a lot lower-risk longer-term than adding this to `vsphere_virtual_machine` as a data source is read-only and there is no state to manage.

The attribute could be something like `opaque_network_id` and conflict with both the current `name` and `datacenter_id` attributes as it would just search the entire vSphere inventory anyway.

Issue #381, opened 2018-02-02
New 'vapp' options do not work with Ubuntu 14.04/16.04 OVA


sdemura:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform
issue trackers reserved for bug reports and feature requests. For general usage
questions, please see: https://www.terraform.io/community.html.

### Terraform Version
```
terraform-vsphere$ terraform -v
Terraform v0.11.3
+ provider.vsphere v1.3.1
```
### vSphere Provider Version

```
terraform-vsphere$ terraform providers
.
└── provider.vsphere 1.3.1
```

### Affected Resource(s)

Please list the resources as a list, for example:
- `vsphere_virtual_machine`

### Terraform Configuration Files
```
resource "vsphere_virtual_machine" "vm" {
  name             = "terraform-xenial"
  resource_pool_id = "${data.vsphere_resource_pool.pool.id}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"

  num_cpus = 2
  memory   = 1024
  guest_id = "ubuntu64Guest"

  network_interface {
    network_id = "${data.vsphere_network.network.id}"
  }

  disk {
    name = "terraform-xenial.vmdk"
    size = 40
  }

  vapp {
    properties {
      "password" = "super-secure-password"
      "hostname" = "custom-hostname"
    }
  }

  clone {
    template_uuid = "${data.vsphere_virtual_machine.template.id}"
  }
}
```

### Expected Behavior
VM should have been created, and configured with the provided vApp properties.

### Actual Behavior
VM hostname and password are not changed.

### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
- vSphere 6.5
- Official Ubuntu cloud image: https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64.ova (Note, the Trusty equivalent does not work either)

Using `govc import.ova -options=xenial.json` works as expected, where xenial.json looks like:
```
{
    "DiskProvisioning": "thin",
    "IPAllocationPolicy": "dhcpPolicy",
    "IPProtocol": "IPv4",
    "InjectOvfEnv": false,
    "Name": null,
    "NetworkMapping": [
        {
            "Name": "VM Network",
            "Network": ""
        }
    ],
    "PowerOn": true,
    "PropertyMapping": [
        {
            "Key": "instance-id",
            "Value": "id-ovf"
        },
[...]

PR #383, opened 2018-02-02
Add OVF deployment example

vancluever:
This example demonstrates how one can use the vApp property management features of the `vsphere_virtual_machine` resource to deploy a 3-node webapp deployment using CoreOS, Ignition, and provisioners. A README is provided for documentation.

The web demonstration application used, [http-hello-go](https://github.com/vancluever/http-hello-go), is one I use to demonstrate similar patterns in other projects. The code does not self-contain the app to avoid possible licensing issues, but rather downloads and builds it as part of the orchestrated build process.


javefang:
Thanks for the example! I'm still puzzled by the use of vapp properties here.

What I'm trying to achieve is to build a vsphere template with Packer which will allow me to pass in cloudinit userdata like I could do with AWS EC2 userdata. Currently we are packaging userdata in an ISO file, upload as vsphere_file and mount to the machine, which is really hacky. But with the new vapp properties feature, I wonder if I could pass userdata via this? I already found a possible cloudinit data source implementation which could read vapp prorperties https://github.com/xing/cloudinit-vmware-guestinfo. But I struggle to find an end-to-end example from building vsphere template with Packer, to deploying with vapp properties in Terraform.

My question is: How can I build a template with Packer that can be used as a template stated in this use-case https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#cloning-from-an-ovf-ova-created-template-with-vapp-properties ?

vancluever:
@javefang to the best of my knowledge, OVF authoring (which entails the editing of vApp options to enough of a templated fashion before you export to OVF/OVA) is currently not a thing in Packer. Sorry!

In regards to the relation of vApp properties to where user data actually needs to be supplied, this varies from appliance to appliance. It's usually best to check the OVF file for the image that you are deploying for the exact keys that you need to use. You can also dump the properties to JSON using `govc import.spec`, which is documented [here](https://www.virtuallyghetto.com/2016/04/slick-way-of-deploying-ovfova-directly-to-esxi-vcenter-server-using-govc-cli.html). Most OVF files for Linux distros give you the option to supply base-level network information directly, or passing along user data where you can supply a complete cloud-config.

vancluever:
PS: For some more advanced deployment options, you could always look into how Ubuntu sets up `cloud-init` with the VMware plugin and work off of that, and supply the vApp options as extra VMX config values. You could also try to supply these options using the `vsphere` post-processor, or with `ovftool` directly after the fact. All of these are pretty advanced topics though so it's hard to explain in the scope of this PR.

If you do end up getting somewhere with this though, keep in mind that you will want to use the guestinfo transport right now to deliver the OVF environment, as we are still currently on trying to make ISO transports work correctly. See #381 for more details.

Issue #348, opened 2018-01-12
vsphere_virtual_machine moid attribute desirable


charandas:
Hi there,

This is a feature request for re-exposing the `moid` for the resource `vsphere_virtual_machine`. A big refactor must have happened in separating this out of the main terraform repo sometime back, which has effectively removed the attribute.

### References

As this [PR](https://github.com/hashicorp/terraform/pull/14793) shows, it was once added. But the current version v1.2.0 does not expose it, only UUIDs of the vm are exposed as far as guids.

This nsx provider(https://github.com/sky-uk/terraform-provider-nsx/blob/master/resource_security_tag_attachment.go#L54) depends on that, and I very much would like to not use old terraform version.

### What I tried as a workaround (and failed)

I tried using a `data.external` with some script like:
```
# ./helpers/get-vsphere-vm-moid/get-vm-moid.sh"
#!/bin/bash

set -e

vsphere_dc_name=$1
vm_guest_name=`jq -r '.vm_guest_name'`
vm_moid=$(/usr/local/bin/govc ls -i "/${vsphere_dc_name}/vm/${vm_guest_name}" | sed -e 's/^VirtualMachine://')

jq -n --arg vm_moid $vm_moid '{"value": $vm_moid}'
```

But I cannot splat on the result object due to a known terraform [bug](https://github.com/hashicorp/terraform/issues/17048#issuecomment-357038473).

charandas:
@vancluever what worked for me after some more tinkering is this:

```
"nsx_security_tag_attachment": {
         "client_vm": {
            "count": "${local.clientCount}",
            "moid": "${lookup(data.external.client_vm_moid.*.result[count.index], \"value\")}",
```

Coupled with something like this:
```
"client_vm_moid": {
    "count": "${local.clientCount}",
    "depends_on": [
        "vsphere_virtual_machine.client_vm"
    ],
    "program": [
        "bash",
        "./helpers/get-vsphere-vm-moid/get-vm-moid.sh",
        "name-of-the-dc"
    ],
    "query": {
        "vm_guest_name": "${element(local.clientHosts, count.index)}"
    }
  },
```

The combination of `lookup` with `[]` (another sugar over `element`) seems to get me the nested value in the map. Would be nice to just be able to get it from the resource `vsphere_virtual_machine` though.

vancluever:
[...]

PR #390, opened 2018-02-06
Issue 348

andreaspeters:
I feel this issue https://github.com/terraform-providers/terraform-provider-vsphere/issues/348 is a very good idea. So I add it and hope u feel ok with that. :-) Actually it was already inside. Only in the older version of this provider. https://github.com/hashicorp/terraform/pull/14793/files#diff-9de67d85a10a7cfacde989a1e65c4089

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -271,7 +271,7 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("error fetching VM properties: %s", err)
 	}
 
-	// Get out the machine object id
+	// Set the managed object id.
 	moid := vm.Reference().Value
 	d.Set("moid", moid)
 	log.Printf("[DEBUG] MOID for VM %q is %q", vm.InventoryPath, moid)
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -176,7 +176,7 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 		"moid": &schema.Schema{
 			Type:        schema.TypeString,
 			Computed:    true,
-			Description: "",
+			Description: "The machine object ID from VMWare",
 		},
 		vSphereTagAttributeKey:    tagsSchema(),
 		customattribute.ConfigKey: customattribute.ConfigSchema(),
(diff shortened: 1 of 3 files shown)


vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -173,6 +173,11 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 			Computed:    true,
 			Description: "A flag internal to Terraform that indicates that this resource was either imported or came from a earlier major version of this resource.",
 		},
+		"moid": &schema.Schema{
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "",
```
vancluever: Can you add a description to this?

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -260,10 +265,19 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 		}
 		return fmt.Errorf("error searching for with UUID %q: %s", id, err)
 	}
+
 	vprops, err := virtualmachine.Properties(vm)
 	if err != nil {
 		return fmt.Errorf("error fetching VM properties: %s", err)
 	}
+
+	err = d.Set("moid", vm.Reference().Value)
```
vancluever: This section looks like it could be optimized a bit, addressing a few things:

* The error you are getting back from `d.Set()` is actually for the state write, not for fetching the MOID, whose path does not yield an error.
* Since you are setting primitives, errors are not necessarily useful as the risk of there being a state write error on such an attribute is negligible.
* The log message that tells you what the MOID is could possibly be a little more informative.

If you can update it so that it resembles what's below that'd be great:

```go
moid := vm.Reference().Value
log.Printf("[DEBUG] MOID for VM %q is %q", vm.InventoryPath, moid)
d.Set("moid", moid)
```

vsphere/resource_vsphere_virtual_machine.go
andreaspeters: Yeah sorry! My mistake! Already update it some minutes ago. :-)

vsphere/resource_vsphere_virtual_machine.go
andreaspeters: thanks @vancluever if its like u are telling me, u are right and it makes no sense. :-) So I update it! I will test it in the next half hour.

andreaspeters:
Test done, its working with the changes! :-)

benfab:
@vancluever Would be great to have this issue merged, let us know is there is anything we could do to help.

I am very much looking forward to using the terraform-provider-vsphere and the nsx terraform provider in my IaC pipelines ;-)

Issue #382, opened 2018-02-02
multiple virtual machines withUUID found


johnjelinek:
### Terraform Version & vSphere Provider Version

```
Terraform v0.11.3
+ provider.vsphere (1.3.1)
```

### Debug Output

```
2018-02-02T14:23:41.957-0600 [DEBUG] plugin.terraform-provider-vsphere: 2018/02/02 14:23:41 [DEBUG] Locating virtual machine with UUID "420ae87a-bce6-2e1b-f6f6-fc813272ba24"
2018-02-02T14:23:41.957-0600 [DEBUG] plugin.terraform-provider-vsphere: 2018/02/02 14:23:41 [DEBUG] Using ContainerView to look up UUID "420ae87a-bce6-2e1b-f6f6-fc813272ba24"
2018/02/02 14:23:42 [ERROR] root.app-server: eval: *terraform.EvalDiff, err: cannot locate virtual machine or template with UUID "420ae87a-bce6-2e1b-f6f6-fc813272ba24": multiple virtual machines with UUID"420ae87a-bce6-2e1b-f6f6-fc813272ba24" found
2018/02/02 14:23:42 [ERROR] root.app-server: eval: *terraform.EvalSequence, err: cannot locate virtual machine or template with UUID "420ae87a-bce6-2e1b-f6f6-fc813272ba24": multiple virtual machines with UUID "420ae87a-bce6-2e1b-f6f6-fc813272ba24" found
2018/02/02 14:23:42 [ERROR] root.postgres-server: eval: *terraform.EvalDiff, err: cannot locate virtual machine or template with UUID "420a8fb9-1dcb-7849-514b-e1895928ae89": multiple virtual machines withUUID "420a8fb9-1dcb-7849-514b-e1895928ae89" found
```

### Expected Behavior

It should've found the template by UUID

### Actual Behavior

It found multiple templates (and the other ones didn't match the UUID)

### Steps to Reproduce

1. `terraform plan`

### Important Factoids

This filter is "finding" more than one instance:
https://github.com/terraform-providers/terraform-provider-vsphere/blob/21985b8fd66ebdb1b2927acd7ff4e7f5cd3b9a67/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go#L131

Here's a conversation about this [upstream](https://github.com/vmware/govmomi/issues/1020#issuecomment-362693459). Essentially, this `govc` command finds multiples (I think erroneously):
```
[...]

PR #391, opened 2018-02-06
r/virtual_machine: Switch ContainerView search to use manual filter

vancluever:
There seems like there is possibly a bug in `RetrieveWithFilter` when it
comes to MO's with missing keys and the function returning a false
positive for the object in question. This has caused issues with the VM
search functionality where duplicate UUID errors will be returned for
VMs with UUIDs that are otherwise unique within vSphere inventory if
there are orphaned VMs in inventory that lack a configuration (as in the
whole config data object is missing).

This bug applies to the provider when being used with vSphere 6.0 and
earlier.

### DIFF
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -128,7 +128,7 @@ func virtualMachineFromContainerView(ctx context.Context, client *govmomi.Client
 	}()
 
 	var vms, results []mo.VirtualMachine
-	err = v.Retrieve(ctx, []string{"VirtualMachine"}, nil, &results)
+	err = v.Retrieve(ctx, []string{"VirtualMachine"}, []string{"config.uuid"}, &results)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


vancluever:
Ref: #382 

@johnjelinek and anyone else having this issue can you give this fix a shot?

Thanks!

den-is:
just tell how to test it? I don't know

vancluever:
Hey @den-is, you will want to follow the instructions [here](https://github.com/terraform-providers/terraform-provider-vsphere#building-the-provider). Make sure you switch to the `b-vm-containerview-search` branch first before you build.

johnjelinek:
@vancluever: I get the same error from this version

johnjelinek:
nvm, I was forgetting a step to push in the new binary. Works on my machine 😄 ! It's slow though:
```
time terraform plan   
17.59s user 1.60s system 75% cpu 25.555 total
```

vancluever:
Great! Glad that it works!

Just making one more change - would you be able to test it? It's a speed improvement. :)

johnjelinek:
much better, thanks!
```
time terraform plan
1.57s user 0.68s system 24% cpu 9.320 total
```

vancluever:
Awesome! Merging now :+1:

PR #398, opened 2018-02-13
vendor: Update govmomi to bb8ff8d

vancluever:
This includes vmware/govmomi#1034, which fixes opaque network support,
ultimately allowing managed VMs to correctly work with NSX-T logical
switches.

A couple of minor code updates have been included to reflect updates to
some types and signatures.


vendor/vendor.json
```diff
@@ -814,134 +814,134 @@
...
+			"revisionTime": "2018-02-13T04:10:52Z"
 		},
 		{
 			"checksumSHA1": "6XLRzLEvncx5ORtvBjp41KfRTjc=",
 			"comment": "v0.15.0",
 			"path": "github.com/vmware/govmomi/vim25/xml",
-			"revision": "d68562200280ff2ad2110719b06496c9e9526118",
-			"revisionTime": "2017-10-13T00:55:03Z"
+			"revision": "bb8ff8d53171157123ae812088fd06223d8d90a3",
+			"revisionTime": "2018-02-13T04:10:52Z"
```
tombuildsstuff: should we not be pinning to the tag of ` 0.16.0`?

vancluever: Versions in govmomi generally are only useful for govc. The SDK is unversioned. As an example, `0.16.0` was actually released back in November.

Issue #402, opened 2018-02-16
vsphere_virtual_machine.vm does not have attribute 'default_ip_address'


den-is:
After upgrading vCenter from 5.5 to 6.5 some optional steps stopped working.
In this case `terraform plan` is trying to get VM IP before VM is actually created.

### Versions
```
❯ terraform -v
Terraform v0.11.3
+ provider.vsphere v1.3.2
```

### Affected Resource(s)
```
❯ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.vsphere_datacenter.dc: Refreshing state...
data.vsphere_resource_pool.rpool: Refreshing state...
data.vsphere_datastore.ds: Refreshing state...
data.vsphere_virtual_machine.template: Refreshing state...
data.vsphere_network.net: Refreshing state...
------------------------------------------------------------------------
Error: Error running plan: 1 error(s) occurred:
* module.tftest.output.ip_addresses: Resource 'vsphere_virtual_machine.vm' does not have attribute 'default_ip_address' for variable 'vsphere_virtual_machine.vm.*.default_ip_address'
```

outputs.tf
```hcl
output "ip_addresses" {
 value = "${vsphere_virtual_machine.vm.*.default_ip_address}"
}
```

vancluever:
@den-is this is usually due to an error higher up the dependency chain. Can you try removing the output and planning again and see if you can plan successfully?

Thanks!

den-is:
@vancluever yes of course when I'm removing it everything works fine.
I've created TF module which contains some default outputs declarations for future use.
with 5.5 it was working

vancluever:
@den-is - I think I see the error in your output actually:

```
module.tftest.output.ip_addresses: Resource 'vsphere_virtual_machine.vm' does not have attribute 'default_ip_address' for variable 'vsphere_virtual_machine.vm.*.default_ip_address'
```

The resource address implies that there is only one VM (ie: no `count` or a `count` of 1), so globbing with `*` won't work here. Try `vsphere_virtual_machine.vm.default_ip_address`.

den-is:
hmmmm can't get it...
if what am I doing wrong
this is my module initialization:
```hcl
module "tftest" {
  source           = "./vmwareprod"
  servers          = 2
  hostname_prefix  = "tftest"
  domain           = "int.example.com"
  template         = "template-centos-7.4"
[...]

PR #405, opened 2018-02-17
r/virtual_machine: Address vCenter upgrade issues

### LINKED ISSUE #403: vsphere_virtual_machine.vm disk.size got 0
After upgrading vCenter from 5.5 to 6.5 some optional steps stopped working.
In this case terraform plan is getting template disk size 0 while it is 250G.
So as workaround I've it "harcoded" in variables.

### Terraform Version
```
❯ terraform -v
Terraform v0.11.3
+ provider.vsphere v1.3.2
MacOS High Siera
```

### Affected Resource(s)
```
❯ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.vsphere_datacenter.dc: Refreshing state...
data.vsphere_datastore.ds: Refreshing state...
data.vsphere_resource_pool.rpool: Refreshing state...
data.vsphere_network.net: Refreshing state...
data.vsphere_virtual_machine.template: Refreshing state...
------------------------------------------------------------------------
Error: Error running plan: 1 error(s) occurred:
* module.tftest.vsphere_virtual_machine.vm: expected disk.0.size to be at least (1), got 0
```

disk declation:
```hcl
  disk {
    label = "${var.hostname_prefix}-${format("%02d", count.index + 1)}.${var.domain}_0"
    # size  = "${var.disk_size}" 
    size  = "${data.vsphere_virtual_machine.template.disks.0.size}"  # FIXME: Disabled after 6.5
  }
```

Adding debug output of how list of disks looks like shows this, actual zero:
```hcl
template_disks = [
    {
        eagerly_scrub = 0,
        size = 0,
        thin_provisioned = 1
    }
]
```

vancluever:
A couple of issues come up when vCenter is upgraded:

* First off, disk `capacityInBytes` values for virtual machine disk devices are cleared. This is a value we rely on to set the capacity of a disk, so a fallback to capacityInKB is necessary. This is specifically important for templates, as while VMs will be powered on and this value will be populated again, more than likely templates will remain off after the upgrade.
* Also post-upgrade, the VM hardware version of an existing virtual machine will not have been updated (this would need to be changed manually), hence `disk_sharing` will be unset and unusable. However, this value will now be read in by Terraform due to the upgraded vCenter, albeit with a blank value, and will trigger a diff that will ultimately result in an error. Hence, we skip setting the sharing value if it is blank.

Note that the second issue presents another problem that is not fixed by this PR, namely that if only vCenter is upgraded, and the hypervisors are not, more than likely deploying VMs using Terraform will fail until VMs can be deployed using a hardware version that supports disk sharing. This won't be fixed until we have some awareness around VM hardware versions, which is currently not the case.

Fixes #403. Possibly helps with #402 as well but would remain to be seen.

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
@@ -1094,7 +1094,7 @@ func ReadDiskAttrsForDataSource(l object.VirtualDeviceList, count int) ([]map[st
 		if backing.ThinProvisioned != nil {
 			thin = *backing.ThinProvisioned
 		}
-		m["size"] = int(structure.ByteToGiB(disk.CapacityInBytes).(int64))
+		m["size"] = diskCapacityInGiB(disk)
 		m["eagerly_scrub"] = eager
 		m["thin_provisioned"] = thin
 		out = append(out, m)
@@ -1194,7 +1194,7 @@ func (r *DiskSubresource) Read(l object.VirtualDeviceList) error {
 			return fmt.Errorf("could not parse path from filename: %s", b.FileName)
 		}
 		r.Set("path", dp.Path)
-		r.Set("size", structure.ByteToGiB(disk.CapacityInBytes))
+		r.Set("size", diskCapacityInGiB(disk))
 	}
 
 	if allocation := disk.StorageIOAllocation; allocation != nil {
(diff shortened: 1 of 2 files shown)


vancluever:
Added tests for `diskCapacityInGiB`:

```
=== RUN   TestDiskCapacityInGiB
=== RUN   TestDiskCapacityInGiB/capacityInBytes
=== RUN   TestDiskCapacityInGiB/capacityInKB
2018/02/26 08:12:18 [DEBUG] diskCapacityInGiB: capacityInBytes missing for for disk-0-0, falling back to capacityInKB
--- PASS: TestDiskCapacityInGiB (0.00s)
    --- PASS: TestDiskCapacityInGiB/capacityInBytes (0.00s)
    --- PASS: TestDiskCapacityInGiB/capacityInKB (0.00s)
```

vancluever:
Thanks @paultyng!

Issue #413, opened 2018-02-27
Deploy Cisco Firepower Management Center cause Terraform crash


Koleon:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform
issue trackers reserved for bug reports and feature requests. For general usage
questions, please see: https://www.terraform.io/community.html.

### Terraform Version
```
Terraform v0.11.3
+ provider.vsphere v1.3.2
```
### vSphere Provider Version
```
.
├── provider.terraform
└── provider.vsphere ~> 1.3.0
```

### Affected Resource(s)

```
resource "vsphere_virtual_machine" "FMCv" {
  count            = "${terraform.workspace == "default" ? 0 : 1}"
  name             = "${local.machine_prefix}-FMCv"
  resource_pool_id = "${local.pool_name}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"
  folder           = "${local.srv_folder}"
  guest_id         = "${data.vsphere_virtual_machine.FMCv.guest_id}"
  scsi_type        = "${data.vsphere_virtual_machine.FMCv.scsi_type}"
  network_interface {
    network_id = "${local.srv_interface}"
  }
  network_interface {
    network_id = "${data.vsphere_network.mgmt.id}"
  }
  disk {
    label            = "disk0"
    size             = "${data.vsphere_virtual_machine.FMCv.disks.0.size}"
    eagerly_scrub    = "${data.vsphere_virtual_machine.FMCv.disks.0.eagerly_scrub}"
    thin_provisioned = "${data.vsphere_virtual_machine.FMCv.disks.0.thin_provisioned}"
  }
  clone {
    template_uuid = "${data.vsphere_virtual_machine.FMCv.id}"
    customize {
      timeout = "${var.timeout}"
      linux_options {
        host_name = "${local.machine_prefix}-fmcv"
        domain    = "cycz.test"
      }
      network_interface {
        ipv4_address = "${local.srv_ip_prefix}.2"
        ipv4_netmask = 24
      }
      network_interface {
        ipv4_address = "${local.mgmt_ip_prefix}.22"
        ipv4_netmask = 18
      }
      ipv4_gateway   = "${local.srv_ip_prefix}.1"
    }
  }
}
```
```
data "vsphere_virtual_machine" "FMCv" {
  name          = "Cisco_Firepower_Management_Center"
  datacenter_id = "${data.vsphere_datacenter.dc.id}"
}
```
[...]

PR #416, opened 2018-02-28
Added a check for invalid ISO file on CD/DVD Drives

jodysimpson:
This is a fix for #413.  

When a VM is created from a template or VM that has an OVF environment,  vSphere will create an ISO and mount it as CD/DVD Drive.  If subsequent VM are created in Terraform after the VM had the ISO loaded,  vSphere will report to Terraform that the CD/Drive is loaded with an ISO of path "[]".  The vSphere provider then tries to access an invalid Datastore field and causes a crash.  

This fix checks to make sure the Datastore is not nil before accessing it, otherwise it skips over it.

### DIFF
--- a/vsphere/helper_test.go
+++ b/vsphere/helper_test.go
@@ -663,3 +663,25 @@ func testResourceHasCustomAttributeValues(s *terraform.State, resourceType strin
 	}
 	return nil
 }
+
+func testSetOvfEnvironmentTransportIso(s *terraform.State, resourceName string) error {
+	vm, err := testGetVirtualMachine(s, resourceName)
+	if err != nil {
+		return err
+	}
+	if err := testPowerOffVM(s, resourceName); err != nil {
+		return err
+	}
+
+	spec := types.VirtualMachineConfigSpec{
+		VAppConfig: &types.VmConfigSpec{
+			OvfEnvironmentTransport: []string{"iso"},
+		},
+	}
+	virtualmachine.Reconfigure(vm, spec)
+	virtualmachine.PowerOn(vm)
+	if err := testPowerOffVM(s, resourceName); err != nil {
+		return err
+	}
+	return nil
+}
--- a/vsphere/helper_test.go
+++ b/vsphere/helper_test.go
@@ -686,23 +686,23 @@ func testGetDatastoreClusterProperties(s *terraform.State, resourceName string) 
 }
 
 func testSetOvfEnvironmentTransportIso(s *terraform.State, resourceName string) error {
-    vm, err := testGetVirtualMachine(s, resourceName)
-    if err != nil {
-        return err
-    }
-    if err := testPowerOffVM(s, resourceName); err != nil {
-        return err
-    }
+	vm, err := testGetVirtualMachine(s, resourceName)
+	if err != nil {
+		return err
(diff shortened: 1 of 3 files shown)


vancluever:
@jodysimpson this seems like a very easy win - thank you for the contribution!

I want to give this a bit more of a run next week just to double check. In the meantime, it would be super helpful if you could get a test added in for this to `resource_vsphere_virtual_machine_test.go`. You can follow [this test](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/resource_vsphere_virtual_machine_test.go#L650) to see how we manipulate state to make OOB changes to virtual machines to create test cases similar to what you would need to do (eject the ISO from a datastore virtual CDROM device).

Regardless I will loop back on this next week.

Thanks!

jodysimpson:
I am working on the test for this situation,  it’s been difficult trying to recreate a VM that’s causes a crash from scratch.  I can create a problem VM inside vCenter but the vsphere provider either prevents me from creating the problematic situation or does not allow adjustment of the settings.  I am almost there,  it’s just been difficult.

Steps to reproduce:
1.  Create a VM/template with VApp or OVF settings and a CD-Rom that has no media.
2.  Set OVFEnvironmentTransport to ISO.
3. Power on machine.  This loads OVF settings into an ISO and mounts it to the CD-rom.
4. Power off the machine.  This sets the CD-rom media to “[]” with no datastore.
5. Clone this machine to a second VM and have Terraform try to read this new VM.  Can either be a Terraform state import of the already cloned VM or a vsphere_virtual_machine that clones the original VM.

In order to replicate this as a test case,  I am doing the following:
1. Start with the Clone VApp test which copies the CoreOS template.
2. Clone the CoreOS template with an added CD-rom
3.  Once the VM is loaded edit the properties to remove the media from the CD-rom and set the OVFEvironmentTransport to “iso”.  The vsphere provider does not directly allow either of these to be done in HCL.  I am able to create the problematic situation by editing the properties similar to the test case you pointed me to.
4. Power on the VM to have vsphere load the ISO.
5. Power off the VM,  this unloads the ISO and sets it to “[]”
6. Import this VM into a new resource and hopefully crash before my fix.  This is the step I’m stuck on but I should have it fixed soon.  I can’t get the test to crash importing in the test suite.  If I try run a command line state import of the exact same VM it crashes every time.

jodysimpson:
The test case is completed.  vSphere only sets the ISO to [] after a ~6 second timeout after powering down.  I added a 30 second sleep just to be safe.

vancluever:
Hey @jodysimpson, sorry about the delay in getting back to you on this.

Thank you for all your hard work on this, especially on the tests.

Coincidentally, as I think you might have seen in the changelog or other PRs, there has been some work on full support for vApp ISO transports. This work was completed a week or so ago and includes checking for ejected datastore ISO drives.

This work unfortunately supersedes the work here - the test scenario is not really valid anymore as we only allow machines that have vApp ISO transports to have a client CDROM device. A more valid test now would just be with a basic virtual machine with an ejected datastore ISO drive to ensure that the nil check was working properly. Hence I'm just going to close this PR.

Sorry for the inconvenience here and again, thanks for your hard work.

PR #418, opened 2018-03-02
Convert all acceptance tests to individual tests

vancluever:
This PR is large but hopefully straightforward, and is a refactor to convert acceptance tests from the provider-local subtest pattern to the more widely used single-function-per-test pattern (ie: `TestAccResourceVSphereSomeResource_basic` et al).

The reason this is needed is that our currently acceptance test runner in CI relies heavily on the common convention, for doing things like collating the output of individual tests into their own streams, formatted into a format that TeamCity is aware of. `go test -list` is also made use of, which currently does not support subtests, so this would be a full stop blocker to getting nightly acceptance tests working here.

A couple of minor fixes have been made to tests that were missing `ImportStateVerify` keys, and also a couple of linter fixes were made as well. These have been noted as individual commits in most cases, or as commit message details in others.

### DIFF
--- a/vsphere/data_source_vsphere_custom_attribute_test.go
+++ b/vsphere/data_source_vsphere_custom_attribute_test.go
@@ -7,50 +7,34 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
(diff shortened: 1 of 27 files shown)


vancluever:
Thanks @tombuildsstuff!

PR #421, opened 2018-03-05
r/virtual_machine: Add support for remote backed CDROMs

bill-rich:
Previously, only ISO backed CDROMs were supported. This adds the ability to use
a remote client device instead.

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
@@ -20,13 +20,19 @@ func CdromSubresourceSchema() map[string]*schema.Schema {
 		// VirtualDeviceFileBackingInfo
 		"datastore_id": {
 			Type:        schema.TypeString,
-			Required:    true,
+			Optional:    true,
 			Description: "The datastore ID the ISO is located on.",
 		},
 		"path": {
 			Type:        schema.TypeString,
-			Required:    true,
+			Optional:    true,
 			Description: "The path to the ISO file on the datastore.",
+		},
+		// VirtualCdromRemoteAtapiBackingInfo
+		"client_device": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Indicates whether the device should be backed by remote client device",
 		},
 	}
 	structure.MergeSchema(s, subresourceSchema())
@@ -378,6 +384,25 @@ func CdromPostCloneOperation(d *schema.ResourceData, c *govmomi.Client, l object
 	return l, spec, nil
 }
 
+// ValidateCdromConfig takes a CdromSubresource and ensures that the associated configuration is valid
+func (r *CdromSubresource) ValidateCdromConfig() error {
+	log.Printf("[DEBUG] ValidateCdromConfig: Beginning CDROM configuration validation")
+	dsID := r.Get("datastore_id").(string)
+	path := r.Get("path").(string)
+	clientDevice := r.Get("client_device").(bool)
+	if clientDevice {
+		if dsID != "" || path != "" {
+			return fmt.Errorf("ValidateCdromConfig: Cannot have both client_device parameter and ISO file parameters (datastore_id, path) set")
+		}
+	} else {
+		if dsID == "" || path == "" {
+			return fmt.Errorf("ValidateCdromConfig: Either client_device or datastore_id and path must be set")
+		}
+	}
+	log.Printf("[DEBUG] ValidateCdromConfig: Config validation complete")
+	return nil
+}
+
 // Create creates a vsphere_virtual_machine cdrom sub-resource.
 func (r *CdromSubresource) Create(l object.VirtualDeviceList) ([]types.BaseVirtualDeviceConfigSpec, error) {
 	log.Printf("[DEBUG] %s: Running create", r)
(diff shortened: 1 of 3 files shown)


vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -20,14 +20,20 @@ func CdromSubresourceSchema() map[string]*schema.Schema {
...
 			Type:        schema.TypeString,
-			Required:    true,
+			Optional:    true,
 			Description: "The path to the ISO file on the datastore.",
 		},
+		// VirtualCdromRemoteAtapiBackingInfo
+		"client_device": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Indicates whether the device should be backed by remote client device",
```
vancluever: _Minor_ can we edit this so it reads:

```
Indicates whether the device should be mapped to a remote client device.
```

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -378,6 +384,25 @@ func CdromPostCloneOperation(d *schema.ResourceData, c *govmomi.Client, l object
 	return l, spec, nil
 }
 
+// ValidateCdromConfig takes a CdromSubresource and ensures that the associated configuration is valid
+func (r *CdromSubresource) ValidateCdromConfig() error {
+	log.Printf("[DEBUG] ValidateCdromConfig: Beginning CDROM configuration validation")
```
vancluever: Subresource debug lines should include the specific resource address. For this log entry, this means it would look like:

```
log.Printf("[DEBUG] %s: Beginning CDROM configuration validation", r)
```

Function does not necessarily need to be referenced here.

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -378,6 +384,25 @@ func CdromPostCloneOperation(d *schema.ResourceData, c *govmomi.Client, l object
...
+	dsID := r.Get("datastore_id").(string)
+	path := r.Get("path").(string)
+	clientDevice := r.Get("client_device").(bool)
+	if clientDevice {
+		if dsID != "" || path != "" {
+			return fmt.Errorf("ValidateCdromConfig: Cannot have both client_device parameter and ISO file parameters (datastore_id, path) set")
```
vancluever: Can we remove the function name from these error messages? They will be exposed to the user during regular output.

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -378,6 +384,25 @@ func CdromPostCloneOperation(d *schema.ResourceData, c *govmomi.Client, l object
...
+		}
+	} else {
+		if dsID == "" || path == "" {
+			return fmt.Errorf("ValidateCdromConfig: Either client_device or datastore_id and path must be set")
```
vancluever: As per above

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -378,6 +384,25 @@ func CdromPostCloneOperation(d *schema.ResourceData, c *govmomi.Client, l object
...
+		}
+	}
+	log.Printf("[DEBUG] ValidateCdromConfig: Config validation complete")
```
vancluever: This line should follow the same kind of formatting as the starting line, example:

```
log.Printf("[DEBUG] %s: Config validation complete", r)
```

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -438,9 +473,10 @@ func (r *CdromSubresource) Read(l object.VirtualDeviceList) error {
 	if !ok {
 		return fmt.Errorf("device at %q is not a virtual CDROM device", l.Name(d))
 	}
-	if backing, ok := device.Backing.(*types.VirtualCdromIsoBackingInfo); ok {
-		// Only read backing info if it's available. If not, this is a host
-		// device/client mapping that is not managed by TF and needs to be deleted.
+	// Only read backing info if it's available.
+	if _, ok := device.Backing.(*types.VirtualCdromRemoteAtapiBackingInfo); ok {
+		r.Set("client_device", true)
+	} else if backing, ok := device.Backing.(*types.VirtualCdromIsoBackingInfo); ok {
```
vancluever: As above, let's try to avoid `else if` as it's not necessarily good Go.

Instead, write this as a type `switch`:

```go
switch backing := device.Backing.(type) {
case *types.VirtualCdromRemoteAtapiBackingInfo:
  r.Set("client_device", true)
case *types.VirtualCdromIsoBackingInfo:
  ...
default:
  return nil, fmt.Errorf("device at %q is not a virtual CDROM device", l.Name(d))
}
```

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -389,30 +414,40 @@ func (r *CdromSubresource) Create(l object.VirtualDeviceList) ([]types.BaseVirtu
...
 
 	// We now have the controller on which we can create our device on.
-	dsID := r.Get("datastore_id").(string)
-	path := r.Get("path").(string)
-	ds, err := datastore.FromID(r.client, dsID)
-	if err != nil {
-		return nil, fmt.Errorf("cannot find datastore: %s", err)
-	}
-	dsProps, err := datastore.Properties(ds)
+	err = r.ValidateCdromConfig()
```
vancluever: `ValidateCdromConfig` should not be called here - this should have been done as part of a diff operation - so a function like `CdromDiffOperation` should have been created similar to the ones seen in disk and network device sub-resources. Each resource should be iterated over and run through the diff validation. This ensures that the validation is done at plan time and does not create confusion when someone attempts to do a create or update and runs into a configuration error they were told was okay before.

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -472,27 +508,39 @@ func (r *CdromSubresource) Update(l object.VirtualDeviceList) ([]types.BaseVirtu
 		return nil, fmt.Errorf("device at %q is not a virtual CDROM device", l.Name(d))
 	}
 
-	// To update, we just re-insert the ISO as per create, and send it as an edit.
 	dsID := r.Get("datastore_id").(string)
 	path := r.Get("path").(string)
-	ds, err := datastore.FromID(r.client, dsID)
+	clientDevice := r.Get("client_device").(bool)
+	err = r.ValidateCdromConfig()
```
vancluever: As per the comment in validation being in `Create`, this needs to be removed in favour of one that happens during diff customization.

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -389,30 +414,40 @@ func (r *CdromSubresource) Create(l object.VirtualDeviceList) ([]types.BaseVirtu
...
+			return nil, fmt.Errorf("could not get properties for datastore: %s", err)
+		}
+		dsName := dsProps.Name
+		dsPath := &object.DatastorePath{
+			Datastore: dsName,
+			Path:      path,
+		}
+		device = l.InsertIso(device, dsPath.String())
+		l.Connect(device)
+	} else if clientDevice {
```
vancluever: Let's try to avoid `else if` as it's not necessarily good Go.

This actually can be written as a switch:

```go
switch {
case dsID != "" && path != "":
  ...
case clientDevice == true:
  // You might want to use the full-form boolean expression above to be specific about intent
  ...
default:
  // This would be where your "else" goes, if you had one
}
```

vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -472,27 +508,39 @@ func (r *CdromSubresource) Update(l object.VirtualDeviceList) ([]types.BaseVirtu
...
-
-	dsPath := &object.DatastorePath{
-		Datastore: dsName,
-		Path:      path,
+		return nil, err
 	}
 
-	device = l.InsertIso(device, dsPath.String())
-	l.Connect(device)
+	if dsID != "" && path != "" {
```
vancluever: This might be a good candidate for a unexported helper as this code looks pretty much the same as it does in `Create`.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -3170,6 +3279,229 @@ resource "vsphere_virtual_machine" "vm" {
...
+    network_id = "${data.vsphere_network.network.id}"
+  }
+
+  disk {
+    label = "disk0"
+    size  = 20
+  }
+
+  cdrom {
+		client_device = true
```
vancluever: Remember to filter the fixture code through `terraform fmt -` to fix these formatting issues that will normally be hidden from you in the editor.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -3170,6 +3279,229 @@ resource "vsphere_virtual_machine" "vm" {
...
+  network_interface {
+    network_id = "${data.vsphere_network.network.id}"
+  }
+
+  disk {
+    label = "disk0"
+    size  = 20
+  }
+
+  cdrom {
```
vancluever: This should be (and will be when it's run through `terraform fmt -`) `cdrom {}`.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -3170,6 +3279,229 @@ resource "vsphere_virtual_machine" "vm" {
...
```
vancluever: Formatting needs to be fixed here too.

vancluever:
LGTM!

PR #427, opened 2018-03-09
r/virtual_machine: Fix broken acceptance tests and CD drives

vancluever:
This does a few things:

* Looks like some acceptance test were broken - first one was due to
something changing in RancherOS that does not allow for the proper
hot-remove of virtual NICs anymore. I've adjusted this test so it
powers off the VM first, which is not so much an issue as we aren't
testing for a hot-remove anyway, but more to make sure that things are
consistent in state.
* The vApp property tests were missing settings for `thin_provisioned` and
`eagerly_scrub`, which should be coming from the template data sources.
These were failing as a result as the CoreOS OVA does not use thin
provisioned disks.
* Finally, #421 introduced a change where we were actually failing if we
didn't recognize the CD drive type. There are a few we still don't
support, namely passthrough CD drives. I've changed this to a debug log
message, and a unsetting of all parameters. This should ensure that
devices that aren't recognized but need to be changed to a recognized
device are properly accounted for in the diff, in addition to removing
unrecognized devices in the event no CDROM drive is necessary.

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
@@ -471,7 +471,16 @@ func (r *CdromSubresource) Read(l object.VirtualDeviceList) error {
 		r.Set("datastore_id", backing.Datastore.Value)
 		r.Set("path", dp.Path)
 	default:
-		return fmt.Errorf("%s: Unknown CDROM type %s", r, backing)
+		// This is an unsupported entry, so we clear all attributes in the
+		// subresource (except for the device address and key, of course).  In
+		// addition to making sure correct diffs get created for these kinds of
+		// devices, this ensures we don't fail on CDROM device types we don't
+		// support right now, such as passthrough devices. We might support these
+		// later.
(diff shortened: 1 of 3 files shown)


vsphere/internal/virtualdevice/virtual_machine_cdrom_subresource.go
```diff
@@ -471,7 +471,16 @@ func (r *CdromSubresource) Read(l object.VirtualDeviceList) error {
...
 		r.Set("path", dp.Path)
 	default:
-		return fmt.Errorf("%s: Unknown CDROM type %s", r, backing)
+		// This is an unsupported entry, so we clear all attributes in the
+		// subresource (except for the device address and key, of course).  In
+		// addition to making sure correct diffs get created for these kinds of
+		// devices, this ensures we don't fail on CDROM device types we don't
+		// support right now, such as passthrough devices. We might support these
+		// later.
+		log.Printf("%s: [DEBUG] Unknown CDROM type %T, clearing all attributes", r, backing)
```
tombuildsstuff: this behaviour should probably also be specified in the docs?

vancluever: Done! Added an extra blurb on this one in the docs.

PR #436, opened 2018-03-20
New resource: vsphere_datastore_cluster

vancluever:
This is the first part of a new push to add clustering and DRS support in the vSphere provider.

This adds `vsphere_datastore_cluster`, which can be used to create datastore clusters and configure Storage DRS for it respectively. Pretty much everything that's exposed in vSphere is exposed in the resource.

Note that the docs reference functionality that is not implemented yet - this is the `datastore_cluster_id` attribute, which will basically be a "pseudo-folder" attribute and conflict with `folder` in the `vsphere_nas_datastore` and `vsphere_vmfs_datastore` resources, allowing a datastore to be created in or moved to a datastore cluster.

After this support is in and functional we can use it to start writing datastore cluster support into the `vsphere_virtual_machine` resource to address #2.

### DIFF
--- a/vsphere/internal/helper/folder/folder_helper.go
+++ b/vsphere/internal/helper/folder/folder_helper.go
@@ -191,6 +191,8 @@ func folderFromObject(client *govmomi.Client, obj interface{}, folderType RootPa
 		p, err = RootPathParticleNetwork.PathFromNewRoot(o.InventoryPath, folderType, relative)
 	case *object.Datastore:
 		p, err = RootPathParticleDatastore.PathFromNewRoot(o.InventoryPath, folderType, relative)
+	case *object.StoragePod:
+		p, err = RootPathParticleDatastore.PathFromNewRoot(o.InventoryPath, folderType, relative)
 	case *object.HostSystem:
 		p, err = RootPathParticleHost.PathFromNewRoot(o.InventoryPath, folderType, relative)
 	case *object.ResourcePool:
--- a/vsphere/resource_vsphere_datastore_cluster.go
+++ b/vsphere/resource_vsphere_datastore_cluster.go
@@ -1,0 +1,526 @@
+package vsphere
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/customattribute"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/folder"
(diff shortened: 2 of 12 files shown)


vsphere/internal/helper/folder/folder_helper.go
```diff
@@ -191,6 +191,8 @@ func folderFromObject(client *govmomi.Client, obj interface{}, folderType RootPa
 		p, err = RootPathParticleNetwork.PathFromNewRoot(o.InventoryPath, folderType, relative)
 	case *object.Datastore:
 		p, err = RootPathParticleDatastore.PathFromNewRoot(o.InventoryPath, folderType, relative)
+	case *object.StoragePod:
+		p, err = RootPathParticleDatastore.PathFromNewRoot(o.InventoryPath, folderType, relative)
```
paultyng: It would be nice if they just used some interfaces occasionally :cry:

vsphere/resource_vsphere_datastore_cluster.go
```diff
@@ -0,0 +1,879 @@
...
+			},
+			"sdrs_io_reservable_threshold_mode": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      string(types.StorageDrsPodConfigInfoBehaviorAutomated),
+				Description:  "The reservable IOPS threshold to use, percent in the event of automatic, or manual threshold in the event of manual.",
+				ValidateFunc: validation.StringInSlice(storageDrsPodConfigInfoBehaviorAllowedValues, false),
+			},
+			"sdrs_load_balance_interval": {
+				Type:         schema.TypeInt,
```
paultyng: I wonder if it would be better UX to express this as a `string`/`time.Duration`

vsphere/internal/helper/folder/folder_helper.go
vancluever: Not 100% ruling out there could be some refactoring here, and `InventoryPath` is part of a `Common` object that everything embeds, but we still need to know what _kind_ of object is being passed in, and there is no common interface for say datastore clusters. In fact, this would not 100% work here even if there was, as for example, `StoragePod` actually would inherit from `Folder` versus `Datastore` - and indeed emebeds that object instead of `Datastore`. :stuck_out_tongue:

vsphere/resource_vsphere_datastore_cluster.go
vancluever: I think this is minor, especially in the face of if we changed it here to something like `60m`/`1h`/`2d`, we would probably need to change it _everywhere_, else we risk harming the UX more than enriching it.

vancluever:
Thanks @paultyng!

PR #447, opened 2018-03-31
r/virtual_machine: Datastore cluster support

vancluever:
Ref: #2 (not necessarily fixes just yet, want to hold off on that until we add the SDRS VM exception resources)

This adds datastore cluster/SDRS support to the `vsphere_virtual_machine` resource, via the new attribute `datastore_cluster_id`. When this option is in use, all changes to underlying datastores in the resource are ignored unless disks drift outside of the datastore cluster.

In addition to this, when the setting is in use, most VM creation, cloning, update, and vMotion workflows are sent to the Storage DRS API to ensure proper placement of disks within datastores. The main exception is reconfigurations to virtual machines when no disks are being created, which produces an error in the SDRS API, hence these are still just sent to the standard reconfiguration workflow.

When `datastore_cluster_id` is in use, datastore assignment to individual disks is not possible, nor is it possible to use this with a VM that is attaching an external disk. This decision was made to simplify the workflows that Terraform would need to handle - especially in the latter scenario, it could be all too easy to have an "attached" disk (which is already purely a Terraform construct) to drift out of sync with state if the appropriate SDRS exemptions were not created and managed correctly.


website/docs/r/virtual_machine.html.markdown
```diff
@@ -325,6 +325,73 @@ resource "vsphere_virtual_machine" "vm" {
...
 
+### Using Storage DRS
+
+The `vsphere_virtual_machine` resource also supports Storage DRS, allowing the
+assignment of virtual machines to datastore clusters. When assigned to a
+datastore cluster, changes to a virtual machine's underlying datastores are
+ignored unless disks drift outside of the datastore cluster. The example below
+makes use of the [`vsphere_datastore_cluster` data
+source][tf-vsphere-datastore-cluster-data-source], and the
+[`datastore_clsuter_id`](#datastore_cluster_id) configuration setting. Note
```
bill-rich: Should be `datastore_cluster_id`

vancluever: Thank you sir!

PR #459, opened 2018-04-07
datastore.html sample code fix

pemcconnell:
Fixing example code so that the datacenter_id in vsphere_datastore is retrieved from the previous vsphere_datacenter block


vancluever:
Nice catch @pemcconnell and thanks a ton!

Issue #457, opened 2018-04-06
Please advise on upgrade path


skywalk7:
So we have couple hundred VMs already deployed with provider version 0.4.3.
Now I'm trying to update code and modules to support 1.3 and I'm running into issues:
1. linux_options and windows_options are mutually exclusive in clone block and it breaks one to one match from old resources to new ones.
2. When I try to rewrite resource configuration it complains about clone block:
`*terraform.EvalDiff, err: this resource was imported or migrated from a previous version and does not support cloning. Please remove the "clone" block from its configuration`. Also it makes `import` support unusable.
Can you please help me to understand how I can upgrade to the newer version of provider without completely redeploying everything.

vancluever:
@skywalk7, the resource was changed a lot in 1.0 to try 1) ensure the configuration was a lot more explicit and matched the API of vSphere a lot closer, and 2) steps were taken to minimize the risk of data loss or accidental deletion when migrating from older versions, namely surrounding the use of cloned virtual machines.

This means a couple of things for your situation:

1) If you were deploying both Linux and Windows VMs with the same configuration settings pre-1.0, you will have to break these configurations up. As you mentioned these two settings are now mutually exclusive and you cannot use Linux settings with a Windows template, or vice versa, nor can you specify both at the same time.

2) Unfortunately, we do not support migrating a pre-1.0 state to a configuration that makes use of VM cloning. This is because the presence of the `clone` sub-resource block, which was introduced in 1.0, forces a new resource. This restriction may be lifted in the future, but for now, the best path for you is to import and upgrade VMs to a configuration that does not use cloning, using a new supported configuration going forward.

I apologize for any inconvenience that this may cause you, and again we have definitely been considering ways to address the second issue specifically, but as the resource required very drastic changes to get it to where it is today, some decisions had to be made about legacy support.

I'm going to close this now, but I would definitely follow the repo for updates regarding legacy migrations and imports.

Thanks!

vancluever:
Hey @skywalk7 I want to take a stab at this (as your question got me thinking)!

If you have not started on any refactoring based on the above info (other than OS configuration breakup), maybe hold off a few days. I'll have an answer on whether or not it's going to be possible for us to change answer 2 anytime soon (including a bit more detail as to why it would not be if it is not easily technically possible).

Thanks!

vancluever:
@skywalk7 1.4.0 is now out with the new updates to import/upgrade capabilities. Give it a go and let us know if you have any issues!

PR #460, opened 2018-04-08
r/virtual_machine: Remove clone config restriction for import/legacy

vancluever:
This removes the restriction on the presence of the `clone` sub-resource
block for imported and legacy resources.

Previously, this was blocked as it forced a new resource, which is
obviously not what people would want during this process. `ForceNew` was 
(naturally) controlled in schema, so there was not much we could do
here.

Some recent changes to diff customization has allowed us to change this
process in a way that has been relatively non-invasive. This has 
resulted in the changes below:

* Rather than using schema to force new resources for clone attributes,
we now go through changed keys in the `clone` namespace and `ForceNew` these
keys through diff customization. This happens for both new and existing
resources to prevent diff mismatches.
* We leverage `imported` as a conditional to provide a one-time exemption
from this behavior. We now also transition `imported` from `true` to `false`
after the first apply post-import/migration.

These two changes have the effect in that on the first apply operation
after a resource has been imported or transitioned from a legacy version
of the provider, all `clone` configuration settings are simply persisted
to state, without forcing a new resource. This, like other post-import
operations like flagging `keep_on_remove` for disks, is a no-op on the 
vSphere API side and does not trigger a reconfigure or a rebuild of the 
VM.

The new behavior has been documented. It should be noted that if
configuration changes to `clone` are made in the first post-apply run they
will be ignored permanently as well (as they will persist to state
without any action carried out on them), however this should be
avoidable when import/migration instructions are followed. This has been
documented as well.

-----

This PR also includes updates to TF and `mapstructure`/`copystructure` (the latter as dependencies of changes in 0.11.6) that were needed to fix some bugs with `ForceNew` that was preventing this from working. The commits are still in PR and `vendor` should be updated when the fixes are fully upstream.


vancluever:
Ref: #457

PR #461, opened 2018-04-08
Fix Subfolders Documentation

Atomics:
Their is a mistake in the documentation. Attribute name does not exist. It should be using path attribute.
**Error :**
* vsphere_folder.folder: Resource 'vsphere_folder.parent' does not have attribute 'name' for variable 'vsphere_folder.parent.name'

**Terraform Version :**
Terraform v0.11.6
+ provider.vsphere v1.3.3


vancluever:
Thanks for the correction @Atomics!

PR #469, opened 2018-04-12
r/virtual_machine: Ensure datastore_id will always have a diff on new disk sub-resources

vancluever:
This is something that has been an issue for while but is now even more prominent now that datastores don't necessarily need to be specified at all. Owing in part that `datastore_id` is not populated during the initial TF apply when using the default datastore for disks, changing both the default datastore (or now, possibly, datastore clusters) and a key that forces a new resource at the same time will create a diff mismatch.

A couple of approaches can fix this issue:

* The first one being marking the `datastore_id` attribute of the `disk` sub-resource as `Computed`, but this method opens the resource up to other bugs when a new disk is taking the place of an old disk in state, and Optional/Computed ultimately results in the old datastore ID setting being grafted into the new sub-resource.
* The second is to set the datastore_id attribute for new disks that enter the diff to the same <computed> value that we give these resources otherwise when the attribute is not present (example: when the disk is not pinned to a datastore).

The latter approach is what was ultimately taken as it fixes the issue without introducing the regression mentioned in the first.

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
@@ -78,7 +78,6 @@ func DiskSubresourceSchema() map[string]*schema.Schema {
 		"datastore_id": {
 			Type:          schema.TypeString,
 			Optional:      true,
-			Computed:      true,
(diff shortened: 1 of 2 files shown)


vancluever:
Thanks @bill-rich!

Issue #462, opened 2018-04-10
terraform apply hangs on still creating and stop with "timeout waiting for a routeable interface"


yaoBohoussou:
Hi,
I'm trying to create a new VMfrom template using terraform. My guest OS is pfSense (an os for routers/firewall) and I have vmware tools installed . when I run terraform apply, I can see the vm created in vsphere web client and all network interfaces correctly configured, but terraform gets stuck on the creating the VM state and stops with the error `timeout waiting for a routeable interface`
I have this issue only for these OS : [pfSense](https://www.pfsense.org/) and [vyos](https://vyos.io/).

I can't use `"wait_for_guest_net_timeout":0` because i will use provisionners to run configurations, and I can't use static ip address

## Configuration file

`{
    "resource":[
        {
            "vsphere_virtual_machine":{
                "rtr":{

                    "name":"rtr",
                    "num_cpus":1,
                    "resource_pool_id":"${data.vsphere_resource_pool.pool.id}",
                    "guest_id":"${data.vsphere_virtual_machine.Router_template.guest_id}",
                    "memory":1024,
                    "datastore_id":"${data.vsphere_datastore.datastore.id}",
                    "folder":"Virtual_machines",
                    "scsi_type":"lsilogic",
                    "disk":{
                        "eagerly_scrub":"${data.vsphere_virtual_machine.Router_template.disks.0.eagerly_scrub}",
                        "name":"rtr.vmdk",
                        "thin_provisioned":"${data.vsphere_virtual_machine.Router_template.disks.0.thin_provisioned}",
                        "size":"${data.vsphere_virtual_machine.Router_template.disks.0.size}"
                    },
                    "clone":{
                        "template_uuid":"${data.vsphere_virtual_machine.Router_template.id}"
                    },
                    "network_interface":[
                        {
                            "network_id":"${data.vsphere_network.network.id}",
                            "adapter_type":"${data.vsphere_virtual_machine.Router_template.network_interface_types[0]}"
                        },

                        {
                            "network_id":"${data.vsphere_network.network.id}",
                            "adapter_type":"${data.vsphere_virtual_machine.Router_template.network_interface_types[1]}"
                        }
                    
                    ]
                }
            }
        } 
    ],
    "data":[
        {
            "vsphere_datacenter":{
                "dc":{
                    "name":"datacenter"
                }
            }
        },
        {
            "vsphere_datastore":{
                "datastore":{
[...]

Issue #446, opened 2018-03-30
Not getting macOS guest IP correctly


VictorBjelkholm:
### Terraform + vSphere Provider Version
```
Terraform v0.11.5
+ provider.vsphere v1.3.3
```

### Affected Resource(s)

- `vsphere_virtual_machine`

### Terraform Configuration Files

```hcl
resource "vsphere_virtual_machine" "vm" {
  name             = "terraform-test"
  resource_pool_id = "${data.vsphere_resource_pool.pool.id}"
  datastore_id     = "${data.vsphere_datastore.datastore.id}"

  num_cpus = 2
  memory   = 1024
  guest_id = "${data.vsphere_virtual_machine.template.guest_id}"

  network_interface {
    network_id = "${data.vsphere_network.network.id}"
  }

  disk {
    label = "disk0"
    size  = 20
  }

  clone {
    template_uuid = "${data.vsphere_virtual_machine.template.id}"

    customize {
      network_interface {}
    }
  }
}
```

### Expected Behavior

It should have created a new instance and be able to provide to it. `customize.network_interface` is defined empty as we have a DHCP server for assigning IP addresses within the network.

### Actual Behavior

Getting an error as the guest ID doesn't support customizations. `Customization of the guest operating system 'darwin15_64Guest' is not supported in this configuration. Microsoft Vista (TM) and Linux guests with Logical Volume Manager are supported only for recent ESX host and VMware Tools versions. Refer to vCenter documentation for supported configurations.`

However, I can not skip the customize step, as then vSphere won't understand we're using DHCP (is my understanding at least, from reading the documentation).

So we're stuck with either Terraform can't contact the instance after creation, or getting the error message above. 

I'm looking for a way to let vSphere provider use the automatically assigned IP address, but without having to define a empty `customize.network_interface` block within the `clone` block.

jason-azze:
@VictorBjelkholm I don't think the customization block is required. But without it, your provisioned VMs will all end up with the same hostname as your template because one of the things the customization step does is rename the guest so it matches the name of the VM.

Maybe this is not a problem for you, in which case you can just make sure of two things:
1) Your source template has a NIC that is set to use DHCP
2) Terraform doesn't change the NIC type from whatever you have in your template to its default of VMXNET3. That is, if your template has an e1000 NIC and you don't specify either:
```
adapter_type = "e1000"
```
or 
```
[...]

Issue #337, opened 2018-01-03
OpenBSD guest IP detection


braddeicide:
OpenBSD vmware tools doesn't export routes so it never satisfies the WaitForGuestNet.  
I set wait_for_guest_net_timeout = 0 to get passed that.

The host ip is never populated now though:

vsphere_virtual_machine.tf (remote-exec): Connecting to remote host via SSH...
vsphere_virtual_machine.tf (remote-exec):   Host:
vsphere_virtual_machine.tf (remote-exec):   User: vagrant
vsphere_virtual_machine.tf (remote-exec):   Password: true
vsphere_virtual_machine.tf (remote-exec):   Private key: false
vsphere_virtual_machine.tf (remote-exec):   SSH Agent: false

vcenter shows the correct IP for the host.

It looks like vsphere/virtual_machine_guest_structure.go relies on more things that aren't exported, like PrefixLength.

terraform -v
Terraform v0.11.1
+ provider.vsphere v1.1.1

Guest info exported in https://github.com/openbsd/src/blob/master/sys/dev/pv/vmt.c
/* VM guest info keys */
#define VM_GUEST_INFO_DNS_NAME		1
#define VM_GUEST_INFO_IP_ADDRESS	2
#define VM_GUEST_INFO_DISK_FREE_SPACE	3
#define VM_GUEST_INFO_BUILD_NUMBER	4
#define VM_GUEST_INFO_OS_NAME_FULL	5
#define VM_GUEST_INFO_OS_NAME		6
#define VM_GUEST_INFO_UPTIME		7
#define VM_GUEST_INFO_MEMORY		8
#define VM_GUEST_INFO_IP_ADDRESS_V2	9

So we could workaround by detecting openbsd and accepting ipaddress, which is a little hacky?

vancluever:
Hey @braddeicide,

The original philosophy behind waiting for a routeable address is that we need to make sure that one is reported to the provisioner, if at all possible - a non-routeable address is no good if you can't reach it.

I think the way forward here is to loosen this with a flag - something like `wait_for_guest_net_routeable` - which will be `true` by default. If `false`, it will skip the logic that checks to see if the IP address is routeable and just pass on any discovered IPv4 (probably excluding auto-configuration).

I don't think OS detection is necessary here and could be a slippery slope - I'd rather that this be solved through explicit configuration.

I've marked this as an enhancement so that we can work on it accordingly.

Thanks!

sofixa:
[...]

Issue #231, opened 2017-11-01
timeout waiting for routeable interface on v0.3.0+


y0ssar1an:
### Terraform Version

`v0.10.8`

### vSphere Provider Version

`provider.vsphere 0.2.2` (works)
`provider.vsphere 0.3.0` (first version that fails)
`provider.vsphere 0.4.2` (also fails)

### Affected Resource(s)

- `vsphere_virtual_machine`

### Terraform Configuration Files

```hcl
provider "vsphere" {
  user                 = ""
  password             = ""
  vsphere_server       = ""
  allow_unverified_ssl = true
  version              = "0.4.2"
}

resource "vsphere_virtual_machine" "myvm" {
  name = "vm-deleteme"

  datacenter = "DC99"
  cluster    = "myclusters"

  vcpu   = "1"
  memory = "4096"

  dns_servers = ["8.8.8.8"]

  network_interface {
    label = "VM Network"
  }

  disk {
    template  = "template-20171005.1-UbuntuTrusty-DS-10-SAS-1"
    datastore = "DS-10-SAS-1"
    type      = "thin"
  }
}
```

### Debug Output

https://gist.github.com/y0ssar1an/7c27bb3ada41d1d90814254b413a0c36

```
$ ps -ef | grep terraform | grep -v grep
  501 18150 16865   0 11:16AM ttys004    0:00.51 terraform apply plan
  501 18151 18150   0 11:16AM ttys004    0:00.19 /usr/local/Cellar/terraform/0.10.8/bin/terraform apply plan
  501 18152 18151   0 11:16AM ttys004    0:00.67 /Users/ryan/clusters/vsphere2/.terraform/plugins/darwin_amd64/terraform-provider-vsphere_v0.4.2_x4
```

Everything looks fine in vCenter.

![failed_apply](https://user-images.githubusercontent.com/1250684/32291265-c9ebdf8c-bef9-11e7-96ea-c4f9f6948166.png)

### Expected Behavior

The VM should create successfully and `terraform` should return exit code 0.

### Actual Behavior

1) `timeout waiting for routeable interface`.
2) `terraform` returns exit code 1

### Steps to Reproduce

1. `terraform plan -out plan`
2. `terraform apply plan`

### Important Factoids

AFAIK there's nothing unique about our vSphere configuration. We're using Ubuntu 14.04.

s-ramirez:
[...]

PR #470, opened 2018-04-12
r/virtual_machine: Controllable routable network waiter behavior

vancluever:
This adds the wait_for_guest_net_routable setting, which allows for the
modification of the behavior of the network waiter with relation to
whether or not it compares IP addresses it discovers against the default
gateway.

When false, the waiter will ignore any gateways - it will exit
successfully on discovery of any "real" address - in other words, one
that is not link-local/loopback/multicast, regardless of if the address
does not match a discovered default gateway, or regardless if a gateway
has even been discovered yet.

-----

Related: #462, #446, #337, #231, #207.

### DIFF
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -199,19 +199,24 @@ func Properties(vm *object.VirtualMachine) (*mo.VirtualMachine, error) {
 	return &props, nil
 }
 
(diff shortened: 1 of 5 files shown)


rismoney:
How will this impact new nodes:

```
15:28:32   ~ module.mynode.vsphere_virtual_machine.wks
15:28:32       wait_for_guest_net_routable:                               "" => "true"
```
Will it restart?

vancluever:
@rismoney nope! This should not send neither a reconfigure nor attempt to restart a VM. If it does, please report it as a bug!

PR #481, opened 2018-04-17
r/virtual_machine: Split disk Diff

bill-rich:
Split virtual disk diff function into validation and normalization. Some
normalization tasks should not be performed on new or unmatched virtual
disks, but validation should be performed on all.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine_test.go
+++ b/vsphere/resource_vsphere_virtual_machine_test.go
@@ -379,6 +379,46 @@ func TestAccResourceVSphereVirtualMachine_highDiskUnitInsufficientBus(t *testing
 	})
 }
 
+func TestAccResourceVSphereVirtualMachine_highDiskUnitInsufficientBusAddAfterCreation(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccResourceVSphereVirtualMachinePreCheck(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceVSphereVirtualMachineConfigBasic(),
+				Check:  testAccResourceVSphereVirtualMachineCheckExists(true),
+			},
+			{
+				Config:      testAccResourceVSphereVirtualMachineConfigMultiHighBusInsufficientBus(),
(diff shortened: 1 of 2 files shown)


vancluever:
Hey @bill-rich, the problems that we talked about with the diff validation on new disks is happening because the entire diff validation is happening during the stage where we are comparing resources in state with the current state of configuration, remember. I noticed that `ValidateDiff` still has some values in it that rely on old state, and the operations are executed one after the other, so I'm not too sure if this will fix things.

Also, I was thinking that `NormalizeDiff` is probably not the best name for the second function, and possibly `ValidateDiff` is not necessarily either as while it is accurate it kind of assumes that we are only validating once when we are validating more times. Maybe some different naming is in order that reflects these realities?

We should also throw in some `PlanOnly` tests to make sure that validation is functional in the stages we are trying to target here (diffs with new disks being added, so either new resources completely or diffs where we are adding disks).

Thanks!

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -379,6 +379,46 @@ func TestAccResourceVSphereVirtualMachine_highDiskUnitInsufficientBus(t *testing
...
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceVSphereVirtualMachineConfigBasic(),
+				Check:  testAccResourceVSphereVirtualMachineCheckExists(true),
+			},
+			{
+				Config:      testAccResourceVSphereVirtualMachineConfigMultiHighBusInsufficientBus(),
+				ExpectError: regexp.MustCompile("unit_number on disk \"disk1\" too high \\(15\\) - maximum value is 14 with 1 SCSI controller\\(s\\)"),
```
vancluever: Can this be in the first step or is there a reason why the VM needs to be created first?

bill-rich: That one is included to make sure the diff validation is working on updates and when there is already a disk present (disk0). This same case is tested at VM creation at: https://github.com/terraform-providers/terraform-provider-vsphere/blob/8027f5e805e92851893755bffa34be1efb9cb2c0/vsphere/resource_vsphere_virtual_machine_test.go#L362

The two tests added in this commit may be a bit overkill as there isn't a specific past bug that I'm trying to check for, but trying to cover other cases where the behavior is slightly different.

bill-rich: I dropped the two update tests, so its just the original `TestAccResourceVSphereVirtualMachine_highDiskUnitInsufficientBus` again.

PR #482, opened 2018-04-20
Fixing example code (documentation)

pemcconnell:
Fixing example code so that the datacenter_id is retrieved from the previous vsphere_datacenter block. Similar to #459 but in two other pages - I think this is them all


vancluever:
Thanks for the correction @pemcconnell!

PR #484, opened 2018-04-23
Minor grammar correction

davidroberts63:
I was reading through the docs and noticed this.


vancluever:
@davidroberts63 I guess we're not trying to predict the future here, so "See" is definitely better than "For see" . :wink: Thanks for the correction!

PR #487, opened 2018-04-25
r/compute_cluster: New resource

vancluever:
This adds the `vsphere_compute_cluster` resource, which allows one to manage compute clusters within Terraform.

The current supported workflow requires that the hosts already exist in vSphere. We have opted to use this method because it allows for a unified workflow for getting hosts into (requiring that they can be referenced as a `HostSystem` object already) and out of (putting a host into maintenance mode, evacuating it, and moving it to a neutral spot in inventory) a cluster, versus having to support this plus additional methods, like allowing hosts to be added with `AddHost_Task` (which has its own implications, such as passwords needing to be added in state).

Eventually, the latter workflow may be relegated to a `vsphere_host` _resource_, which we could use to handle the lifecycle of a standalone host, which could then be used as a dependency to `vsphere_cluster`. For now, this allows people to add their hosts with a tool like [govc](https://github.com/vmware/govmomi/tree/master/govc), and then get their data as inputs for the cluster with the `vsphere_host` data source.


website/docs/r/compute_cluster.html.markdown
```diff
@@ -0,0 +1,523 @@
...
+  `vmMonitoringOnly`, or `vmAndAppMonitoring`. Default: `vmMonitoringDisabled`.
+* `ha_vm_failure_interval` - (Optional) If a heartbeat from a virtual machine
+  is not received within this configured interval, the virtual machine is
+  marked as failed. The value is in seconds. Default: `30`.
+* `ha_vm_minimum_uptime` - (Optional) The time, in seconds, that HA waits after
+  powering on a virtual machine before monitoring for heartbeats. Default:
+  `120` (2 minutes).
+* `ha_vm_maximum_resets` - (Optional) The maximum number of resets that HA will
+  perform to a virtual machine when responding to a failure event. Default: `3`
+* `ha_vm_maximum_failure_window` - (Optional) The length of the reset window in
```
bill-rich: Is this in minutes or seconds?

website/docs/r/compute_cluster.html.markdown
```diff
@@ -0,0 +1,523 @@
...
+  configured in `ha_vm_maximum_resets`. `-1` means no window, meaning an
+  unlimited reset time is allotted. Default: `-1` (no window).
+
+#### vSphere HA Admission Control settings
+
+The following settings control vSphere HA Admission Control, which controls
+whether or not specific VM operations are permitted in the cluster in order to
+protect the reliability of the cluster. Based on the constraints defined in
+these settings, operations such as power on or migration operations may be
+blocked. To ensure that enough capacity remains to react to host failures.
```
bill-rich: "blocked to ensure..."

website/docs/r/compute_cluster.html.markdown
```diff
@@ -0,0 +1,523 @@
...
+
+The above would import the cluster named `compute-cluster` that is located in
+the `dc1` datacenter.
+
+## vSphere Version Requirements
+
+A large number of settings in the `vsphere_compute_cluster` resource require a
+specific version of vSphere to function. Rather than include warnings at every
+setting or section, these settings are documented below.  Note that this list
+are for cluster-specific attributes only and does not include [`tags`](#tags),
```
bill-rich: "is for..."

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -0,0 +1,1774 @@
...
+// expandClusterConfigSpecEx reads certain ResourceData keys and returns a
+// ClusterConfigSpecEx.
+func expandClusterConfigSpecEx(d *schema.ResourceData, version viapi.VSphereVersion) *types.ClusterConfigSpecEx {
+	obj := &types.ClusterConfigSpecEx{
+		DasConfig: expandClusterDasConfigInfo(d, version),
+		DpmConfig: expandClusterDpmConfigInfo(d),
+		DrsConfig: expandClusterDrsConfigInfo(d),
+	}
+
+	if version.Newer(viapi.VSphereVersion{Product: version.Product, Major: 6, Minor: 5}) {
```
bill-rich: These are all doing the same version check. It looks like they either need to be checking different versions or can be combined into one conditional.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -0,0 +1,1774 @@
...
+		return err
+	}
+	if err := flattenClusterDpmConfigInfo(d, obj.DpmConfigInfo); err != nil {
+		return err
+	}
+	if err := flattenClusterDrsConfigInfo(d, obj.DrsConfig); err != nil {
+		return err
+	}
+
+	if version.Newer(viapi.VSphereVersion{Product: version.Product, Major: 6, Minor: 5}) {
```
bill-rich: Same here as above

vancluever:
Thanks @bill-rich!

PR #492, opened 2018-04-26
d/compute_cluster: New data source

vancluever:
This adds the `vsphere_compute_cluster` data source, which allows one to
look up an existing cluster.

It also closes a convenience gap in the provider which has been a point
of confusion since around 1.0, in that people looking to supply the
`resource_pool_id` for a virtual machine were confused in that they didn't
think they had a resource pool. This now allows someone to specify a
cluster directly, and then use the `resource_pool_id` attribute from this
data source as the input to `vsphere_virtual_machine`.

### DIFF
--- a/vsphere/data_source_vsphere_compute_cluster.go
+++ b/vsphere/data_source_vsphere_compute_cluster.go
@@ -1,0 +1,50 @@
+package vsphere
+
+import (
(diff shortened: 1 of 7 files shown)


vancluever:
Thanks @bill-rich!

PR #508, opened 2018-05-08
New resource: vsphere_compute_cluster_host_group

vancluever:
This is the other group resource, which manages hosts, and is mainly used in the vm/host affinity rules.

This sets the ground work for the cluster rules to be added now.

### DIFF
--- a/vsphere/resource_vsphere_compute_cluster_host_group.go
+++ b/vsphere/resource_vsphere_compute_cluster_host_group.go
@@ -1,0 +1,374 @@
+package vsphere
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/clustercomputeresource"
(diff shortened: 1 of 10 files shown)


vsphere/resource_vsphere_compute_cluster_host_group.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	log.Printf("[DEBUG] %s: Deleted successfully", resourceVSphereComputeClusterHostGroupIDString(d))
+	return nil
+}
+
+func resourceVSphereComputeClusterHostGroupImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	var data map[string]string
+	if err := json.Unmarshal([]byte(d.Id()), &data); err != nil {
+		return nil, err
```
paultyng: add some additional context on this `err` before returning?

vancluever: This will just return the JSON error which will hopefully be clear enough to the user. I think there are a couple of other resources that have similar import patterns that don't necessarily include any additional error context, so might as well just leave it for consistency for now.

PR #512, opened 2018-05-09
Add attribute to create parent directories when creating a virtual disk

luis-silva:
This PR introduces the attribute to create the parent directories in the datastore when creating a `resource_vsphere_virtualdisk`. It uses the same logic seen in `resource_vsphere_file.go` with an extra check to ensure the directory is already created before creating the disk since the 'MakeDirectory' call does not return a task we can wait on.

### DIFF
--- a/vsphere/resource_vsphere_virtual_disk.go
+++ b/vsphere/resource_vsphere_virtual_disk.go
@@ -3,6 +3,7 @@ package vsphere
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"errors"
 	"path"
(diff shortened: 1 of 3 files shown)


luis-silva:
@vancluever I see you have several PRs open and probably know the answer to my question. 
Do I need to tag anyone to notify them of this PR since I'm an 'external' contributor?

vancluever:
@luis-silva thanks for your contribution! Everything looks good and I will be merging it momentarily.

For future reference, no, you don't need to tag or mention anyone for a PR to be reviewed - there's no prioritization system for this kind of thing. Sometimes PRs can take us a few days to get to though, depending on complexity and the current workload of maintainers.

Thanks again!

luis-silva:
@vancluever Thank you for that, I wasn't expecting it to be merged that fast ! 
Just checking what the correct procedure was regarding contributing to this project, cheers.

PR #520, opened 2018-05-10
r/datastore_cluster_vm_anti_affinity_rule: New resource

vancluever:
This contains the full CRUD/tests/docs for the 
`vsphere_datastore_cluster_vm_anti_affinity_rule` resource.

One thing to note here - during testing, it was found that a
`virtual_machine_ids` length of less than 2 is essentially a no-op on the 
cluster, resulting in no rules created. This was causing issues where
the apply was succeeding but no rule was being returned, resulting in
odd errors post-creation. In addition to this, we cannot restrict this
with `MinItems`, as there currently seems to be an issue calculating the 
length of a glob of computed values - ie: 
`vsphere_virtual_machine.vm.*.id`, even when the `count` in
`vsphere_virtual_machine` is static.

If I'm assuming correctly, this is because when the computed/unknown
value is seen, the entire field becomes unknown with a length of 1,
regardless of if the final length is known beforehand.

Further reading can be found at: 
https://github.com/hashicorp/terraform/issues/14677#issuecomment-302772685


vancluever:
Updated!

PR #535, opened 2018-05-24
r/resource_pool: Add resource pool resource

bill-rich:
Add the resource pool resource along with the modiciations required to support the new resource.

### DIFF
--- a/website/docs/r/resource_pool.html.markdown
+++ b/website/docs/r/resource_pool.html.markdown
@@ -1,7 +1,7 @@
 ---
 layout: "vsphere"
 page_title: "VMware vSphere: vsphere_resource_pool"
-sidebar_current: "docs-vsphere-resource-resource-pool"
+sidebar_current: "docs-vsphere-resource-compute-resource-pool"
 description: |-
   Provides a vSphere resource pool resource. This can be used to create and manage resource pools.
 ---
@@ -41,7 +41,7 @@ data "vsphere_compute_cluster" "compute_cluster" {
 }
 
 resource "vsphere_resource_pool" "resource_pool" {
-  name                    = "terraform-compute-cluster-test"
+  name                    = "terraform-resource-pool-test"
   parent_resource_pool_id = "${data.vsphere_compute_cluster.compute_cluster.id}"
 }
 ```
--- a/website/vsphere.erb
+++ b/website/vsphere.erb
@@ -97,6 +97,9 @@
             <li<%= sidebar_current("docs-vsphere-resource-compute-ha-vm-override") %>>
               <a href="/docs/providers/vsphere/r/ha_vm_override.html">vsphere_ha_vm_override</a>
             </li>
+            <li<%= sidebar_current("docs-vsphere-resource-compute-resource-pool") %>>
+              <a href="/docs/providers/vsphere/r/resource_pool.html">vsphere_resource_pool</a>
+            </li>
           </ul>
         </li>
 


vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+		return err
+	}
+	log.Printf("[DEBUG] %s: Deleted successfully", resourceVSphereResourcePoolIDString(d))
+	return nil
+}
+
+// resourceVSphereResourcePoolIDString prints a friendly string for the
+// vsphere_virtual_machine resource.
+func resourceVSphereResourcePoolIDString(d structure.ResourceIDStringer) string {
+	return structure.ResourceIDString(d, "vsphere_resource_pool")
```
vancluever: Replace bare string here with the defined `resourceVSphereResourcePoolName` const.

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+		"memory_expandable":  obj.ExpandableReservation,
+		"memory_limit":       obj.Limit,
+		"memory_shares":      obj.Shares.Shares,
+		"memory_share_level": obj.Shares.Level,
+	})
+}
+
+func expandResourcePoolConfigSpec(d *schema.ResourceData) *types.ResourceConfigSpec {
+	return &types.ResourceConfigSpec{
+		CpuAllocation:    *expandResourcePoolCPUAllocation(d),
```
vancluever: I don't think the indirection here is necessary - `expandResourcePoolCPUAllocation` et al will only be ever returning values for fields that are always required, so the referenced functions should be modified to not return pointers, and the indirection removed here.

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+		MemoryAllocation: *expandResourcePoolMemoryAllocation(d),
```
vancluever: As above re: indirection

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func expandResourcePoolCPUAllocation(d *schema.ResourceData) *types.ResourceAllocationInfo {
```
vancluever: As above re: indirection

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+		ExpandableReservation: structure.GetBoolPtr(d, "cpu_expandable"),
+		Limit: structure.GetInt64Ptr(d, "cpu_limit"),
+		Shares: &types.SharesInfo{
+			Level:  types.SharesLevel(d.Get("cpu_share_level").(string)),
+			Shares: int32(d.Get("cpu_shares").(int)),
+		},
+	}
+}
+
+func expandResourcePoolMemoryAllocation(d *schema.ResourceData) *types.ResourceAllocationInfo {
```
vancluever: As above re: indirection

vsphere/resource_vsphere_resource_pool_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Config: testAccResourceVSphereResourcePoolConfigBasic(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccResourceVSphereResourcePoolCheckExists(true),
+				),
+			},
+		},
+	})
+}
+
+func TestAccResourceVSphereResourcePool_updateToCustom(t *testing.T) {
```
vancluever: See above, re: resource pool settings check.

vsphere/resource_vsphere_resource_pool_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/viapi"
+)
+
+func TestAccResourceVSphereResourcePool_basic(t *testing.T) {
```
vancluever: There should be a check here that checks that the expected configuration settings are present in the resource pool. This would be a function that would check the pertinent (if not all) resource pool settings, possibly in a fashion that it can be re-used in future tests (such as the update test).

vsphere/resource_vsphere_resource_pool_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Config: testAccResourceVSphereResourcePoolConfigNonDefault(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccResourceVSphereResourcePoolCheckExists(true),
+				),
+			},
+		},
+	})
+}
+
+func TestAccResourceVSphereResourcePool_updateToDefaults(t *testing.T) {
```
vancluever: See above, re: resource pool settings check.

vsphere/resource_vsphere_resource_pool_test.go
```diff
@@ -0,0 +1,222 @@
+package vsphere
```
vancluever: There are some missing tests:

* A test for checking that a resource pool can be moved between different roots (with checks that validate the correct parent)
* An import test
* Tests for tagging (see comment in CRUD file regarding tags)

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
+package vsphere
```
vancluever: It appears that tagging is missing from the schema and CRUD.

It looks like resource pools can be tagged, but not have custom attributes assigned to them, so tagging should be added (the appropriate schema fields), but not custom fields. Tests and docs would need to be updated as well.

website/docs/r/resource_pool.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+[ref-vsphere-resource_pools]: https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.resmgmt.doc/GUID-60077B40-66FF-4625-934A-641703ED7601.html
+
+## Example Usage
+
+The following example sets up a resource pool in a compute cluster which uses
+the default settings for CPU and memory reservations, shares, and limits. The
+compute cluster needs to already exist in vSphere.  
+
+```hcl
```
vancluever: The example code needs to be run through `terraform fmt`.

website/docs/r/resource_pool.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the resource pool.
+* `root_resource_pool_id` - (Required) The [managed object ID][docs-about-morefs]
+  of the root resource pool of the compute resource the resource pool is in.
```
vancluever: _Grammar_ this could probaly be re-worded. Suggested something like:

> `parent_resource_pool_id` - (Required) The managed object ID of the parent resource pool. This can be the root resource pool for a cluster or standalone host, or a resource pool itself.

Note the new naming of the field - see the appropriate comment for this field in the schema.

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+const resourceVSphereResourcePoolName = "vsphere_resource_pool"
+
+func resourceVSphereResourcePool() *schema.Resource {
+	s := map[string]*schema.Schema{
+		"name": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Name of resource pool.",
+		},
+		"root_resource_pool_id": {
```
vancluever: This should be changed to `parent_resource_pool_id`, as the current language implies that this field can only be used to supply _root_ resource pools, where in reality you can create resource pools for as many layers of recursion as you need (with each child pool slicing off portions of its parent). Documentation language should be modified as needed as well.

website/docs/r/resource_pool.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+  unreserved resources. Default: `true`
+* `memory_limit` - (Optional) The CPU utilization of a resource pool will not exceed
+  this limit, even if there are available resources. Set to `-1` for unlimited.
+  Default: `-1`
+
+[docs-about-morefs]: /docs/providers/vsphere/index.html#use-of-managed-object-references-by-the-vsphere-provider
+
+## Attribute Reference
+
+The following attributes are exported:
```
vancluever: If the only attribute exported is the ID, change this to the paragraph form as seen in places like [here](https://www.terraform.io/docs/providers/vsphere/r/compute_cluster_host_group.html#attribute-reference).

website/docs/r/resource_pool.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+* `id`: The [managed object ID][docs-about-morefs] of the resource pool.
+
+[docs-about-morefs]: /docs/providers/vsphere/index.html#use-of-managed-object-references-by-the-vsphere-provider
+
+## Importing
+
+An existing resource pool can be [imported][docs-import] into this resource via
+the path to the resource, using the following command:
```
vancluever: > the path to the resource _pool_

vsphere/data_source_vsphere_host.go
```diff
@@ -22,6 +22,11 @@ func dataSourceVSphereHost() *schema.Resource {
 				Description: "The managed object ID of the datacenter to look for the host in.",
 				Required:    true,
 			},
+			"resource_pool_id": &schema.Schema{
```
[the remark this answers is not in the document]
bill-rich: It seems to. I've tested on hosts that are in clusters and they still have root resource pools.

vsphere/data_source_vsphere_host.go
```diff
@@ -38,9 +43,12 @@ func dataSourceVSphereHostRead(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return fmt.Errorf("error fetching host: %s", err)
 	}
-
+	rp, err := hostsystem.ResourcePool(hs)
+	if err != nil {
+		return err
+	}
+	_ = d.Set("resource_pool_id", rp.Reference().Value)
```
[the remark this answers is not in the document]
bill-rich: Sounds good. For whatever reason I had it in my head to ignore the Set errors. I'll get those fixed up.

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	return &schema.Resource{
+		Create: resourceVSphereResourcePoolCreate,
+		Read:   resourceVSphereResourcePoolRead,
+		Update: resourceVSphereResourcePoolUpdate,
+		Delete: resourceVSphereResourcePoolDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceVSphereResourcePoolImport,
+		},
+		SchemaVersion: 3,
```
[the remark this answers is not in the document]
bill-rich: This was originally put in to learn about schema versions, but it got missed when the resource ended up growing so much. Got it removed now. Thanks for the info on how its supposed to be used. Seems it served its purpose anyway!

vsphere/resource_vsphere_resource_pool_test.go
```diff
@@ -0,0 +1,727 @@
...
+	return func(s *terraform.State) error {
+		rp, err := testGetResourcePool(s, "resource_pool")
+		if err != nil {
+			return err
+		}
+		prp, err := testGetResourcePool(s, parent)
+		if err != nil {
+			return err
+		}
+		props, err := resourcepool.Properties(rp)
```
vancluever: I usually consolidate these into a single properties function to avoid the extra error check. See [here](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/helper_test.go#L765).

website/docs/r/resource_pool.html.markdown
```diff
@@ -0,0 +1,116 @@
...
+
+[docs-about-morefs]: /docs/providers/vsphere/index.html#use-of-managed-object-references-by-the-vsphere-provider
+[docs-applying-tags]: /docs/providers/vsphere/r/tag.html#using-tags-in-a-supported-resource
+
+## Attribute Reference
+
+The only attribute this resource exports is the `id` of the resource, which is
+the [managed object ID][docs-about-morefs] of the resource pool.
+
+[docs-about-morefs]: /docs/providers/vsphere/index.html#use-of-managed-object-references-by-the-vsphere-provider
```
vancluever: Reference slugs in Markdown don't need to be supplied more than once, this can be omitted.

vancluever:
This looks good now - I just made sure the docs were available via the sidebar, but other than that looks great! Good work @bill-rich!

PR #541, opened 2018-06-03
r/resource_pool: Correct docs

marsavela:
parent_resource_pool_id should reference to the resource_pool_id output of data.vsphere_compute_cluster and not to the id.


bill-rich:
Looks good. Thanks for the contribution!

Issue #558, opened 2018-06-21
Cannot get any attribute from vsphere_virtual_machine


gaeldb:
Hi there,

When using data.vpshere_virutal_machine on a template vm, I cannot get any attribute from it (id, guest_id, ...).
Strange thing : VM UUID seems to be empty in debug trace.

### Terraform Version && vSphere Provider Version

Terraform v0.11.7
provider.vsphere v1.6.0

### Affected Resource(s)

- `vsphere_virtual_machine`

### Terraform Configuration Files

```hcl
data "vsphere_datacenter" "datacenter" {
    name = "mydc"
}

data "vsphere_virtual_machine" "template" {
  name          = "templatecentos7"
  datacenter_id = "${data.vsphere_datacenter.datacenter.id}"
}

output "uuid" {
  value = "${data.vsphere_virtual_machine.template.id}"
}

output "guest_id" {
  value = "${data.vsphere_virtual_machine.template.guest_id}"
}
```

### Debug Output

```
2 error(s) occurred:

* output.guest_id: Resource 'data.vsphere_virtual_machine.template' not found for variable 'data.vsphere_virtual_machine.template.guest_id'
* output.uuid: Resource 'data.vsphere_virtual_machine.template' not found for variable 'data.vsphere_virtual_machine.template.id'
```
Full debug output : https://gist.github.com/gaeldb/344c2325280a65c430971dedf08a0476
where UUID seems to be empty : 
```
[DEBUG] VM search for "templatecentos7" completed successfully (UUID "")
[ERROR] root: eval: *terraform.EvalWriteOutput, err: Resource 'data.vsphere_virtual_machine.template' not found for variable 'data.vsphere_virtual_machine.template.guest_id'
```

Thanks a lot !

gaeldb:
In data_source_vsphere_virtual_machine.go line 105, _virtualmachine.Properties(vm)_ return an empty string Uuid.
Exploring why...

gaeldb:
Ow, govc doesn’t know the UUID neither...

$ govc vm.info templatecentos7
```
Name:           templatecentos7
  Path:         /RD Datacenter/vm/templates/templatecentos7
  UUID:         
  Guest name:   Red Hat Enterprise Linux 7 (64-bit)
  Memory:       4096MB
  CPU:          2 vCPU(s)
  Power state:  poweredOff
  Boot time:    <nil>
  IP address:   
  Host:         xxxxxxxxx 
```

[...]

PR #559, opened 2018-06-21
vsphere_virtual_machine: Add a test to catch empty UUID

gaeldb:
Prevent #558 (which is not a bug)

### DIFF
--- a/vsphere/data_source_vsphere_virtual_machine.go
+++ b/vsphere/data_source_vsphere_virtual_machine.go
@@ -109,6 +109,10 @@ func dataSourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("no configuration returned for virtual machine %q", vm.InventoryPath)
 	}
 
+	if props.Config.Uuid == "" {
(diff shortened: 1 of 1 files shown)


vsphere/data_source_vsphere_virtual_machine.go
```diff
@@ -109,6 +109,10 @@ func dataSourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("no configuration returned for virtual machine %q", vm.InventoryPath)
 	}
 
+	if props.Config.Uuid == "" {
+		return fmt.Errorf("configuration returned for virtual machine %q is scrappy: any UUID", vm.InventoryPath)
```
vancluever: Can this error message be changed?  I'd suggest something like:

```
return fmt.Errorf("virtual machine %q does not have a UUID", vm.InventoryPath)
```

gaeldb: Agree with you, let's go.

gaeldb:
I changed error message

PR #566, opened 2018-07-13
r/vapp_container: Add vapp_container resource

bill-rich:


vancluever:
@bill-rich could this PR be cleaned up a bit? I see a bunch of my commits in there. Maybe rebase your current changes on master?

```
git checkout f-vapp-container
git fetch origin master
git rebase master
```

That should make things a lot clearer and may resolve merge conflicts as well.

bill-rich:
@vancluever Thanks! I missed that. Got it sorted now.

vsphere/internal/helper/storagepod/storage_pod_helper.go
```diff
@@ -305,16 +335,16 @@ func RelocateVM(
...
-	_, err = recommendAndApplySDRS(client, sps, time.Minute*time.Duration(timeout))
+	placement, err := recommendSDRS(client, sps, time.Minute*time.Duration(timeout))
+	if err != nil {
+		return err
+	}
+	_, err = applySDRS(client, placement, time.Minute*time.Duration(timeout))
 	return err
 }
 
-func recommendAndApplySDRS(
```
vancluever: Rather than deleting this function, I'd recommend taking the de-coupled logic that has been duplicated to other parts and just throwing that here. Then the rest of the places that reference `recommendAndApplySDRS` that don't require a special path to address vApp containers don't need changes at all.

vsphere/internal/helper/storagepod/storage_pod_helper.go
```diff
@@ -228,7 +250,11 @@ func CloneVM(
 		Type: string(types.StoragePlacementSpecPlacementTypeClone),
 	}
 
-	return recommendAndApplySDRS(client, sps, time.Minute*time.Duration(timeout))
```
vancluever: See comment on `recommendAndApplySDRS`

vsphere/internal/helper/storagepod/storage_pod_helper.go
```diff
@@ -268,7 +294,11 @@ func ReconfigureVM(
 		ConfigSpec: &spec,
 	}
 
-	_, err = recommendAndApplySDRS(client, sps, provider.DefaultAPITimeout)
```
vancluever: See comment on `recommendAndApplySDRS`

vsphere/internal/helper/storagepod/storage_pod_helper.go
```diff
@@ -305,16 +335,16 @@ func RelocateVM(
 		Type:         string(types.StoragePlacementSpecPlacementTypeRelocate),
 	}
 
-	_, err = recommendAndApplySDRS(client, sps, time.Minute*time.Duration(timeout))
```
vancluever: See comment on `recommendAndApplySDRS`

vsphere/resource_vsphere_resource_pool.go
```diff
@@ -153,7 +153,7 @@ func resourceVSphereResourcePoolCreate(d *schema.ResourceData, meta interface{})
 	}
 	d.SetId(rp.Reference().Value)
 	log.Printf("[DEBUG] %s: Create finished successfully", resourceVSphereResourcePoolIDString(d))
-	return nil
+	return resourceVSphereResourcePoolRead(d, meta)
```
vancluever: Good find! :+1:

vsphere/resource_vsphere_vapp_container.go
```diff
@@ -0,0 +1,369 @@
...
+		f, err = folder.FromID(client, pf.(string))
+		if err != nil {
+			return err
+		}
+	} else {
+		p := strings.Split(prp.InventoryPath, "/")
+		if len(p) > 2 {
+			return fmt.Errorf("unable to locate datacenter name from parent resource pool")
+		}
+		dc, err := getDatacenter(client, p[1])
```
vancluever: This gives me a warning on `vetshadow` as a redundant `err` declaration and can probably be fixed to just be an assignment without declaration (`=` instead of `:=`).

vsphere/resource_vsphere_vapp_container.go
```diff
@@ -0,0 +1,369 @@
...
+	if err != nil {
+		return err
+	}
+	if err = resourceVSphereVAppContainerApplyTags(d, meta, vc); err != nil {
+		return err
+	}
+	op, np := d.GetChange("parent_resource_pool_id")
+	if op != np {
+		log.Printf("[DEBUG] %s: Parent resource pool has changed. Moving from %s, to %s", resourceVSphereVAppContainerIDString(d), op, np)
+		p, err := resourcepool.FromID(client, np.(string))
```
vancluever: This gives me a warning on `vetshadow` as a redundant err declaration and can probably be fixed to just be an assignment without declaration (`=` instead of `:=`).

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -310,12 +311,15 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 	if vprops.ResourcePool != nil {
 		d.Set("resource_pool_id", vprops.ResourcePool.Value)
 	}
-	// Set the folder
-	f, err := folder.RootPathParticleVM.SplitRelativeFolder(vm.InventoryPath)
-	if err != nil {
-		return fmt.Errorf("error parsing virtual machine path %q: %s", vm.InventoryPath, err)
+	// If the VM is part of a vApp, the inventory path will not point to a host
+	// path rather than a VM path, so this step must be skipped.
```
vancluever: I'm having issues understanding this - what does the path of a VM in a vApp container look like then, and how should we be operating on it?

website/docs/r/vapp_container.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+layout: "vsphere"
+page_title: "VMware vSphere: vsphere_vapp_container"
+sidebar_current: "docs-vsphere-resource-compute-vapp-container"
+description: |-
+  Provides a vSphere vApp container resource. This can be used to create and manage vApp container.
+---
+
+# vsphere\_vapp\_container
+
+The `vsphere_vApp_container` resource can be used to create and manage
```
vancluever: Should be all lowercase here - `vsphere_vapp_container`.

website/docs/r/vapp_container.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+Virtual Appliances.
```
vancluever: This is not accurate, correct? Eventually we might allow OVAs through it (which could allow for the deploy of multi-VM vApps), but currently it's just a container.

website/docs/r/vapp_container.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+For more information on vSphere Virtual Appliances, see [this
```
vancluever: Terminology also here. Might be worth it to scan the doc for all reference of "Virtual Appliance" and just replace with "vApp".

website/docs/r/vapp_container.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the vApp container.
+* `parent_resource_pool_id` - (Required) The [managed object ID][docs-about-morefs]
+  of the parent resource pool. This can be the root resource pool for a cluster
+  or standalone host, or a resource pool itself. When moving a vApp container
+  from one parent resource pool to another, both must share a common root
+  resource pool or the move will fail.
+* `parent_folder_id` - (Optional) The [manage object ID][docs-about-morefs] of
```
vancluever: `s/manage/managed/`

vsphere/resource_vsphere_vapp_container_test.go
```diff
@@ -0,0 +1,948 @@
...
+			return err
+		}
+		if vm.ParentVApp != nil && vm.ParentVApp.Reference() != vc.Reference() {
+			return fmt.Errorf("VM is not a part of vApp container")
+		}
+		return nil
+	}
+}
+
+func testGetVAppContainer(s *terraform.State, resourceName string) (*object.VirtualApp, error) {
```
vancluever: Can this be thrown into `helper_test.go` just in case it's needed by other tests other than vApp container tests?

This is where most of the other test helper code is stored.

website/docs/r/vapp_container.html.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+The `vsphere_vApp_container` resource can be used to create and manage
+Virtual Appliances.
+
+For more information on vSphere Virtual Appliances, see [this
+page][ref-vsphere-vapp].
+
+[ref-vsphere-vapp]: https://docs.vmware.com/en/VMware-vSphere/6.5/com.vmware.vsphere.vm_admin.doc/GUID-2A95EBB8-1779-40FA-B4FB-4D0845750879.html
+
+## Example Usage
```
vancluever: Do you think it's worthwhile to add another example on how to deploy a VM into a vApp container? It may not be exactly obvious from the docs to use the ID of the container as the resource pool.

vsphere/resource_vsphere_vapp_container.go
bill-rich: I've been ignoring those in cases like this because dc hasn't been declared yet. I considered declaring it on its own line and then just assigning both, but it seemed less clear that way. I'm open to either way.

vsphere/internal/helper/storagepod/storage_pod_helper.go
bill-rich: The reason I did it this way was because I still need the recommendation in order to determine which datastore to use in the case where the VM is being created in a vApp. I didn't want to clutter up recommendAndApplySDRS with the vApp code. Does that make sense? Very possible I'm just overlooking something simple.

vsphere/resource_vsphere_vapp_container_test.go
bill-rich: Done

vsphere/internal/helper/storagepod/storage_pod_helper.go
vancluever: I'm not saying to actually revert the de-coupling, but rather than removing the function altogether and then duplicating code a few places, just change recommendAndApplySDRS to:

```go
func recommendAndApplySDRS(
	client *govmomi.Client,
	sps types.StoragePlacementSpec,
	timeout time.Duration,
) (*object.VirtualMachine, error) {
	placement, err := recommendSDRS(client, sps, time.Minute*time.Duration(timeout))
	if err != nil {
		return nil, err
	}
	return applySDRS(client, placement, time.Minute*time.Duration(timeout))
}
```

This way, you don't have to update the other places that this is being used (where you don't need the recommendation), such as https://github.com/terraform-providers/terraform-provider-vsphere/pull/566#discussion_r202789715, https://github.com/terraform-providers/terraform-provider-vsphere/pull/566#discussion_r202789732, or https://github.com/terraform-providers/terraform-provider-vsphere/pull/566#discussion_r202789745.

vsphere/resource_vsphere_vapp_container.go
vancluever: Yeah, declare it. It's a minor inconvenience to get this (and the only other linter hit in this file) gone so that there's eventually less noise in the linter. Might as well have the file completely clean.

vsphere/resource_vsphere_vapp_container.go
vancluever: Just referencing https://github.com/terraform-providers/terraform-provider-vsphere/pull/566#discussion_r203573010 here

vsphere/resource_vsphere_virtual_machine.go
bill-rich: I made some changes so the folder isn't skipped when the VM is in a vApp container, but it also cannot be moved to a new folder since it results in an error.

* Is this the inventory path returned by InventoryPath (assuming yes here) - Yes it is.
* Can you see the VM still by govc ls /dc1/vm/ - This does still work
* What happens to the path if you move a VM that belongs to a vApp container to a folder - You cannot. It results in an error.
* Can we make a helper that works off of the above? - Not certain on what you mean here. I did make it so that the virtual machine's VM folder is what `folder` is populated with.

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -636,6 +652,11 @@ func resourceVSphereVirtualMachineCustomizeDiff(d *schema.ResourceDiff, meta int
 	if err := virtualdevice.DiskDiffOperation(d, client); err != nil {
 		return err
 	}
+	// When a VM is a member of a vApp container, it is no longer part of the VM
+	// tree, and therefor cannot have its VM folder set.
```
vancluever: `s/therefor/therefore/`

Issue #542, opened 2018-06-04
Question - maybe an Enhancement Request - 


surajsub:
Hi there,

### Terraform Version
0.11.3

### vSphere Provider Version
1.5

### Affected Resource(s)
- `vsphere_virtual_machine`

Question
When a scsi controller is created we can set the scsi_type to the supported 3 values. What would it take to adjust / set the scsi bus sharing param as well. We are required to create some shared disks and would like to have the capability to set the SCSI Bus Sharing mode to Physical - The default setting is None and cannot be altered.

Any thoughts if it's worth a minor fix ?

vancluever:
Hey @surajsub, I think the feasibility of this depends on whether or not a shared SCSI bus means that _all_ luns are shared as well. If it is, more than likely it will not be possible as this also means that the two virtual machines would have to share _state_, which is counter to Terraform's patterns, unfortunately.

Just going to leave this one open until this is 100% determined, but I would probably look at trying to do this another way.

surajsub:
@vancluever  - sure. This seems to be a very common method to implement the suse pacemaker cluster for sap  . We can  however make sure that there is only a dedicated scsi controller for this shared disk and no other disks can be attached to that controller.

vancluever:
@surajsub one thing that I was thinking about was that this might be a good use case for the [`disk_sharing`](https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#disk_sharing) parameter in the `disk` sub-resource. The disk would be created with the `vsphere_virtual_disk` resource, and then attached to both VMs with this setting set. Have you tried that by any chance yet?

surajsub:
I have that as well 
```
disk 
{
	attach = true
	path = "${vsphere_virtual_disk.disk_1.vmdk_path}"
	label = "diskshared"
	unit_number = 2
	datastore_id = "${data.vsphere_datastore.datastore.id}"
	disk_sharing = "sharingMultiWriter"
	 disk_mode = "independent_persistent"
	
	}
```
The rationale for bus_sharing  = Physical is because when drs is enabled and vms land on other hosts they don't loose the bus connection

surajsub:
I checked with the SAP Susepacemaker team as well. They do suggest this since during maintenance of the infra the vm can move anywhere in the esxcluster and they do not want the bus connection to be broken.

surajsub:
@vancluever  - Any additional thoughts on this ?

surajsub:
[...]

PR #574, opened 2018-07-27
r/virtual_machine: Support SCSI bus sharing

bill-rich:
Adds feature requested in #542


vancluever:
Hey @bill-rich, just reviewing the example here again. I was under the impression that the disk sharing would happen via a `vsphere_virtual_disk` resource and `attach`. It seems the example creates a datastore and puts a "shared" disk there. Have we tested this example to make sure that this is indeed the same data?

Let's chat this morning about this so that we can get it in.

Thanks!

PR #582, opened 2018-08-07
r/virtual_machine: Ignore empty controller specs

bill-rich:
Only append config specs (cspecs) if they are not empty.


bill-rich:
@vancluever I didn't create an issue for it, but its something I picked up running acceptance tests after #574. When the SCSI bus sharing doesn't need to be changed, an empty controller spec is still returned and appended to the spec slice which results in a nil value error later.

PR #583, opened 2018-08-09
Relocate VMs after create

bill-rich:
The host attribute of CreateVM_Task is being ignored in vCenter 6.7.
Adding a host check and reloaction step to ensure that VMs land on
the correct host.


bill-rich:
Good call. It was caught originally in the vMotion tests, but now I added on in to check specifically for this issue.

skywalk7:
Please always check if parameter was actually set with `GetOk` before comparing value.
This code forces VM vmotion even if host_system_id was not set.
```
if d.Get("host_system_id").(string) != vprops.Runtime.Host.Reference().Value {
```

PR #625, opened 2018-09-25
Vswitch import

pablo-ruth:
This update adds import functionality on ```vsphere_host_virtual_switch ``` resource.

### DIFF
--- a/vsphere/host_network_system_helper.go
+++ b/vsphere/host_network_system_helper.go
@@ -45,6 +45,7 @@ func hostVSwitchFromName(client *govmomi.Client, ns *object.HostNetworkSystem, n
 
 	for _, sw := range mns.NetworkInfo.Vswitch {
 		if sw.Name == name {
+			sw.Spec.Mtu = sw.Mtu
 			return &sw, nil
 		}
 	}
--- a/vsphere/host_network_system_helper.go
+++ b/vsphere/host_network_system_helper.go
@@ -45,6 +45,9 @@ func hostVSwitchFromName(client *govmomi.Client, ns *object.HostNetworkSystem, n
 
 	for _, sw := range mns.NetworkInfo.Vswitch {
 		if sw.Name == name {
+			// Spec.Mtu is not set for vSwitches created directly
+			// in ESXi. TODO: Use sw.Mtu instead of sw.Spec.Mtu in
+			// flattenHostVirtualSwitchSpec.
 			sw.Spec.Mtu = sw.Mtu
 			return &sw, nil
 		}
--- a/vsphere/resource_vsphere_host_virtual_switch.go
+++ b/vsphere/resource_vsphere_host_virtual_switch.go
@@ -149,11 +149,6 @@ func resourceVSphereHostVirtualSwitchImport(d *schema.ResourceData, meta interfa
 	d.Set("host_system_id", hostID)
 	d.Set("name", switchName)
 
-	err = resourceVSphereHostVirtualSwitchRead(d, meta)
-	if err != nil {
-		return []*schema.ResourceData{}, err
-	}
-
 	return []*schema.ResourceData{d}, nil
 }
 
--- a/vsphere/resource_vsphere_host_virtual_switch.go
+++ b/vsphere/resource_vsphere_host_virtual_switch.go
@@ -148,12 +148,12 @@ func resourceVSphereHostVirtualSwitchImport(d *schema.ResourceData, meta interfa
 
 	err = d.Set("host_system_id", hostID)
 	if err != nil {
-		return err
+		return []*schema.ResourceData{}, err
 	}
 
 	err = d.Set("name", switchName)
 	if err != nil {
-		return err
+		return []*schema.ResourceData{}, err
 	}
 
 	return []*schema.ResourceData{d}, nil
--- a/vsphere/resource_vsphere_host_virtual_switch.go
+++ b/vsphere/resource_vsphere_host_virtual_switch.go
@@ -48,7 +48,10 @@ func resourceVSphereHostVirtualSwitch() *schema.Resource {
 		Update:        resourceVSphereHostVirtualSwitchUpdate,
 		Delete:        resourceVSphereHostVirtualSwitchDelete,
 		CustomizeDiff: resourceVSphereHostVirtualSwitchCustomizeDiff,
-		Schema:        s,
+		Importer: &schema.ResourceImporter{
+			State: resourceVSphereHostVirtualSwitchImport,
+		},
+		Schema: s,
 	}
 }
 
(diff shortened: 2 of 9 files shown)


vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -137,6 +140,23 @@ func resourceVSphereHostVirtualSwitchDelete(d *schema.ResourceData, meta interfa
...
+func resourceVSphereHostVirtualSwitchImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	hostID, switchName, err := splitHostVirtualSwitchID(d.Id())
+	if err != nil {
+		return []*schema.ResourceData{}, err
+	}
+
+	d.Set("host_system_id", hostID)
+	d.Set("name", switchName)
+
+	err = resourceVSphereHostVirtualSwitchRead(d, meta)
```
bill-rich: No need to call `Read` here as it will be called after import.

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -137,6 +140,23 @@ func resourceVSphereHostVirtualSwitchDelete(d *schema.ResourceData, meta interfa
 	return nil
 }
 
+func resourceVSphereHostVirtualSwitchImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	hostID, switchName, err := splitHostVirtualSwitchID(d.Id())
+	if err != nil {
+		return []*schema.ResourceData{}, err
+	}
+
+	d.Set("host_system_id", hostID)
```
bill-rich: The `error` returned from `d.Set` should be handled.

vsphere/resource_vsphere_host_virtual_switch.go
```diff
@@ -137,6 +140,23 @@ func resourceVSphereHostVirtualSwitchDelete(d *schema.ResourceData, meta interfa
...
+	d.Set("name", switchName)
```
bill-rich: The `error` returned from `d.Set` should be handled.

website/docs/r/host_virtual_switch.html.markdown
```diff
@@ -174,3 +174,16 @@ the resource. This is set to an ID value unique to Terraform - the convention
 is a prefix, the host system ID, and the virtual switch name. An example would
 be `tf-HostVirtualSwitch:host-10:vSwitchTerraformTest`.
 
+## Importing
+
+An existing vSwitch can be [imported][docs-import] into this resource via the id
```
bill-rich: The `id` for vswitches isn't very intuitive. I think having an explaination that the ID is the <prefix (tf-HostVirtualSwitch)>:<host MOID>:<switch name> would help make it more clear. Sorry the `id` isn't simpler in this case!

website/docs/r/host_virtual_switch.html.markdown
```diff
@@ -174,3 +174,18 @@ the resource. This is set to an ID value unique to Terraform - the convention
 is a prefix, the host system ID, and the virtual switch name. An example would
 be `tf-HostVirtualSwitch:host-10:vSwitchTerraformTest`.
 
+## Importing
+
+An existing vSwitch can be [imported][docs-import] into this resource by its id.
+The convention of the id is a prefix, the host system ID, and the virtual switch
```
bill-rich: One last thing on this, can ```host system ID``` be changed to ```host system [managed objectID][docs-about-morefs]```? And add:
```[docs-about-morefs]: /docs/providers/vsphere/index.html#use-of-managed-object-references-by-the-vsphere-provider```

There are several different ID types used throughout the provider, so it is just helpful to be extra clear when specifying.

File: website/docs/r/host_virtual_switch.html.markdown [hunk 6]
```diff
@@ -174,3 +174,18 @@ the resource. This is set to an ID value unique to Terraform - the convention
 is a prefix, the host system ID, and the virtual switch name. An example would
 be `tf-HostVirtualSwitch:host-10:vSwitchTerraformTest`.
 
+## Importing
+
+An existing vSwitch can be [imported][docs-import] into this resource by its id.
```
bill-rich: id -> ID

vsphere/host_network_system_helper.go
```diff
@@ -45,6 +45,7 @@ func hostVSwitchFromName(client *govmomi.Client, ns *object.HostNetworkSystem, n
 
 	for _, sw := range mns.NetworkInfo.Vswitch {
 		if sw.Name == name {
+			sw.Spec.Mtu = sw.Mtu
```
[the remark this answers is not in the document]
bill-rich: Thats strange because the `mtu` is being read from the `sw.Mtu` instead of `sw.Spec.Mtu`. When testing in my environment, the MTU seems to be properly read without this. Does it happen after every import? If you set the MTU in vCenter, do you still get `0` in terraform?

vsphere/resource_vsphere_host_virtual_switch_test.go
```diff
@@ -187,6 +187,41 @@ func TestAccResourceVSphereHostVirtualSwitch_changeToStandby(t *testing.T) {
...
+				Config: testAccResourceVSphereHostVirtualSwitchConfig(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccResourceVSphereHostVirtualSwitchExists(true),
+				),
+			},
+			{
+				ResourceName:      "vsphere_host_virtual_switch.switch",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateIdFunc: func(s *terraform.State) (string, error) {
```
pablo-ruth: I'm not sure what I have to do in this section...

bill-rich: That function just needs to return the string you would use when importing the resource using ```terraform import```. In this case, it is pretty straight forward since we're just using the `id`. The function below returns the `id` from the switch that was created in the first test step.
```
ImportStateIdFunc: func(s *terraform.State) (string, error) {
  vars, err := testClientVariablesForResource(s, fmt.Sprintf("vsphere_host_virtual_switch.%s", "switch"))
  if err != nil {
    return "", err
  } 
  return vars.resourceID, err
},
```

vsphere/host_network_system_helper.go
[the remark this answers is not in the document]
bill-rich: It seems to be related to if the vSwitch is created through vCenter or ESXi. I think the ideal way to fix this will be to change the `read` behavior so that the MTU is pulled from the switch managed object instead of the spec, but that is probably better left for another PR. This seems like a good way to deal with it until then. Thanks for documenting your method so well! It made it really easy to pinpoint the difference.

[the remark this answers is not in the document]
pablo-ruth: ok

vsphere/resource_vsphere_host_virtual_switch_test.go
pablo-ruth: Ah it's simpler than what I thought. :) Thanks

PR #627, opened 2018-09-26
Make format 1.11 compatible

bill-rich:
Run gofmt on existing files to be 1.11 format compatible.


paddycarver:
I approved, but you probably want to update your Travis config as well, to fix the check.

PR #668, opened 2018-12-05
Use correct function to get TagsClient

arnisoph:
TagsClient can be nil, so when tags aren't defined at all don't use it.


bill-rich:
Hi @arnisoph! Can you provide some additional background on the issue you're addressing here? In the original `Read` code for the folder resource, there is still a check for if the tags client in `nil`.

tagsClientIfDefined is only used in Create and Update because it only returns the tags client if there are tags defined.  In the proposed change, new tags will not be picked up on read unless there are already tags defined for the resource.

arnisoph:
Hi, sorry for the late reply! I can see your point and will provide feedback during the next days!

bill-rich:
I'm going to close this PR for now since there hasn't been any activity. We can continue to discuss here or a new PR can be created. Thanks!

PR #684, opened 2019-01-08
Add support for using guest.ipAddress for older versions of VM Tools

cbascom:
### Description
There are two parts to this change:

- Add a new waiter that waits for guest.ipAddress to be populated. This waiter is controlled via the `wait_for_guest_ip_timeout` property that is disabled by default. This waiter is intended to be used in place of the `wait_for_guest_net_timeout` waiter.
- Regardless of which waiter is enabled, the code that populates `default_ip_address` and `guest_ip_addresses` will first look at `guest.net` and `guest.ipStack` to find the IP address. Only if no IP address is found via that method will it then fall back to looking at the `guest.ipAddress` property instead.

### Use Case
The VMs I am creating unfortunately have an older version of VM Tools that does not populate the `guest.net` or `guest.ipStack` properties. However, the VMs I am creating do populate the `guest.ipAddress` property which can be used for this same purpose. This change allows me to just wait for `guest.ipAddress` and use that IP address to proceed with provisioning.

### Todo

- Tests
- Documentation updates

I wanted to see if there was a chance of getting something like this in before I spent time adding those as well.

### DIFF
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -210,7 +210,7 @@ func Properties(vm *object.VirtualMachine) (*mo.VirtualMachine, error) {
 //
 // The timeout is specified in minutes. If zero or a negative value is passed,
 // the waiter returns without error immediately.
-func WaitForGuestIP(client *govmomi.Client, vm *object.VirtualMachine, timeout int) error {
+func WaitForGuestIP(client *govmomi.Client, vm *object.VirtualMachine, timeout int, ignoredGuestIPs []interface{}) error {
 	if timeout < 1 {
 		log.Printf("[DEBUG] Skipping IP waiter for VM %q", vm.InventoryPath)
 		return nil
@@ -236,7 +236,7 @@ func WaitForGuestIP(client *govmomi.Client, vm *object.VirtualMachine, timeout i
 			}
 
 			ip := net.ParseIP(c.Val.(string))
-			if skipIPAddrForWaiter(ip) {
+			if skipIPAddrForWaiter(ip, ignoredGuestIPs) {
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


bill-rich:
Hi @cbascom! 

Thanks for asking about this. Which version is guest.ipAddress supported in? I don't know right off where the cutoff is. I'd like to support as much as possible with the vSphere provider, but we can't go too far back without adding a lot of potential issues and complexity.

cbascom:
@bill-rich The version of VMTools I am forced to deal with for now is 9.4.6.33107 (build-1770165). Based on the VMware docs I have looked at, I think it has been available from the beginning. The guest.ipStack that is currently being used wasn't added until API version 4.1. FWIW, The guest.ipAddress field is what the vSphere client uses to display the primary IP address of the VM I believe.

I understand about not wanting to go too far back, if this is a no go I will just have to maintain my fork for the time being. If you have any other ideas on how I can work around being stuck with the older VMTools, I'm all ears - this was the only way I have come up with so far that gets me what I need. Thanks!

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -206,14 +206,66 @@ func Properties(vm *object.VirtualMachine) (*mo.VirtualMachine, error) {
 	return &props, nil
 }
 
+// WaitForGuestIP waits for a virtual machine to have an IP address.
+//
+// The timeout is specified in minutes. If zero or a negative value is passed,
+// the waiter returns without error immediately.
+func WaitForGuestIP(client *govmomi.Client, vm *object.VirtualMachine, timeout int, ignoredGuestIPs []interface{}) error {
```
bill-rich: What does this do differently than `WaitForGuestNet` with `wait_for_guest_net_routable` set to `false`?

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -103,6 +109,12 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 			Default:     true,
 			Description: "Controls whether or not the guest network waiter waits for a routable address. When false, the waiter does not wait for a default gateway, nor are IP addresses checked against any discovered default gateways as part of its success criteria.",
 		},
+		"ignored_guest_ips": {
```
bill-rich: Please add this to the documentation for `virtual_machine` resource.

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
cbascom: The only real difference is that it waits for the string guest.ipAddress property instead of the ArrayOfGuestNicInfo guest.net property. I thought about just adding guest.ipAddress to the list of properties that WaitForGuestNet monitors with a new case statement in the switch, but then it becomes a race between which one gets populated first in newer VMTools versions I believe. This seemed like a potential backwards compatibility problem, so I chose to make them completely independent waiters to be safe.

PR #716, opened 2019-02-25
Add Data Source and Resource for Role and Entity Permission [WIP]

bill-rich:
Copy of #657 along with additional tests and a few cleanup items.


bodymindarts:
Hi,

we're trying to use this branch. Currently having issues using the `vsphere_entity_permission` data source

```
data "vsphere_entity_permission" "cluster_permissions" {
  principal = "VSPHERE.LOCAL\\Administrator"
}
```
We get the error message:
```
Error: Error refreshing state: 1 error(s) occurred:

* data.vsphere_entity_permission.cluster_permissions: 1 error(s) occurred:

* data.vsphere_entity_permission.cluster_permissions: data.vsphere_entity_permission.cluster_permissions: role ID does not contain principal, entity type, and entity ID
```

From looking at the debug output I can see that the `Id()` function called [here](https://github.com/terraform-providers/terraform-provider-vsphere/blob/827caf13703c1ab75acefdff955ef3c3740f09c6/vsphere/data_source_vsphere_entity_permission.go#L42-L44):
```
func dataSourceVSphereEntityPermissionRead(d *schema.ResourceData, meta interface{}) error {
	log.Printf("[DEBUG] Reading entity permission %q", d.Id())
	client := meta.(*VSphereClient).vimClient
```
contains the empty string `""`.

We would love to use this branch, any ideas what might be the issue? @bill-rich @juliosueiras

juliosueiras:
@bodymindarts Hi, as of now, the entity permission data source is not finished refactoring yet by @bill-rich 

since from the look of it, the approach is changed from normalize folder entity permission (/, or /vm/*) to ManagedObject 

pseudo-example:

```terraform
data "vsphere_entity_permission" "cluster_permissions" {
   filter = "datacenter-2:Datacenter:VSPHERE.LOCAL\\Administrator"
}
```

so the pattern is <The entity ID that you want to find>:<The entity type, this can be found using govc or similar>:<The principal, so the user/role>

this change of approach is apply to all the new data source/resource of this pull request

if you would really want to use it right now due to situation, then for now you can use the original [pull request](https://github.com/juliosueiras/terraform-provider-vsphere/tree/f-role-permissions)

bill-rich:
I'm going to close the PR for now to do some additional work.

PR #731, opened 2019-03-25
vprops.ResourcePool is null,will panic for resourceVSphereVirtualMachineRead

xiaopengzhang:
when vprops.ResourcePool is null ,this math vappcontainer.IsVApp(client, vprops.ResourcePool.Value) will panic for vprops.ResourcePool.Value

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine.go
+++ b/vsphere/resource_vsphere_virtual_machine.go
@@ -365,16 +365,17 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 	// Resource pool
 	if vprops.ResourcePool != nil {
 		d.Set("resource_pool_id", vprops.ResourcePool.Value)
-	}
-	// If the VM is part of a vApp, InventoryPath will point to a host path
-	// rather than a VM path, so this step must be skipped.
-	if !vappcontainer.IsVApp(client, vprops.ResourcePool.Value) {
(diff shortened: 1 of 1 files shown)


bill-rich:
Great catch. The only issue with the proposed change is that `folder` will only be set if `ResourcePool` is not `nil`. What do you think of either adding a check that `vprops.ResourcePool` is not `nil` before attempting to use `vprops.ResourcePool.Value`, or rearranging the conditional so that `folder` can still be set?

xiaopengzhang:
I think the most appropriate modification is add check for vprops.ResourcePool  before attempting to use vprops.ResourcePool.Value.

bill-rich:
I was going back and thinking through the different cases on this and I realized vprops.ResourcePool wont be nil on [line 370](https://github.com/terraform-providers/terraform-provider-vsphere/blob/b09ed245cf3b203920ffd63a328f42d716f77a12/vsphere/resource_vsphere_virtual_machine.go#L370) since the check on  [line 366](https://github.com/terraform-providers/terraform-provider-vsphere/blob/b09ed245cf3b203920ffd63a328f42d716f77a12/vsphere/resource_vsphere_virtual_machine.go#L366) checks for that first. 

Did you come across a use case that was triggering a panic there?

bill-rich:
I'm going to close this one out for now as I am not sure on the use case that leads to the issue being addressed. Let me know if I'm missing something and I'd be happy to reopen and take another look. Thanks!

PR #771, opened 2019-05-23
Use terraform-v0.12.0

bill-rich:


rismoney:
is this a pre-req for using .12?

bill-rich:
@rismoney It isn't. There is already a provider release that uses Terraform 0.12-rc1, so the support is already there. The new provider release using v0.12.1 should be out today or tomorrow though.

PR #802, opened 2019-06-27
Mark windows_sysprep_text as sensitive

kvonbredow:
In a sysprep file it is fairly common to include passwords as part of the setup.  Our particular use case is during the domain joining step, though specifying an administrator password is another common case.  It seems that Microsoft's [official guidance](https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/wsim/hide-sensitive-data-in-an-answer-file) on "hiding" this sensitive data is simply base64 encoding it which obviously is not truly protecting this information.  The PR simply marks the field as sensitive so that any passwords in the file will not be revealed during the plan phase.


kvonbredow:
Hey @bill-rich, just curious if there were any concerns in getting this merged before the next release

PR #821, opened 2019-08-14
Updated properties section of OVA to include an equals sign

rayterrill:
Properties section appears incorrect in Terraform v0.12.


koikonom:
Thanks for taking the time to report this!

Issue #810, opened 2019-07-16
Getting error while creating VM under vAPP resource pool.


balubalananda:
Terraform v0.11.10

bill-rich:
Hi @balubalananda! This is due to the VM's path changing from being under the `vm` tree to being under the `host` tree when it is part of a vApp. There is a condition to handle this when the VM's immediate resource pool is a vApp, but if there is an intermediary resource pool, then the check fails to account for this. I will look into getting a bug fix in for this.

aareet:
@balubalananda this issue should be resolved with the next release.

balubalananda:
is there anyway, we can this latest fix in our environment ?

aareet:
@balubalananda you can create a build from the `master` branch by following the README if you wish

balubalananda:
> Terraform v0.11.10
> provider.vsphere: version = "~> 1.12"
> 
> We are getting below error while creating VM under vAPP resource pool.
> 
> • vsphere_virtual_machine.soa: 1 error(s) occurred:
> 
> * vsphere_virtual_machine.soa: error parsing virtual machine path "/Axxxx/host/HA-Hxxx/Resources/JCATVAPP/balutest": could not split path "/Axxx/host/HA-HPxx/Resources/JCATVAPP/balutest" on "/vm"
> 
> The below code is working fine with out any issue when we are using resource pool name ,which is not under Vsphere vAPP
> 
> #####working resource pool name format###
> 
> data "vsphere_resource_pool" "pool" {
> name = "HA-HPxxxx/Resources/HSS_Chef_Test_7.2"
> 
> #name format =<"cluster name>/Resource/
> 
> ###End working resource pool ####
> 
> ###Not working resource pool name format (under vAPP)####
> 
> data "vsphere_resource_pool" "pool" {
> name = "HA-Hxxxxxxxx/Resources/JCATVAPP/testvappresourcepool"
> $nameformat = Resources/<vAPPname/<resourcepoolname"
> 
> #######Not working resource pool code end##########
> 
> ##########Code: ###################
> 
> ```
> provider "vsphere" { 
> ```
> 
> user = "xxxx"
> password = "xxxxxx"
> vsphere_server = "10.3.xxx"
> 
> # if you have a self-signed cert
> allow_unverified_ssl = true
> }
> data "vsphere_datacenter" "dc" {
> name = "Axxx"
> }
> data "vsphere_datastore" "datastore" {
> name = "Txxxx"
[...]

PR #825, opened 2019-08-19
Change the way we detect if a VM is in a vApp

koikonom:
The previous way of detecting if a VM belogs to a resource pool that is
inside a vApp would fail if there was an intermediate rp between the
vApp itself and the VM.

We're now checking the `ParentVApp` property of the Virtual Machine
instead of just looking up the Resource Pool it belongs to.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine_test.go
+++ b/vsphere/resource_vsphere_virtual_machine_test.go
@@ -13610,19 +13610,19 @@ resource "vsphere_virtual_machine" "vm" {
 
 func testAccResourceVSphereVirtualMachineReadVappChildResourcePool() string {
 	return fmt.Sprintf(`
-variable "dc" {default="%s"}
-variable "datastore" {default="%s"}
-variable "cluster" {default="%s"}
-variable "network" {default="%s"}
-variable "template" {default="%s"}
-variable "vm_name" {default="%s"}
-variable "disk_size" {default="%s"}
-variable "hostname" {default="%s"}
-variable "domain" {default="%s"}
-variable "ipv4address" {default="%s"}
-variable "ipv4cidrbits" {default="%s"}
-variable "ipv4gateway" {default="%s"}
-
+variable "dc" { default = "%s" }
+variable "datastore" { default = "%s" }
+variable "cluster" { default = "%s" }
+variable "network" { default = "%s" }
+variable "template" { default = "%s" }
+variable "vm_name" { default = "%s" }
+variable "disk_size" { default = "%s" }
+variable "hostname" { default = "%s" }
+variable "domain" { default = "%s" }
+variable "ipv4address" { default = "%s" }
+variable "ipv4cidrbits" { default = "%s" }
+variable "ipv4gateway" { default = "%s" }
+variable "vapp_resource_pool" { default = "%s" }
 data "vsphere_datacenter" "dc" {
   name = "${var.dc}"
 }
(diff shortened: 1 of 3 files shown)


vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -13591,3 +13607,102 @@ resource "vsphere_virtual_machine" "vm" {
...
+    adapter_type = "${data.vsphere_virtual_machine.template.network_interface_types[0]}"
+  }
+
+  disk {
+    label            = "disk0"
+    size             = "${var.disk_size}"
+  }
+
+  clone {
+    template_uuid = "${data.vsphere_virtual_machine.template.id}"
```
bill-rich: I'd recommend using `linked_clone` here to speed up the test.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -13591,3 +13607,102 @@ resource "vsphere_virtual_machine" "vm" {
 		os.Getenv("VSPHERE_DATASTORE"),
 	)
 }
+
+func testAccResourceVSphereVirtualMachineReadVappChildResourcePool() string {
```
bill-rich: Run the TF config through `terraform fmt` to get the formatting fixed.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -13591,3 +13607,102 @@ resource "vsphere_virtual_machine" "vm" {
...
+}
+
+data "vsphere_resource_pool" "pool" { 
+	name = "hashi-cluster1/Resources/hashi-vapp/ko-rp"
+	datacenter_id = "${data.vsphere_datacenter.dc.id}" 
+  }
+    
+resource "vsphere_virtual_machine" "vm" {
+  name             = "${var.vm_name}"
+  #resource_pool_id = "${data.vsphere_compute_cluster.cluster.resource_pool_id}"
```
bill-rich: Remove the commented line

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -13591,3 +13607,102 @@ resource "vsphere_virtual_machine" "vm" {
...
+  datacenter_id = "${data.vsphere_datacenter.dc.id}"
+}
+
+data "vsphere_virtual_machine" "template" {
+  name          = "${var.template}"
+  datacenter_id = "${data.vsphere_datacenter.dc.id}"
+}
+
+data "vsphere_resource_pool" "pool" { 
+	name = "hashi-cluster1/Resources/hashi-vapp/ko-rp"
```
bill-rich: Either make this a variable so and add it to the .devrc, or you could also have a resource_pool resource to create it on demand.

vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -2957,6 +2957,22 @@ func TestAccResourceVSphereVirtualMachine_importClone(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccResourceVSphereVirtualMachinePreCheck(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceVSphereVirtualMachineReadVappChildResourcePool(),
+			},
```
bill-rich: It probably isn't really necessary, but it might be good to throw in a testAccResourceVSphereVirtualMachineCheckExists check here just to make sure the VM gets created as expected.

aareet:
Closes #810

PR #836, opened 2019-09-05
Add vSphere host resource

koikonom:
This is a first draft of a resource that allows terraform users to add new ESXi hosts to a vsphere cluster and manage things like:

[x] connection status
[x] maintenance status
[x] lockdown mode

Once we settle on what it should look like we can work on adding more features.

### DIFF
--- a/vsphere/internal/helper/hostsystem/host_system_helper.go
+++ b/vsphere/internal/helper/hostsystem/host_system_helper.go
@@ -51,7 +51,7 @@ func FromID(client *govmomi.Client, id string) (*object.HostSystem, error) {
 	defer cancel()
 	hs, err := finder.ObjectReference(ctx, ref)
 	if err != nil {
-		return nil, fmt.Errorf("could not find host system with id: %s: %s", id, err)
+		return nil, err
 	}
 	log.Printf("[DEBUG] Host system found: %s", hs.Reference().Value)
 	return hs.(*object.HostSystem), nil
@@ -98,6 +98,18 @@ func NameOrID(client *govmomi.Client, id string) string {
 	return name
 }
 
+// HostInMaintenance checks a HostSystem's maintenance mode and returns true if the
+// the host is in maintenance mode.
+func HostInMaintenance(host *object.HostSystem) (bool, error) {
+	hostObject, err := Properties(host)
+	if err != nil {
+		return false, err
+	}
+
+	return hostObject.Runtime.InMaintenanceMode, nil
+
+}
+
 // EnterMaintenanceMode puts a host into maintenance mode. If evacuate is set
 // to true, all powered off VMs will be removed from the host, or the task will
 // block until this is the case, depending on whether or not DRS is on or off
@@ -105,6 +117,12 @@ func NameOrID(client *govmomi.Client, id string) string {
 func EnterMaintenanceMode(host *object.HostSystem, timeout int, evacuate bool) error {
 	if err := viapi.VimValidateVirtualCenter(host.Client()); err != nil {
 		evacuate = false
+	}
+
+	maintMode, err := HostInMaintenance(host)
+	if maintMode {
+		log.Printf("[DEBUG] Host %q is already in maintenance mode", host.Name())
+		return nil
 	}
 
 	log.Printf("[DEBUG] Host %q is entering maintenance mode (evacuate: %t)", host.Name(), evacuate)
@@ -116,11 +134,31 @@ func EnterMaintenanceMode(host *object.HostSystem, timeout int, evacuate bool) e
 		return err
 	}
 
-	return task.Wait(ctx)
+	err = task.Wait(ctx)
+	if err != nil {
+		return err
+	}
+	var to mo.Task
+	err = task.Properties(context.TODO(), task.Reference(), nil, &to)
+	if err != nil {
+		log.Printf("[DEBUG] Failed while getting task results: %s", err)
+		return err
+	}
+
+	if to.Info.State != "success" {
+		return fmt.Errorf("Error while putting host(%s) in maintenance mode: %s", host.Reference(), to.Info.Error)
+	}
+	return nil
 }
 
 // ExitMaintenanceMode takes a host out of maintenance mode.
 func ExitMaintenanceMode(host *object.HostSystem, timeout int) error {
+	maintMode, err := HostInMaintenance(host)
+	if !maintMode {
+		log.Printf("[DEBUG] Host %q is already not in maintenance mode", host.Name())
+		return nil
+	}
+
 	log.Printf("[DEBUG] Host %q is exiting maintenance mode", host.Name())
 
 	// Add 5 minutes to timeout for the context timeout to allow for any issues
@@ -134,5 +172,29 @@ func ExitMaintenanceMode(host *object.HostSystem, timeout int) error {
 		return err
 	}
 
-	return task.Wait(ctx)
+	err = task.Wait(ctx)
+	if err != nil {
+		return err
+	}
+	var to mo.Task
+	err = task.Properties(context.TODO(), task.Reference(), nil, &to)
+	if err != nil {
+		log.Printf("[DEBUG] Failed while getting task results: %s", err)
+		return err
+	}
+
+	if to.Info.State != "success" {
+		return fmt.Errorf("Error while getting host(%s) out of maintenance mode: %s", host.Reference(), to.Info.Error)
+	}
+	return nil
+}
+
+// GetConnectionState returns the host's connection state (see vim.HostSystem.ConnectionState)
+func GetConnectionState(host *object.HostSystem) (types.HostSystemConnectionState, error) {
+	hostProps, err := Properties(host)
+	if err != nil {
+		return "", err
+	}
+
+	return hostProps.Runtime.ConnectionState, nil
 }
--- a/vsphere/resource_vsphere_host.go
+++ b/vsphere/resource_vsphere_host.go
@@ -1,0 +1,700 @@
+package vsphere
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/hostsystem"
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/viapi"
+	"github.com/vmware/govmomi/license"
+	"github.com/vmware/govmomi/property"
+
+	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/clustercomputeresource"
+
+	"github.com/vmware/govmomi/object"
+	gtask "github.com/vmware/govmomi/task"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceVsphereHost() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceVsphereHostCreate,
+		Read:   resourceVsphereHostRead,
+		Update: resourceVsphereHostUpdate,
+		Delete: resourceVsphereHostDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"datacenter": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "ID of the vSphere datacenter the host will belong to.",
+			},
+			"cluster": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "ID of the vSphere cluster the host will belong to.",
+			},
+			"hostname": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "FQDN or IP address of the host.",
+			},
+			"username": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Username of the administration account of the host.",
+			},
+			"password": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Password of the administration account of the host.",
+				Sensitive:   true,
+			},
+			"thumbprint": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Host's certificate SHA-1 thumbprint. If not set then the CA that signed the host's certificate must be trusted.",
+			},
+			"license": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "License key that will be applied to this host.",
+			},
+			"force": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Force add the host to vsphere, even if it's already managed by a different vSphere instance.",
+				Default:     false,
+			},
+			"connected": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Set the state of the host. If set to false then the host will be asked to disconnect.",
+				Default:     true,
+			},
+			"maintenance": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Set the host's maintenance mode. Default is false",
+				Default:     false,
+			},
+			"lockdown": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Set the host's lockdown status. Default is disabled. Valid options are 'disabled', 'normal', 'strict'",
+				Default:     "disabled",
+			},
+		},
+	}
+}
+
+func resourceVsphereHostRead(d *schema.ResourceData, meta interface{}) error {
+
+	// NOTE: Destroying the host without telling vsphere about it will result in us not
+	// knowing that the host does not exist any more.
+
+	// Look for host
+	client := meta.(*VSphereClient).vimClient
+	hostID := d.Id()
+
+	// Find host and get reference to it.
+	hs, err := hostsystem.FromID(client, hostID)
+	if err != nil {
+		if viapi.IsManagedObjectNotFoundError(err) {
+			d.SetId("")
+			return nil
+		}
+		log.Printf("[DEBUG] Error while searching host %s. Error: %s", hostID, err)
+		return err
+	}
+
+	maintenanceState, err := hostsystem.HostInMaintenance(hs)
+	if err != nil {
+		return err
+	}
+	d.Set("maintenance", maintenanceState)
+
+	// Retrieve host's properties.
+	log.Printf("[DEBUG] Got host %s", hs.String())
+	host, err := hostsystem.Properties(hs)
+	if err != nil {
+		log.Printf("[DEBUG] Error while retrieving properties for host %s. Error: %s", hostID, err)
+		return err
+	}
+
+	if host.Parent != nil && host.Parent.Type == "ClusterComputeResource" {
+		d.Set("cluster", host.Parent.Value)
+	} else {
+		d.Set("cluster", "")
+	}
+
+	connectionState, err := hostsystem.GetConnectionState(hs)
+	if err != nil {
+		return err
+	}
+
+	if connectionState == types.HostSystemConnectionStateDisconnected {
+		// Config and LicenseManager cannot be used while the host is
+		// disconnected.
+		d.Set("connected", false)
+		return nil
+	}
+	d.Set("connected", true)
+
+	lockdownMode, err := hostLockdownString(host.Config.LockdownMode)
+	if err != nil {
+		return nil
+	}
+
+	log.Printf("Setting lockdown to %s", lockdownMode)
+	d.Set("lockdown", lockdownMode)
+
+	lm := license.NewManager(client.Client)
+	am, err := lm.AssignmentManager(context.TODO())
+	if err != nil {
+		return err
+	}
+	licenses, err := am.QueryAssigned(context.TODO(), hostID)
+	if err != nil {
+		return err
+	}
+
+	licenseKey := d.Get("license").(string)
+	licFound := false
+	for _, lic := range licenses {
+		if licenseKey == lic.AssignedLicense.LicenseKey {
+			licFound = true
+			break
+		}
+	}
+
+	if !licFound {
+		d.Set("license", "")
+	}
+
+	return nil
+}
+
+func resourceVsphereHostCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*VSphereClient).vimClient
+
+	hcs := buildHostConnectSpec(d)
+
+	licenseKey := d.Get("license").(string)
+
+	lm := license.NewManager(client.Client)
+	ll, err := lm.List(context.TODO())
+	if err != nil {
+		return err
+	}
+
+	licFound := false
+	for _, l := range ll {
+		if l.LicenseKey == licenseKey {
+			licFound = true
+			break
+		}
+	}
+	if !licFound {
+		return fmt.Errorf("license key supplied (%s) did not match against known license keys", licenseKey)
+	}
+
+	var connectedState bool
+	val := d.Get("connected")
+	if val == nil {
+		connectedState = true
+	} else {
+		connectedState = val.(bool)
+	}
+
+	var task *object.Task
+	clusterID := d.Get("cluster").(string)
+	if clusterID != "" {
+		ccr, err := clustercomputeresource.FromID(client, clusterID)
+		if err != nil {
+			log.Printf("[DEBUG] Error while searching cluster %s. Error: %s", clusterID, err)
+			return err
+		}
+
+		task, err = ccr.AddHost(context.TODO(), hcs, connectedState, &licenseKey, nil)
+		if err != nil {
+			log.Printf("[DEBUG] Error while adding host with hostname %s to cluster %s.  Error: %s", d.Get("hostname").(string), clusterID, err)
+		}
+	} else {
+		dcId := d.Get("datacenter").(string)
+		dc, err := datacenterFromID(client, dcId)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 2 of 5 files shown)


vsphere/internal/helper/hostsystem/host_system_helper.go
```diff
@@ -51,6 +53,13 @@ func FromID(client *govmomi.Client, id string) (*object.HostSystem, error) {
 	defer cancel()
 	hs, err := finder.ObjectReference(ctx, ref)
 	if err != nil {
+		if soap.IsSoapFault(err) {
```
bill-rich: Check out [this](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/internal/helper/viapi/vim_helper.go#L54) function. Should make this check a little easier.

vsphere/internal/helper/hostsystem/host_system_helper.go
```diff
@@ -98,6 +107,19 @@ func NameOrID(client *govmomi.Client, id string) string {
 	return name
 }
 
+// HostInMaintenance checks a HostSystem's maintenance mode and returns true if the
+// the host is in maintenance mode.
+func HostInMaintenance(host *object.HostSystem) (bool, error) {
+	var hostObject mo.HostSystem
+	err := host.Properties(context.Background(), host.Reference(), nil, &hostObject)
```
bill-rich: With [this](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/internal/helper/hostsystem/host_system_helper.go#L62), you don't need to worry about supplying the context and the provider default timeout is used automatically.

vsphere/internal/helper/hostsystem/host_system_helper.go
```diff
@@ -116,11 +144,31 @@ func EnterMaintenanceMode(host *object.HostSystem, timeout int, evacuate bool) e
...
+		return err
+	}
+	var to mo.Task
+	err = task.Properties(context.TODO(), task.Reference(), nil, &to)
+	if err != nil {
+		log.Printf("[DEBUG] Failed while getting task results: %s", err)
+		return err
+	}
+
+	if to.Info.State != "success" {
```
bill-rich: Is there a non-success task state that can occur without an error?

vsphere/internal/helper/hostsystem/host_system_helper.go
```diff
@@ -134,5 +182,30 @@ func ExitMaintenanceMode(host *object.HostSystem, timeout int) error {
...
+		return fmt.Errorf("Error while getting host(%s) out of maintenance mode: %s", host.Reference, to.Info.Error)
+	}
+	return nil
+}
+
+// GetConnectionState returns the host's connection state (see vim.HostSystem.ConnectionState)
+func GetConnectionState(host *object.HostSystem) (types.HostSystemConnectionState, error) {
+	var hostProps mo.HostSystem
+	err := host.Properties(context.TODO(), host.Reference(), nil, &hostProps)
```
bill-rich: Same as the properties comment above.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+
+func resourceVsphereHostRead(d *schema.ResourceData, meta interface{}) error {
+
+	// NOTE: Destroying the host without telling vsphere about it will result in us not
+	// knowing that the host does not exist any more.
+
+	// Look for host
+	client := meta.(*VSphereClient).vimClient
+	hostID := d.Id()
+	if hostID == "" {
```
bill-rich: It should be safe to skip checking this. If the ID is empty, then the resource shouldn't exist.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+		return nil
+	}
+
+	// Find host and get reference to it.
+	hs, err := hostsystem.FromID(client, hostID)
+	if err != nil {
+		if soap.IsSoapFault(err) {
```
bill-rich: Same as ManagedObjectNotFound comment above.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	}
+
+	maintenanceState, err := hostsystem.HostInMaintenance(hs)
+	if err != nil {
+		return err
+	}
+	d.Set("maintenance", maintenanceState)
+
+	// Retrieve host's properties.
+	log.Printf("[DEBUG] Got host %#v", hs)
```
bill-rich: This can be ok, but can also be overly verbose to have the entire object. Might be better to just use the ID or name.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	return &schema.Resource{
+		Create: resourceVsphereHostCreate,
+		Read:   resourceVsphereHostRead,
+		Update: resourceVsphereHostUpdate,
+		Delete: resourceVsphereHostDelete,
+		// Importer: ,
+		Schema: map[string]*schema.Schema{
+			"cluster": {
+				Type:        schema.TypeString,
+				Required:    true,
```
bill-rich: Does this need to be required?

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	am, err := lm.AssignmentManager(context.TODO())
+	if err != nil {
+		return err
+	}
+	licenses, err := am.QueryAssigned(context.TODO(), hostID)
+	if err != nil {
+		return err
+	}
+
+	licenseKey := d.Get("license").(string)
```
bill-rich: If no license is currently known, then the `Read` function won't set it in the state here. Is that intentional?

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+
+	lm := license.NewManager(client.Client)
+	ll, err := lm.List(context.TODO())
+	if err != nil {
+		return err
+	}
+
+	licFound := false
```
bill-rich: This could be broken out into a helper function since it is used several times.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	if !licFound {
+		return fmt.Errorf("license key supplied (%s) did not match against known license keys", licenseKey)
+	}
+
+	var connectedState bool
+	val := d.Get("connected")
+	if val == nil {
+		connectedState = true
+	} else {
+
```
bill-rich: Extra blank line

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	}
+
+	task, err := ccr.AddHost(context.TODO(), hcs, connectedState, &licenseKey, nil)
+	if err != nil {
+		log.Printf("[DEBUG] Error while adding host with hostname %s to cluster %s.  Error: %s", d.Get("hostname").(string), clusterID, err)
+	}
+	task.Wait(context.TODO())
+
+	var to mo.Task
+	err = task.Properties(context.TODO(), task.Reference(), nil, &to)
```
bill-rich: Same question as above on non-success state that doesn't result in an error.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+
+	lockdownModeString := d.Get("lockdown").(string)
+	lockdownMode, err := hostLockdownType(lockdownModeString)
+	if err != nil {
+		return err
+	}
+
+	if connectedState {
+		var hostProps mo.HostSystem
+		err = host.Properties(context.TODO(), host.ConfigManager().Reference(), []string{"configManager.hostAccessManager"}, &hostProps)
```
bill-rich: Same as properties comment above.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+		hamRef := hostProps.ConfigManager.HostAccessManager.Reference()
+		ham := NewHostAccessManager(client.Client, hamRef)
+		err = ham.ChangeLockdownMode(context.TODO(), lockdownMode)
+		if err != nil {
+			return err
+		}
+	}
+
+	maintenanceMode := d.Get("maintenance").(bool)
+	var maintErr error
```
bill-rich: You can just use `err` here.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	task, err := hs.Destroy(context.TODO())
+	if err != nil {
+		return err
+	}
+	err = task.Wait(context.TODO())
+	if err != nil {
+		return fmt.Errorf("Error while waiting for host (%s) to be removed: %s", hostID, err)
+	}
+
+	var to mo.Task
```
bill-rich: Same as above about states and errors.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	client := meta.(*VSphereClient).vimClient
+	hostID := d.Id()
+
+	host, err := hostsystem.FromID(client, hostID)
+	if err != nil {
+		return err
+	}
+
+	maintenanceMode := new.(bool)
+	var maintErr error
```
bill-rich: Should be ok just using `err`

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	ham := NewHostAccessManager(client.Client, hamRef)
+	err = ham.ChangeLockdownMode(context.TODO(), lockdownMode)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func modifyMaintenanceMode(d *schema.ResourceData, meta, old, new interface{}) error {
```
bill-rich: Can we use the existing [EnterMaintenanceMode](https://github.com/terraform-providers/terraform-provider-vsphere/blob/master/vsphere/internal/helper/hostsystem/host_system_helper.go#L105) and [ExitMaintenanceMode](ExitMaintenanceMode) functions instead?

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	client := meta.(*VSphereClient).vimClient
+	hostID := d.Id()
+
+	host, err := hostsystem.FromID(client, hostID)
+	if err != nil {
+		return err
+	}
+
+	maintenanceMode := new.(bool)
```
bill-rich: Better to use something other than `new` for the variable name since its a built in function in Go.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+}
+
+func modifyLicense(d *schema.ResourceData, meta, old, new interface{}) error {
+	client := meta.(*VSphereClient).vimClient
+	lm := license.NewManager(client.Client)
+	lam, err := lm.AssignmentManager(context.TODO())
+	if err != nil {
+		return err
+	}
+	lam.Update(context.TODO(), d.Id(), new.(string), "")
```
bill-rich: Same on `new` name.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+		return err
+	}
+
+	if to.Info.State != "success" {
+		return fmt.Errorf("Error while removing host(%s): %s", hostID, to.Info.Error)
+	}
+	return nil
+}
+
+func modifyLockdownMode(d *schema.ResourceData, meta, old, new interface{}) error {
```
bill-rich: This is more a matter of just keeping a consistent style, but I'd recommend something along the lines of `resourceVSphereHostUpdateLockdownMode`. Its kind of long, but extra descriptive.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	} else {
+		maintErr = hostsystem.ExitMaintenanceMode(host, int(defaultAPITimeout))
+	}
+	if maintErr != nil {
+		return maintErr
+	}
+	return nil
+}
+
+func modifyLicense(d *schema.ResourceData, meta, old, new interface{}) error {
```
bill-rich: Same as above on function names.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	lm := license.NewManager(client.Client)
+	lam, err := lm.AssignmentManager(context.TODO())
+	if err != nil {
+		return err
+	}
+	lam.Update(context.TODO(), d.Id(), new.(string), "")
+	return nil
+}
+
+func modifyCluster(d *schema.ResourceData, meta, old, new interface{}) error {
```
bill-rich: Same as above on function names.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = hostsystem.EnterMaintenanceMode(hs, int(defaultAPITimeout), true)
+	if err != nil {
+		return fmt.Errorf("error while putting host to maintenance mode: %s", err.Error())
+	}
+
+	task, err := newCluster.MoveInto(context.TODO(), hs)
```
bill-rich: I think it would be worth adding a helper function here since it involves the context and waiting on a task.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	}
+
+	if to.Info.State != "success" {
+		return fmt.Errorf("Error while moving host to new cluster (%s): %s", newClusterID, to.Info.Error)
+	}
+
+	return nil
+}
+
+func handleReconnect(d *schema.ResourceData, meta interface{}) error {
```
bill-rich: Same as above on function names.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+	hostID := d.Id()
+	client := meta.(*VSphereClient).vimClient
+	host := object.NewHostSystem(client.Client, types.ManagedObjectReference{Type: "HostSystem", Value: d.Id()})
+	hcs := types.HostConnectSpec{
```
bill-rich: Creating a function to return the spec could cut back on repeated code and make things look a little cleaner.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,683 @@
...
+		HostName:      d.Get("hostname").(string),
+		UserName:      d.Get("username").(string),
+		Password:      d.Get("password").(string),
+		SslThumbprint: d.Get("thumbprint").(string),
+		Force:         d.Get("force").(bool),
+	}
+
+	task, err := host.Reconnect(context.TODO(), &hcs, nil)
```
bill-rich: Same as above for adding a function since you're dealing with a context and waiting on tasks.

vsphere/resource_vsphere_host_test.go
```diff
@@ -0,0 +1,438 @@
...
+					testAccVSphereHostExists("vsphere_host.h1"),
+				),
+			},
+		},
+	})
+
+}
+
+func TestAccResourceVSphereHost_connection(t *testing.T) {
+
```
bill-rich: Extra blank line

vsphere/resource_vsphere_host_test.go
```diff
@@ -0,0 +1,438 @@
...
+					testAccVSphereHostConnected("vsphere_host.h1", true),
+				),
+			},
+		},
+	})
+
+}
+
+func TestAccResourceVSphereHost_maintenance(t *testing.T) {
+
```
bill-rich: Extra blank line

vsphere/resource_vsphere_host_test.go
```diff
@@ -0,0 +1,438 @@
...
+					testAccVSphereHostMaintenanceState("vsphere_host.h1", false),
+				),
+			},
+		},
+	})
+
+}
+
+func TestAccResourceVSphereHost_lockdown(t *testing.T) {
+
```
bill-rich: Extra blank line

vsphere/internal/helper/hostsystem/host_system_helper.go
koikonom: Yeah one of my concerns while writing this resource was that I'd probably end up re-implementing things that are already done elsewhere in the provider :).

Thanks for the pointer!

vsphere/internal/helper/hostsystem/host_system_helper.go
koikonom: 👍

vsphere/internal/helper/hostsystem/host_system_helper.go
koikonom: My understanding is that all vSphere API calls that return a task are asynchronous. If we get an error when we make the call it usually means that our parameters are wrong or that vsphere wasn't able to start doing the task. 

My expectation is that if something goes wrong between the time the API call returns a task ID and the time the task completes then it will be reflected in the task's Info structure.

(For example vsphere timed out while trying to talk to the host)

vsphere/internal/helper/hostsystem/host_system_helper.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: Good call. Probably a left over from my tests.

vsphere/resource_vsphere_host.go
koikonom: Yeah, the host is added to a cluster using ClusterComputeResource.AddHost()

Of course now that I'm typing these lines I remembered there's also Folder.AddStandaloneHost() so I guess I can always use that if a cluster is not specified.

vsphere/resource_vsphere_host.go
koikonom: I do set the license to an empty string in line 181 hoping that it will prompt the user to update the config and set a proper licence. Can you suggest an alternative solution ?

vsphere/resource_vsphere_host.go
koikonom: See above :)

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: ⬆️

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: 👍

vsphere/resource_vsphere_host.go
koikonom: Unfortunately it's not quite the same. Each time I'm iterating different structures, first one is list of licences assigned to a host and the second time is a list of licenses installed on the system.

I'm concerned that getting that into a function that would support both would make the code more complicated with little benefit, but I'm open to suggestions :).

vsphere/resource_vsphere_host.go
koikonom: That's what the function does, depending on what the config says.

vsphere/internal/helper/hostsystem/host_system_helper.go
bill-rich: You are correct on the vSphere API, but `govmomi` includes some helpers to wait for the task to complete and make sure it is successful. A few steps down the rabbit hole, and `task.Wait()` leads to [this](https://github.com/vmware/govmomi/blob/master/task/wait.go#L117) function which waits for the task to finish with either an "failure" or "success" state.

As long as the task is using `Wait`, you should be ok to trust that no error means it completed successfully.

vsphere/resource_vsphere_host.go
bill-rich: That makes sense! I just overlooked the difference. Makes sense to keep them separate.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,711 @@
...
+			return err
+		}
+	}
+
+	taskResult, err := processTask(task)
+	if err != nil {
+		return fmt.Errorf("Host addition failed. %s", err)
+	}
+
+	log.Printf("[DEBUG] Task Result: %#v", taskResult)
```
bill-rich: I think its generally better to avoid the fully verbose interpolation (especially with vSphere since it can get _quite_ verbose),  but there are certainly cases where it makes sense. Just want to make sure its purposeful here.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,711 @@
...
+	var hostID string
+	taskResultType := taskResult.(types.ManagedObjectReference).Type
+	switch taskResultType {
```
bill-rich: `ComputeResource` and `HostSystem` should be sufficient, but I'd recommend a `default` with an error about an unexpected type just in case.

vsphere/resource_vsphere_host.go
```diff
@@ -0,0 +1,711 @@
...
+		}
+	}
+
+	hostProps, err := hostsystem.Properties(hs)
+	if err != nil {
+		log.Printf("Error while retrieving properties fort host %s. Error: %s", hostID, err)
+		return err
+	}
+
+	// If this is a standalone host we need to destroy the ComputeResource object
```
bill-rich: What are the two options here? What does it mean to be a `ComputeResource` vs the alternative? This is a question for my understanding more than an question of the implementation or criticism.

vsphere/internal/helper/hostsystem/host_system_helper.go
```diff
@@ -51,7 +51,7 @@ func FromID(client *govmomi.Client, id string) (*object.HostSystem, error) {
 	defer cancel()
 	hs, err := finder.ObjectReference(ctx, ref)
 	if err != nil {
-		return nil, fmt.Errorf("could not find host system with id: %s: %s", id, err)
+		return nil, err
```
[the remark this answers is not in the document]
koikonom: Because the error messages coming from the API can be a bit vague sometimes. Adding a little bit of text gives better context and we can also track the source of the error just by grepping for the error message if we have to :) .

vsphere/resource_vsphere_host.go
koikonom: The parent for the hostsystem is either a ComputeResource (which means this is a standalone host we're deleting) or a ClusterComputeResouce (which means the host is a member of a cluster). 

Calling HostSystem.Destroy() on a "standalone" host (i.e a host that's not part of a vsphere cluster) then the task returns with a failure. In order to remove the HostSystem, you need to destroy the parent ComputeResource, and then vsphere will remove the HostSystem on its own.

vsphere/resource_vsphere_host.go
koikonom: For what it's worth I've ended up splitting both blocks on functions of their own, to keep the main CRUD functions small and easier (I hope) to read. 

I'll follow up with a second PR for that though :)

vsphere/resource_vsphere_host.go
koikonom: Oops :) Yeah these are not left like that purpose. I'll do a quick sweep to be sure I haven't left any others like that.

vsphere/internal/helper/hostsystem/host_system_helper.go
koikonom: I had no idea :) ! I will give that a look, thanks!

vsphere/internal/helper/hostsystem/host_system_helper.go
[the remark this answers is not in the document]
bill-rich: You went the other way here though, right? It looks like it originally included "could not find host system with..." and that part was removed.

vsphere/resource_vsphere_host.go
bill-rich: :+1:

vsphere/internal/helper/hostsystem/host_system_helper.go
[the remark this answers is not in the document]
koikonom: Hmm you're right.... I guess I read the diff the other way. This must have happened by accident. The good news is that I have another branch I'm working on where I'm revisiting all this code and adding some text instead of just returning the vsphere error, so this will be adressed there.

vsphere/resource_vsphere_host.go
koikonom: done

vsphere/resource_vsphere_host.go
koikonom: done

vsphere/internal/helper/hostsystem/host_system_helper.go
koikonom: done

koikonom:
(Squashed all commits into one)

PR #841, opened 2019-09-16
Extend ignored_guest_ips to support CIDR networks

isometry:
* Correctly parse CIDR networks included in the `ignored_guest_ips` list, allowing entire arbitrary IP ranges to be ignored.
* Compare guest IP addresses to `ignored_guest_ips` using `net.IP.Equal()` rather than raw string comparison, enhancing support for variations in IPv6 representation.

### DIFF
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/terraform-providers/terraform-provider-vsphere/vsphere/internal/helper/folder"
@@ -358,8 +359,19 @@ func skipIPAddrForWaiter(ip net.IP, ignoredGuestIPs []interface{}) bool {
 		return true
 	default:
 		for _, ignoredGuestIP := range ignoredGuestIPs {
-			if ip.String() == ignoredGuestIP.(string) {
-				return true
+			if strings.Contains(ignoredGuestIP.(string), "/") {
+				_, ignoredIPNet, err := net.ParseCIDR(ignoredGuestIP.(string))
+				if err != nil {
+					log.Printf("[ERROR] Failed to parse ignored_guest_ips entry: %q", ignoredGuestIP)
+					continue
+				} else if ignoredIPNet.Contains(ip) {
+					return true
+				}
+			} else {
+				ignoredIP := net.ParseIP(ignoredGuestIP.(string))
+				if ignoredIP != nil && ignoredIP.Equal(ip) {
+					return true
+				}
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


isometry:
Patch updated following testing. Is there anything else required from me before this can be merged?

We need the functionality in order to prevent the provider from binding to the temporary DHCP IP temporarily detected even when we assign a fixed IP via Guest Customization, for example if the attached subnet hands out dynamic addresses in `10.0.0.128/25` but we need to assign a static IP outside that range, `foo.default_ip_address` will be a DHCP address which is no longer valid after VMware Guest Customization has completed:

```terraform
resource "vsphere_virtual_machine" "foo" {
  …
  ignored_guest_ips = ["10.0.0.128/25"]
  clone {
    …
    customize {
      …
      network_interface {
        ipv4_address = 10.0.0.10
        ipv4_netmask = 24
      }
    }
  }
}
```

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -358,8 +359,19 @@ func skipIPAddrForWaiter(ip net.IP, ignoredGuestIPs []interface{}) bool {
 		return true
 	default:
 		for _, ignoredGuestIP := range ignoredGuestIPs {
-			if ip.String() == ignoredGuestIP.(string) {
-				return true
+			if strings.Contains(ignoredGuestIP.(string), "/") {
+				_, ignoredIPNet, err := net.ParseCIDR(ignoredGuestIP.(string))
+				if err != nil {
+					log.Printf("[ERROR] Failed to parse ignored_guest_ips entry: %q", ignoredGuestIP)
+					continue
```
koikonom: I definitely agree that checking if the string actually represents a CIDR is the right thing to do, but in this case just dropping it and moving to the next element in the list may cause trouble, since all that happens during runtime.

How about moving the checks into a validation function in the schema? What I was thinking is that you could replace resource_vsphere_virtual_machine.go:116 with something like this:

```
			Elem:        &schema.Schema{
				Type: schema.TypeString,
				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
					v := val.(string)
                                        // cidr and ip validatio logic goes here
					return
				},
			},

```

An example of a simple validation function can be found here:
https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#validatefunc

Since the parameters will be validated when terraform evaluates the config we can be confident that when skipIPAddrForWaiter is called the list will only contain either IPs or CIDRs.

isometry: Great idea. Implemented in 6d76080.

isometry:
@koikonom : I'm not sure why this PR is still marked "Changes Requested", but I implemented the changes you suggested two weeks ago. Any chance of getting this merged before v1.14.0?

koikonom:
Hi @isometry, I am sorry for the delay. We have been having some issues with our acceptance tests lately so I've spent all my time getting these in order first. Once we are confident that things are back in working order I will merge this PR (before v1.14.0).

Issue #809, opened 2019-07-15
Add support in `d/virtual_machine` to return network adapters and mac addresses


gvcgael:
Hello,

It would be great if the `data` `virtual_machine` could return the mac addresses of the network adapters of the VM : 

https://www.terraform.io/docs/providers/vsphere/d/virtual_machine.html

At the moment, we have to set the `mac addresses` hardcoded, but its not possible since of this bug  : https://github.com/terraform-providers/terraform-provider-vsphere/issues/488

aareet:
Relates to #488

azbpa:
Same requirement here. Any chance this gets implemented?

arizvisa:
It seems that the vsphere_virtual_machine data source needs another entry added to its schema. Either it's the following attribute in the schema.
https://github.com/hashicorp/terraform-provider-vsphere/blob/814675b5da80f286a082664d22455b7c8264134a/vsphere/data_source_vsphere_virtual_machine.go#L135-L140

Or the field could be added to the schema here:
https://github.com/hashicorp/terraform-provider-vsphere/blob/814675b5da80f286a082664d22455b7c8264134a/vsphere/virtual_machine_guest_structure.go#L15-L30

In the `buildAndSelectGuestIPs()` function (https://github.com/hashicorp/terraform-provider-vsphere/blob/814675b5da80f286a082664d22455b7c8264134a/vsphere/virtual_machine_guest_structure.go#L43
), the addresses are aggregated into the `addrs` list from both the `v4net2addrs` and `v6net2addrs` maps and then set to "guest_ip_addresses" at the bottom of the function.

https://github.com/hashicorp/terraform-provider-vsphere/blob/814675b5da80f286a082664d22455b7c8264134a/vsphere/virtual_machine_guest_structure.go#L123-L128

Both of those maps are keyed by `n.DeviceConfigId` which should be the hwaddr for each device, so it seems like all one would need to do would be to aggregate each of the keys into another list that would then be set directly into the schema similar to the "guest_ip_addresses" field.

As this issue was tagged with "help wanted", it might be faster for someone to contribute and submit a PR.

appilon:
This actually went out in v2.0.0 of the provider

arizvisa:
I think it was #1274?

PR #843, opened 2019-09-17
Enhancement to virtual machine data source

brian57860:
Adds new property 'network_interfaces' to vSphere Virtual Machine data source which returns information about each network interface on the virtual machine or template.

This is useful for discovering certain network interfaces while performing a linked clone, specifically where the desired state is to replicate the network settings of the source virtual machine on the destination.

The sub-attributes which are exported are as follows:

- network_id: The managed object reference ID of the network backing the interface.
- adapter_type: The network interface type, i.e.  e1000, e1000e, or vmxnet3. (deprecates requirement for separate 'network_interfaces_types' property)
- mac_address: The mac_address of the network interface

Properties can be referenced as follows:

"${data.vsphere_virtual_machine.source.network_interfaces.0.adapter_type}"

Resolves: #809

### DIFF
--- a/vsphere/data_source_vsphere_virtual_machine.go
+++ b/vsphere/data_source_vsphere_virtual_machine.go
@@ -77,6 +77,27 @@ func dataSourceVSphereVirtualMachine() *schema.Resource {
 					},
 				},
 			},
+			"network_interfaces": {
+				Type:        schema.TypeList,
+				Description: "The network interfaces found on the virtual machine, sorted by unit number.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"adapter_type": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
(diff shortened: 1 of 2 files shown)


koikonom:
Hi @brian57860, sorry for the long delay. The good news is that we're getting back on top of the PR queue, the bad news is that I need to ask you a couple of things. 

- Can you please rebase your change on top of master in order to resolve the merge conflicts?
- It would be nice if you could also write an acceptance test for this change, so we can tell if a future change causes this feature to break.

Thank you for taking the time to submit this PR!

tenthirtyam:
@appilon - Would you mind taking a take a look at this PR and determining if we can reopen and rebase on @brian57860's behalf so that the contribution is recognized?

**Ryan Johnson**
Staff II Solutions Architect | VMware, Inc.

appilon:
This actually went out in commits that landed in v2.0.0 of the provider

PR #848, opened 2019-09-25
    [vsphere_host] Minor fixes

koikonom:
- Create cluster while testing instead of expecting one to be there
   already.

- Moved the licence checks into their own functions. Also fixed a
   logic bug that occurs when users don't pass a license attribute,
   and added the relevant test.

- Add argument validation and tests also fix bug by adding missing
   return statement.

### DIFF
--- a/vsphere/resource_vsphere_host.go
+++ b/vsphere/resource_vsphere_host.go
@@ -91,6 +91,13 @@ func resourceVsphereHost() *schema.Resource {
 				Optional:    true,
 				Description: "Set the host's lockdown status. Default is disabled. Valid options are 'disabled', 'normal', 'strict'",
 				Default:     "disabled",
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					v := val.(string)
+					if v != "disabled" && v != "normal" && v != "strict" {
+						errs = append(errs, fmt.Errorf("%s must be one of 'disabled', 'normal', or 'strict'. Got: %s", key, v))
+					}
+					return
+				},
 			},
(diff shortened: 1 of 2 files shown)


vsphere/resource_vsphere_host.go
```diff
@@ -91,6 +91,13 @@ func resourceVsphereHost() *schema.Resource {
 				Optional:    true,
 				Description: "Set the host's lockdown status. Default is disabled. Valid options are 'disabled', 'normal', 'strict'",
 				Default:     "disabled",
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
```
bill-rich: I believe `validate.StringInSlice` will do the same for you here, and then you don't need to write a new function.

vsphere/resource_vsphere_host.go
```diff
@@ -157,55 +164,43 @@ func resourceVsphereHostRead(d *schema.ResourceData, meta interface{}) error {
...
+		if !licFound {
+			d.Set("license", "")
+		}
 	}
 
 	return nil
 }
 
 func resourceVsphereHostCreate(d *schema.ResourceData, meta interface{}) error {
+	err := validateFields(d)
```
bill-rich: This way of implementing the validation works. You can also use `CustomizeDiff` which will run before the CRUD functions. Thats how we've done it in some other resources, but either is fine. Mainly just wanted to point out the possibility.

koikonom: Thanks for the advice!

PR #850, opened 2019-09-26
[vsphere_host] Add error messages

koikonom:


vsphere/resource_vsphere_host.go
```diff
@@ -657,7 +652,7 @@ func hostLockdownString(lockdownMode types.HostLockdownMode) (string, error) {
 		log.Printf("Found match for %s. Returning %s.", lockdownMode, modeString)
 		return modeString, nil
 	}
-	return "", fmt.Errorf("Unknwown Lockdown mode encountered")
+	return "", fmt.Errorf("unknwown Lockdown mode encountered")
```
aareet: "Unknown"

vsphere/resource_vsphere_host.go
```diff
@@ -642,7 +637,7 @@ func hostLockdownType(lockdownMode string) (types.HostLockdownMode, error) {
...
```
aareet: "Unknown"

vsphere/resource_vsphere_host.go
```diff
@@ -114,22 +114,20 @@ func resourceVsphereHostRead(d *schema.ResourceData, meta interface{}) error {
 			d.SetId("")
 			return nil
 		}
-		log.Printf("[DEBUG] Error while searching host %s. Error: %s", hostID, err)
-		return err
+		return fmt.Errorf("error while searching host %s. Error: %s ", hostID, err)
```
aareet: Is it worth pulling the string "error while searching ...." into a local variable? Seems to be repeated a lot

aareet:
@koikonom bump on the comments :)

vsphere/resource_vsphere_host.go
koikonom: Seems like an overkill, I'd rather come up with original error messages :)

koikonom:
Thanks for the reminder! Take a look and if you're happy I can merge on Monday.

PR #876, opened 2019-10-21
Support for vNic resources

koikonom:
This resource manages `vmk` interfaces on ESXi hosts. Using a
combination of this resource and `vsphere_host` a terraform user can
setup a cluster of esx hosts and configure vmotion on them.

### DIFF
--- a/vsphere/resource_vsphere_vnic.go
+++ b/vsphere/resource_vsphere_vnic.go
@@ -42,9 +42,8 @@ func vNicSchema() map[string]*schema.Schema {
 func resourceVsphereNicRead(d *schema.ResourceData, meta interface{}) error {
 	ctx := context.TODO()
 	client := meta.(*VSphereClient).vimClient
-	tfNicID := d.Id()
 
-	toks := strings.Split(tfNicID, "_")
+	toks := strings.Split(d.Id(), "_")
 	hostId := toks[0]
 	nicId := toks[1]
 
@@ -62,7 +61,6 @@ func resourceVsphereNicRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	_ = d.Set("mtu", vnic.Spec.Mtu)
 	_ = d.Set("mac", vnic.Spec.Mac)
-	//_ = d.Set("netstack", vnic.Spec.NetStackInstanceKey)
 
 	// Do we have any ipv4 config ?
 	if _, ok := d.GetOk("ipv4"); ok {
@@ -76,7 +74,10 @@ func resourceVsphereNicRead(d *schema.ResourceData, meta interface{}) error {
 				ipv4dict["gw"] = vnic.Spec.IpRouteSpec.IpRouteConfig.GetHostIpRouteConfig().DefaultGateway
 			}
 		}
-		_ = d.Set("ipv4", []map[string]interface{}{ipv4dict})
+		err = d.Set("ipv4", []map[string]interface{}{ipv4dict})
+		if err != nil {
+			return err
+		}
 	}
 
 	// Do we have any ipv6 config ?
@@ -100,7 +101,10 @@ func resourceVsphereNicRead(d *schema.ResourceData, meta interface{}) error {
 			// There is a gw set in the config, but none set on the Host.
 			ipv6dict["gw"] = ""
 		}
-		d.Set("ipv6", []map[string]interface{}{ipv6dict})
+		err = d.Set("ipv6", []map[string]interface{}{ipv6dict})
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


vsphere/resource_vsphere_vnic.go
```diff
@@ -0,0 +1,517 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: vNicSchema(),
+	}
+}
+
+func vNicSchema() map[string]*schema.Schema {
+	base := BaseVMKernelSchema()
+	base["host"] = &schema.Schema{
```
bill-rich: Why is `host` separate from the rest of the schema?

vsphere/resource_vsphere_vnic.go
```diff
@@ -0,0 +1,517 @@
...
+
+	return base
+}
+
+func resourceVsphereNicRead(d *schema.ResourceData, meta interface{}) error {
+	ctx := context.TODO()
+	client := meta.(*VSphereClient).vimClient
+	tfNicID := d.Id()
+
+	toks := strings.Split(tfNicID, "_")
```
bill-rich: I like to just use d.Id() directly here. That way you aren't storing the ID in a separate var again that is only used once.

vsphere/resource_vsphere_vnic.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	_ = d.Set("portgroup", vnic.Portgroup)
+	if vnic.Spec.DistributedVirtualPort != nil {
+		_ = d.Set("distributed_switch_port", vnic.Spec.DistributedVirtualPort.SwitchUuid)
+		_ = d.Set("distributed_port_group", vnic.Spec.DistributedVirtualPort.PortgroupKey)
+	}
+	_ = d.Set("mtu", vnic.Spec.Mtu)
+	_ = d.Set("mac", vnic.Spec.Mac)
+	//_ = d.Set("netstack", vnic.Spec.NetStackInstanceKey)
```
bill-rich: If this isn't needed in the current implementation or as a future reminder, it should be removed to clean up the clutter.

vsphere/resource_vsphere_vnic.go
```diff
@@ -0,0 +1,517 @@
...
+		ipv4dict := make(map[string]interface{})
+		ipv4dict["dhcp"] = vnic.Spec.Ip.Dhcp
+		if !vnic.Spec.Ip.Dhcp {
+			ipv4dict["ip"] = vnic.Spec.Ip.IpAddress
+			ipv4dict["netmask"] = vnic.Spec.Ip.SubnetMask
+			if vnic.Spec.IpRouteSpec != nil {
+				ipv4dict["gw"] = vnic.Spec.IpRouteSpec.IpRouteConfig.GetHostIpRouteConfig().DefaultGateway
+			}
+		}
+		_ = d.Set("ipv4", []map[string]interface{}{ipv4dict})
```
bill-rich: The error from d.Set should be checked with non-scalar values since that is usually where a `Set` will fail.

vsphere/resource_vsphere_vnic.go
```diff
@@ -0,0 +1,517 @@
...
+		addrList := make([]string, 0)
+		for _, addr := range vnic.Spec.Ip.IpV6Config.IpV6Address {
+			if addr.Origin == "manual" {
+				addrList = append(addrList, fmt.Sprintf("%s/%d", addr.IpAddress, addr.PrefixLength))
+			}
+		}
+		ipv6dict["addresses"] = addrList
+		if vnic.Spec.IpRouteSpec != nil {
+			ipv6dict["gw"] = vnic.Spec.IpRouteSpec.IpRouteConfig.GetHostIpRouteConfig().IpV6DefaultGateway
+		} else if _, ok := d.GetOk("ipv6.0.gw"); ok {
```
bill-rich: Why is `gw` only set if it exists in the config? If it isn't in the config, it can still be set to "", right?

vsphere/resource_vsphere_vnic.go
koikonom: Because the rest of the schema can be reused if we decide to include vnic as a block within a vsphere_host resource.

vsphere/resource_vsphere_vnic.go
koikonom: If we don't set one in the config then there shouldn't be one set on the host either. Otherwise someone could set the gw via the vsphere UI and cause a configuration drift that we'd be unable to remediate.

bill-rich: Maybe I'm misreading, but this saying that if `ipv6.0.gw` is not set in the state or config and the `spec` is `nil`, then set `ipv6.0.gw` to `""`, right? Wouldn't we want it to set `ipv6.0.gw` to `""` any time vsphere says `IpRouteSpec` is empty?

koikonom:
(pushed a rebased branch, since github could not rebase on its own due to conflicts)

Issue #891, opened 2019-11-01
failed to delete a compute cluster with ha_admission_control_policy = "failoverHosts"


aeremenk0:
### Terraform Version
Terraform v0.12.12

### vSphere Provider Version
commit a4b63434233fd5893d0907580e1e75439c563dc5

### Affected Resource(s)
affected resource: `vsphere_compute_cluster`

### Terraform Configuration Files
go test -v -run=TestAccResourceVSphereComputeCluster_explicitFailoverHost -timeout 240m

### Debug Output
--- FAIL: TestAccResourceVSphereComputeCluster_explicitFailoverHost (28.59s)
    testing.go:630: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: errors during apply: error force-removing old hosts out of cluster: error moving host "host1" out of cluster "terraform-compute-cluster-test": The current operation cannot be performed on the host 'host1' because it has been configured as a failover host for a cluster.

### Pre-condition
in TestAccResourceVSphereComputeCluster_explicitFailoverHost file: resource_vsphere_compute_cluster_test.go:104 - 113
comments from the test:
// This step is required to remove the dedicated failover host in
// admission control - otherwise cleanup will fail.
this test case hides the problem if it is removed a compute cluster delete will fail.
That is why I think that code between lines 104 - 113 should be removed.

### Expected Behavior
When removing a compute cluster with Admission Policy set to FailoverHost spec.DasConfig.Enabled should be set to structure.BoolPtr(false) first. And then the cluster remove command should be issued.

### Actual Behavior
Error during the compute cluster delete:

Error: errors during apply: error force-removing old hosts out of cluster: error moving host "host1" out of cluster "terraform-compute-cluster-test": The current operation cannot be performed on the host 'host1' because it has been configured as a failover host for a cluster.

### Steps to Reproduce
comment out those lines of the code:
`resource_vsphere_compute_cluster_test.go:104 - 113`

`cd vsphere`
and run:

`go test -v -run=TestAccResourceVSphereComputeCluster_explicitFailoverHost -timeout 240m`

koikonom:
Hi @aeremenk0, thanks for filing this issue! I see you have already opened a PR against this issue so I'll close this and continue the discussion in #892.

PR #892, opened 2019-11-01
fix for compute cluster delete when Admission Policy is set to FailoverHost issue #891

aeremenk0:
ResourceReductionToToleratePercent should be set to ha_admission_control_performance_tolerance
compute cluster delete should set HA to OFF for Admission Control Failover host policy before the delete


aeremenk0:
could anybody explain me PR process with the project?

PR #897, opened 2019-11-04
Pick default datastore for extra disks

koikonom:
When cloning virtual machines it is possible to define additional disks
to be created. If a datastore is not defined either for the disk or for
the virtual machine then the creation of the additional virtual machines
will fail.

This change introduces a fallback where if a datastore is not set we
pick the first datastore found that is associated with the virtual
machine.


koikonom:
Fixes #883

vkmellon:
Hi! When it will be released?

PR #899, opened 2019-11-06
ha_admission_control_performance_tolerance failed to set to 0.

aeremenk0:
ha_admission_control_performance_tolerance check added to failover host configuration and now the test is failing.

the reason behind that is govmomi library where:
`ResourceReductionToToleratePercent int32 `xml:"resourceReductionToToleratePercent,omitempty"``

that means that resourceReductionToToleratePercent will be dropped from the SOAP payload if it is 0.

I opened the issue and pull request was created https://github.com/vmware/govmomi/pull/1687
the problem is that pull request changes `ResourceReductionToToleratePercent` declaration to:
 
`ResourceReductionToToleratePercent *int32 xml:"resourceReductionToToleratePercent,omitempty"`

that will affect the code of the provider.


aeremenk0:
Pull Request is merged to master: https://github.com/vmware/govmomi/pull/1687

commit SHA1: 399b1c2446910fd020a93a4dde1839f6d4056ab8

PR #819, opened 2019-08-07
Sort collected IPv4 and IPv6 addresses by network device config ID

sirlatrom:
Resolves #368 and helps ensure the order of guest IP addresses corresponds to the declared order of network interfaces. If each interface has exactly one IP address, it becomes possible to get the IP address for a specific interface, e.g. for `eth1` by using the expression `vsphere_virtual_machine.vm.guest_ip_addresses[1]`.

PR #902, opened 2019-11-18
Set guest IP addresses on data.vsphere_virtual_machine if available

### LINKED ISSUE #901: Guest IP addresses not available on data.vsphere_virtual_machine
### Terraform Version

```
Terraform v0.12.13
```

### vSphere Provider Version

```
.
└── provider.vsphere ~> 1.13
```

### Affected Resource(s)

- `data.vsphere_virtual_machine`

main.tf
```hcl
provider "vsphere" {
...
}
data "vsphere_datacenter" "dc" {
  name = var.datacenter
}
data "vsphere_virtual_machine" "consul_server" {
  for_each      = toset(var.vm_paths)
  name          = each.value
  datacenter_id = data.vsphere_datacenter.dc.id
}
```

outputs.tf
```hcl
output "vm_eth1_ip_addresses" {
    value = [for n in data.vsphere_virtual_machine.vm : n.guest_ip_addresses[1]]
}
```

### Debug Output

N/A

### Expected Behavior

A list of IP addresses.

### Actual Behavior

It is not implemented. Output is as follows:

```
Error: Unsupported attribute

  on outputs.tf line 2, in output "vm_eth1_ip_addresses":
   2:     value = [for n in data.vsphere_virtual_machine.vm : n.guest_ip_addresses[1]]

This object does not have an attribute named "guest_ip_addresses".
```

### Steps to Reproduce

1. Have an output as in the example config
2. Point to a list of existing VMs with a `data.vsphere_virtual_machine` block
3. `terraform refresh`

### Important Factoids

The VMs have been created with another Terraform config, and in the resource attributes of those I can see the guest IP addresses. It is just not implemented on the data source side.

### References

The information will be more useful if #819 were to be merged.

sirlatrom:
This PR adds a `guest_ip_addresses` attribute to the `data.vsphere_virtual_machine` datasource. It is intended to be used when a Terraform configuration needs to know the IP addresses of existing VMs in vSphere, e.g. a HashiCorp Vault cluster that needs its Consul agents to join an existing Consul server cluster.

In the case of VMs with multiple NICs, it becomes especially useful when combined with #819, which sorts the IP addresses by network device ID for each VM.

Fixes #901

### DIFF
--- a/vsphere/data_source_vsphere_virtual_machine.go
+++ b/vsphere/data_source_vsphere_virtual_machine.go
@@ -83,6 +83,12 @@ func dataSourceVSphereVirtualMachine() *schema.Resource {
 				Computed:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
+			"guest_ip_addresses": {
+				Type:        schema.TypeList,
+				Description: "The current list of IP addresses on this virtual machine.",
+				Computed:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


koikonom:
Hi @sirlatrom, thanks for taking the time to submit this PR! 

Looks good to me.

sirlatrom:
@koikonom Thanks! Any chance you can have a look at the related #819?

PR #906, opened 2019-11-20
fix typo on no snapshots error check

rickhlx:
Currently, if a snapshot is manually deleted, the following error will happen:

```
Error: Error while finding the Snapshot :no snapshots for this VM
```

The `strings.Contains` being used to compare is case sensitive, so the comparison is never `true`


rickhlx:
Fixes #908

PR #918, opened 2019-12-02
Acceptance test fixes

koikonom:
This PR addresses a number of issues caused the integrations test to fail.

- Only reverts the snapshot once. Flaky tests are still something we should be looking into. In the long run we should improve the lab resources.
- Addresses a couple of genuine bugs that were discovered by the tests.
  - f2a08c2041359cce7da90f323b69b2ae92228452 addresses a race condition when powering up VMs.
  - 97337c2308eac05c97a7d4732454b8378562c722 fixes import for compute clusters.
- Fixes some resource names that were needed after applying some linter changes.
- Changes the detection of certain error messages that changed after using the new tag library.
- Reverts PR #899 until there is a new version of govmomi released, that contains the fix required.


vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -367,6 +367,40 @@ func skipIPAddrForWaiter(ip net.IP, ignoredGuestIPs []interface{}) bool {
 	return false
 }
 
+func blockUntilReadyForMethod(method string, vm *object.VirtualMachine, waitFor time.Duration) error {
+	log.Printf("[DEBUG] blockUntilReadyForMethod: Going to block until %q is no longer in the Disabled Methods list for vm %s", method, vm.Reference().Value)
```
aareet: Is there a way to configure logging by log level so this isn't always printed?

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -441,6 +475,19 @@ func PowerOn(vm *object.VirtualMachine) error {
...
+	if err != nil {
+		return err
+	}
+
+	// This is the controversial part. Although we take every precaution to make sure the VM
+	// is in a state that can be started we have noticed that vsphere will randomly fail to
+	// power on the vm with "InvalidState" errors.
+	//
+	// We're adding a small delay here to avoid this issue.
+	time.Sleep(500 * time.Millisecond)
```
aareet: Should we pull this out into a variable so it can be configured at the top of the file? We might need to play around with the length of time we wait until we get it right.

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
koikonom: The config library will take care of that.

PR #955, opened 2020-01-24
Use govmomi REST client

bill-rich:


koikonom:
There were some conflicts on go.mod and go.sum that I couldn't for some reason fix via the UI, so I did it on my local branch, rebased on top of origin/master and then pushed with `--force-with-lease`.

Issue #940, opened 2020-01-10
vcsim Terraform support


Satak:
Does Terraform vSphere provider support VMware **vcsim** simulator? I'm having hard time getting Terraform working against vcsim. Opened a issue here for VMware: https://github.com/vmware/govmomi/issues/1779

It would really speed up Terraform development if one could run templates against vcsim instead of an actual vSphere. Are there any other alternatives for rapid Terraform development for vSphere?

**Status June 2020**: Terraform can provision VMs against vcsim but only without clone block. Clone block will cause Terraform crash!
Crash log here: https://gist.github.com/Satak/ec149546617397bc41f70291d64fe030

bill-rich:
Hi @Satak! We haven't done much in the area of testing with vcsim. Based on the error in you govmomi issue, it looks like vcsim is probably missing a few attributes required by the Terraform provider, but it is possible it is a configuration issue also. 

As you said in the other issue, I think the first step would be defining a simple working configuration to see if we could get that working, and then expand from there to see what does or does not work.

Satak:
So I have been playing around with this vcsim and govc and managed to pinpoint the problem to this `customize {}` block. This tf config works fine against vcsim but only if you don't use customize. If you use customize block it will give you an error:
`Error: cannot find OS family for guest ID "otherGuest": could not find guest ID "otherGuest"`

## `main.tf`

```terraform
provider "vsphere" {
  user                 = "username"
  password             = "password"
  vsphere_server       = "localhost"
  allow_unverified_ssl = true
}

data "vsphere_datacenter" "dc" {
  name = "DC0"
}
data "vsphere_datastore" "datastore" {
  name          = "datastore/LocalDS_0"
  datacenter_id = data.vsphere_datacenter.dc.id
}
data "vsphere_resource_pool" "pool" {
  name          = "DC0_H0/Resources"
  datacenter_id = data.vsphere_datacenter.dc.id
}
data "vsphere_network" "network" {
  name          = "network/VM Network"
  datacenter_id = data.vsphere_datacenter.dc.id
}
data "vsphere_virtual_machine" "template" {
  name          = "DC0_H0_VM0"
  datacenter_id = data.vsphere_datacenter.dc.id
}

resource "vsphere_virtual_machine" "vm" {
  name             = "terraform-test"
  resource_pool_id = data.vsphere_resource_pool.pool.id
  datastore_id     = data.vsphere_datastore.datastore.id
  num_cpus         = 2
[...]

PR #959, opened 2020-01-29
Optimize OS family query

dougm:
Optimize OS family query

Switching from QueryConfigOption to QueryConfigOptionEx to query the specific descriptor,
rather than returning all supported descriptors.
Against vCenter 6.7U3, this reduces the response from ~600K to ~10K

This change will also simplify the vcsim implementation of this query (Issue #940)

### DIFF
--- a/vsphere/internal/helper/envbrowse/environment_browser_helper.go
+++ b/vsphere/internal/helper/envbrowse/environment_browser_helper.go
@@ -77,10 +77,13 @@ func (b *EnvironmentBrowser) OSFamily(ctx context.Context, guest string) (string
 		return "", err
 	}
 
-	req := types.QueryConfigOption{
+	req := types.QueryConfigOptionEx{
 		This: b.Reference(),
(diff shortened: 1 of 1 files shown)


koikonom:
Hi @dougm, 

I tried running the acceptance tests on this branch and they failed. 

After some investigation it seems like we get back nil, once you set the key to the guest os name. Could it be possible that instead of the gust os name we should be passing some key that represents the list of all guest OSs ?

dougm:
Hi @koikonom , sorry, I've yet to run the acceptance tests myself. The proper change needs to use `QueryConfigOptionEx`, I'll push that update. I'll also look at how to run the acceptance tests myself.

koikonom:
HI @dougm, 

This looks OK and a couple of the tests I tried passed. We're in the middle of preparing the next release that has some big changes so I'd rather hold this one for the one after that.

dougm:
Sounds good, thanks @koikonom

PR #963, opened 2020-02-03
Properly set up environment for REST session persistence

bill-rich:


vsphere/config.go
```diff
@@ -534,8 +559,8 @@ func (c *Config) LoadAndVerifyRestSession(client *govmomi.Client) (*rest.Client,
 				Name:  "vmware-api-session-id",
 				Value: sessionId,
 			}
-			client.Jar.SetCookies(cookiePath, append(cookies, &newcookie))
 			restClient = rest.NewClient(client.Client)
+			restClient.Jar.SetCookies(cookiePath, append(cookies, &newcookie))
 		}
 	}
 	if restSessionValid(restClient) {
```
koikonom: Would it make sense to add a check here if sessionId is equal to "" ?

bill-rich: Already got that on [557](https://github.com/terraform-providers/terraform-provider-vsphere/blob/4465c913018191c1659074cb33b0bf545dc795a0/vsphere/config.go#L557). It just got cut off in the diff. Good call though!

Issue #968, opened 2020-02-04
Error: disk.0: virtual machine with UUID "" not found (1.16)


mitchellmaler:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform
issue trackers reserved for bug reports and feature requests. For general usage
questions, please see: https://www.terraform.io/community.html.

### Terraform Version

Terraform v0.12.20

### vSphere Provider Version

provider.vsphere v1.16.0

### Affected Resource(s)

Please list the resources as a list, for example:
- `vsphere_virtual_machine`
- `vsphere_distributed_port_group`

### Terraform Configuration Files

```hcl
resource "vsphere_virtual_machine" "vm" {
  depends_on       = [tls_private_key.vm_terraform_ssh_key]
  name             = var.vsphere_vm_name
  resource_pool_id = var.vsphere_resource_pool_id
  datastore_id     = var.vsphere_datastore.id
  folder           = var.vsphere_vm_folder_path

  num_cpus = var.vsphere_virtual_machine_cpu_count
  memory   = var.vsphere_virtual_machine_memory_mb
  guest_id = var.vsphere_vm_template.guest_id

  scsi_type = var.vsphere_vm_template.scsi_type

  network_interface {
    network_id   = var.vsphere_network.id
    adapter_type = var.vsphere_vm_template.network_interface_types[0]
  }

  disk {
    label            = "disk0"
    size             = var.vsphere_vm_template.disks.0.size + var.vsphere_virtual_machine_srv_extend_size
    eagerly_scrub    = var.vsphere_vm_template.disks.0.eagerly_scrub
    thin_provisioned = var.vsphere_vm_template.disks.0.thin_provisioned
  }

  extra_config = {
    "guestinfo.userdata" = base64gzip(
      templatefile("${path.module}/templates/userdata.yml", {
        vm_hostname            = var.vsphere_vm_name,
        vm_domain              = var.vsphere_vm_domain,
        vm_rancher_ssh_pub_key = tls_private_key.vm_terraform_ssh_key.public_key_openssh
    }))
    "guestinfo.userdata.encoding" = "gzip+base64"
  }

  clone {
    template_uuid = var.vsphere_vm_template.id
  }
}
```

### Debug Output

https://gist.github.com/mitchellmaler/cd290f3c33bd7d1723e661e1c37a7bbf

### Expected Behavior

What should have happened?

[...]

PR #970, opened 2020-02-06
Base storage_policy_id off vm

bill-rich:
Set the storage_policy_id based off the VM uuid rather than the template.


bill-rich:
Fixes #968

PR #979, opened 2020-02-25
In vsphere_file replace FileManager with Datastore FileManager

jcpowermac:
When using `vsphere_file` to upload then copy a VMDK
the file is invalid - error: `Unsupported or invalid disk type 22 for 'scsi0:0'.`

Replace FileManager with DatastoreFileManager and use its `Copy` method
that properly manages the copy of VMDK.


jcpowermac:
In testing I found an issue.  So I am closing this pr. Will reopen a new PR with additional changes.
The DatastoreFileManager Copy method uses the datastore object to create the path.
The problem is if you copy between datastores this will fail if path is not handled before calling `Copy`.

PR #985, opened 2020-03-04
Content Library

bill-rich:
Adding support for content library as a resource, data source, and clone source for virtual machines.

### DIFF
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -506,7 +506,7 @@ func DeployDest(name string, annotation string, rp *object.ResourcePool, host *o
 }
 
 func Deploy(c *rest.Client, item *library.Item, deploy *vcenter.Deploy, timeout int) (*types.ManagedObjectReference, error) {
-	log.Printf("[DEBUG] virtualmachine.Deply: Deploying VM from Content Library item %s", item.Name)
+	log.Printf("[DEBUG] virtualmachine.Deploy: Deploying VM from Content Library item %s", item.Name)
 	m := vcenter.NewManager(c)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*time.Duration(timeout))
 	defer cancel()
@@ -518,7 +518,7 @@ func Deploy(c *rest.Client, item *library.Item, deploy *vcenter.Deploy, timeout 
 		}
 		return nil, err
 	}
-	log.Printf("[DEBUG] virtualmachine.Deply: Successfully deployed VM from Content Library item %s", item.Name)
+	log.Printf("[DEBUG] virtualmachine.Deploy: Successfully deployed VM from Content Library item %s", item.Name)
 	return mo, nil
 }
 
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/vmware/govmomi/vapi/library"
+	"github.com/vmware/govmomi/vapi/rest"
+	"github.com/vmware/govmomi/vapi/vcenter"
 	"log"
 	"net"
 	"strings"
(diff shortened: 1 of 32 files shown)


vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -464,6 +467,59 @@ func Clone(c *govmomi.Client, src *object.VirtualMachine, f *object.Folder, name
...
+			NetworkMappings:     nm,
+			StorageMappings:     sm,
+			StorageProvisioning: "",
+			StorageProfileID:    "",
+			Locale:              "",
+			Flags:               nil,
+			AdditionalParams:    nil,
+			DefaultDatastoreID:  "",
+		},
+		Target: vcenter.Target{
```
koikonom: What happens if these are empty strings?

vsphere/resource_vsphere_content_library.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+	d.SetId(lib.ID)
+	sb := contentlibrary.FlattenStorageBackings(lib.Storage)
+	d.Set("name", lib.Name)
+	d.Set("description", lib.Description)
+	err = d.Set("storage_backing", sb)
+	if err != nil {
+		return err
```
koikonom: It would be nice if you could put some prefix in front of the error message to make tracking them in the code easier.

vsphere/resource_vsphere_content_library.go
```diff
@@ -0,0 +1,76 @@
...
+			},
+		},
+	}
+}
+
+func resourceVSphereContentLibraryRead(d *schema.ResourceData, meta interface{}) error {
+	c := meta.(*VSphereClient).restClient
+	lib, err := contentlibrary.FromName(c, d.Get("name").(string))
+	if err != nil {
+		return err
```
koikonom: If the library is not found because someone deleted it outside terraform we should be handling it gracefully here. 
I think setting an empty ID and returning nil at that point will remove the object from state. Like 
```suggestion
		if viapi.IsManagedObjectNotFoundError(err) {
			d.SetId("")
			return nil
		}
		return fmt.Errorf("error while searching library %s. Error: %s ", d.Get("name").(string), err)
```

vsphere/resource_vsphere_content_library_item.go
```diff
@@ -0,0 +1,86 @@
...
+		return err
+	}
+	return contentlibrary.DeleteLibraryItem(rc, item)
+}
+
+func resourceVSphereContentLibraryItemRead(d *schema.ResourceData, meta interface{}) error {
+	rc := meta.(*VSphereClient).restClient
+	item, err := contentlibrary.ItemFromID(rc, d.Id())
+	if err != nil {
+		return err
```
koikonom: Same suggestion as I made for the Content Library Read() function.

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -848,7 +850,9 @@ func resourceVSphereVirtualMachineCustomizeDiff(d *schema.ResourceDiff, meta int
 			// flagging the imported flag to off.
 			d.SetNew("imported", false)
 		case d.Id() == "":
-			if err := vmworkflow.ValidateVirtualMachineClone(d, client); err != nil {
+			if contentlibrary.IsContentLibraryItem(meta.(*VSphereClient).restClient, d.Get("clone.0.template_uuid").(string)) {
+				// Do some content library checks
```
koikonom: I am guessing this is WIP?

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1191,25 +1195,71 @@ func resourceVSphereVirtualMachineCreateClone(d *schema.ResourceData, meta inter
...
+		host := &object.HostSystem{}
+		if hid := d.Get("host_system_id").(string); hid != "" {
+			host, err = hostsystem.FromID(client, hid)
+			if err != nil {
+				return nil, err
+			}
+		}
+		sm := []vcenter.StorageMapping{}
+		disks := d.Get("disk").([]interface{})
+		log.Printf("BILLLLLLLLLLLLLLLLLLLLLLLLLLLLLLL - %v", disks)
```
koikonom: Hi Bill! :D

website/docs/r/content_library.html.markdown
```diff
@@ -0,0 +1,68 @@
...
+[tf-vsphere-vm-resource]: /docs/providers/vsphere/r/virtual_machine.html
+
+```hcl
+data "vsphere_datacenter" "dc" {
+  name = "dc1"
+}
+
+data "vsphere_datastore" "datastore" {
+  name          = "datastore1"
+  datacenter_id = "${data.vsphere_datacenter.dc.id}"
```
koikonom: should we remove ${} ?

bill-rich: Definitely should. Fixed it.

bill-rich:
Content Libraries are at the vCenter level, so they're above datacenters in the tree. The naming thing is kind of an odd one. You _can_ have two Content Libraries with the same name, but the `GetLibraryByName` function in govmomi returns the first one it comes accross with a matching name. We could use one of the listing functions to get the whole list back and filter by other criteria, but I think the use case of multiple libraries with the same name is going to be pretty uncommon.

PR #987, opened 2020-03-05
Update resource vsphere file to support vmdk files

jcpowermac:
Create two new functions:
- `createDirectory` - create directory on target datacenter and datastore
- `fileUpload` - upload file

Fix golinting issues
Change variable names explicit destination (dst) vs. source (src) to make it more clear.

Modified copyFile to use DatastoreFileManager that supports the proper `Copy` of VMDK files.

Added branch for the upload VMDK case:
- Creates a temporary file location for the VMDK to be uploaded
- Upload with uploadFile
- Use DatastoreFileManager to `Move` the VMDK that will properly move the disk

Modified upload case to use uploadFile

### DIFF
--- a/vsphere/resource_vsphere_file.go
+++ b/vsphere/resource_vsphere_file.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"path"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


jcpowermac:
cc: @abhinavdahiya

abhinavdahiya:
Should we add a test for this?

PR #990, opened 2020-03-10
Add configurable timeout for VM PowerOn operations

koikonom:
There is a conflict right now between the way the vsphere provider
builds VMs and how vSphere expects VMs that are part of a vApp to
behave, especially in the way things are powered on.

vApps expect that the VMs will be built but not powered on, because it
handles powering on the VM on its own in a specific order. On the other
hand the provider will build the VM and instantly power it on regardless
of where it belongs.

The result of this conflict is that vsphere will not allow any VM to
power on until all VMs in the vApp are ready to be powered on. So for
example while cloning two VMs in the same vApp, we are only allowed to
power on the first VM once both of them are done cloning. Any attempts
to power on a VM earlier result in a "InvalidState" response.

This commit introduces a configurable timeout, `poweron_timeout`, that
represents the amount of time we are going to give a VM to power on. The
default (and minimum) is 300 seconds (5 minutes).

The full power on process goes like this:

- Timer starts (for as long as poweron_timeout is set to).
- First we poll the `Disabled Methods` list for the VM object until
PowerOnVM_Task is no longer in it.
- Next we enter a loop for the remainder of the time where we issue
PowerOnVM requests, and check the response. If it's successful we
return, otherwise we check the error message. If the error message is
about the VM being in an invalid state then we sleep for 500ms and
retry, otherwise we return an error.
- If we hit the timeout and we have not been able to successfully power
on the VM so far, we return an error.

### DIFF
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -518,8 +518,6 @@ powerLoop:
 					log.Printf("[DEBUG] Failed to submit PowerOn task for vm %q. Error: %s", vmPath, err)
 					return fmt.Errorf("failed to submit poweron task for vm %q: %s", vmPath, err)
 				}
-				//tctx, tcancel := context.WithTimeout(context.Background(), provider.DefaultAPITimeout)
-				//defer tcancel()
 				err = task.Wait(ctx)
 				if err != nil {
 					if err.Error() == "The operation is not allowed in the current state." {
--- a/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
+++ b/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
@@ -376,10 +376,8 @@ func skipIPAddrForWaiter(ip net.IP, ignoredGuestIPs []interface{}) bool {
 	return false
 }
 
-func blockUntilReadyForMethod(method string, vm *object.VirtualMachine, waitFor time.Duration) error {
+func blockUntilReadyForMethod(method string, vm *object.VirtualMachine, ctx context.Context) error {
 	log.Printf("[DEBUG] blockUntilReadyForMethod: Going to block until %q is no longer in the Disabled Methods list for vm %s", method, vm.Reference().Value)
-	ctx, cancel := context.WithTimeout(context.TODO(), waitFor)
-	defer cancel()
 
 	for {
 		vprops, err := Properties(vm)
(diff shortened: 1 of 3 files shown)


vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1329,7 +1347,12 @@ func resourceVSphereVirtualMachineCreateClone(d *schema.ResourceData, meta inter
 		}
 	}
 	// Finally time to power on the virtual machine!
-	if err := virtualmachine.PowerOn(vm); err != nil {
+	pTimeoutStr := fmt.Sprintf("%ds", d.Get("poweron_timeout").(int))
+	pTimeout, err := time.ParseDuration(pTimeoutStr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse poweron_timeout as a valid duration: %s", err)
+	}
```
bill-rich: Might be worth adding an IntToString function to the structure helper since you've its been used a couple times.

rchadalawada:
@koikonom Why are you resorting to time outs? Why not check if the VM is part of a vApp and skip power on? Wouldn't it make sense to treat vApp as a single entity.

koikonom:
Hey @rchadalawada, great question!

In terraform we have two vApp related resources, one is the vapp_container and the other is the vapp_entity. Unfortunately none of them is able to control powering on the VM because neither of them is aware of the full picture.  The container just creates the vApp and vapp_entity can be used to associate VMs to the vApp, manage their startup order etc.  When the provider runs each resource is only given its own configuration, as a result neither of those entities is aware of the full picture. 

You are correct when you say that I could skip powering on the VM if it's a member of a vApp, but then there is no resource to control the power status of the vApp once everything is built.

In order to do this right we would have to revisit many basic elements of how the VMs are built and how we manage vApps as a whole and this is a much larger project that needs to be prioritised accordingly.

vsphere/internal/helper/virtualmachine/virtual_machine_helper.go
```diff
@@ -494,16 +500,44 @@ func PowerOn(vm *object.VirtualMachine) error {
...
+			}
+			if vprops.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOff {
+				log.Printf("[DEBUG] VM %q is powered off, attempting to power on.", vmPath)
+				task, err := vm.PowerOn(ctx)
+				if err != nil {
+					log.Printf("[DEBUG] Failed to submit PowerOn task for vm %q. Error: %s", vmPath, err)
+					return fmt.Errorf("failed to submit poweron task for vm %q: %s", vmPath, err)
+				}
+				//tctx, tcancel := context.WithTimeout(context.Background(), provider.DefaultAPITimeout)
+				//defer tcancel()
```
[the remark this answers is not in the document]
koikonom: Yup, will do.

vsphere/resource_vsphere_virtual_machine.go
koikonom: I decided to change the way the timeout is calculated, so this is no longer used.

PR #1014, opened 2020-03-31
r/virtual_machine: Ignore guest_id with content library

bill-rich:


vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1261,6 +1261,9 @@ func resourceVSphereVirtualMachineCreateClone(d *schema.ResourceData, meta inter
 		if err != nil {
 			return nil, err
 		}
+		// There is not currently a way to pull config values from Content Library items. If we do not send the values,
+		// the defaults from the template will be used.
+		d.Set("guest_id", "")
```
Satak, line 1266: Could we have an in statement before nulling this value? If user has passed guest_id it should null that out. Now this prevents any kind of customization to the VM.

PR #1038, opened 2020-04-13
Add issue triage actions

aareet:
### Description

Add hashibot configuration to lock closed issues and close stale issues. This config will:
* Close issues older than 180 days as stale, after giving a 30 day warning
* Lock issues that are closed for more than 30 days if there has been no activity

### Acceptance tests


.hashibot.hcl
```diff
@@ -0,0 +1,11 @@
+poll "closed_issue_locker" "locker" {
+    schedule = "0 50 13 * * *"
```
appilon, line 2: Unfortunately a lot of providers are running at the same time, I would suggest picking a new time, a shortcoming of hashicorp (we want to avoid API limiting)
```
schedule = "0 10 5 * * *"
```

.hashibot.hcl
```diff
@@ -0,0 +1,11 @@
...
+    closed_for = "720h" # 30 days
+    max_issues = 50
```
appilon, line 4: You can crank this number a bit, otherwise it will take many many many days/runs to burn down the backlog.
```
max_issues = 500
```

.hashibot.hcl
```diff
@@ -0,0 +1,11 @@
...
+    sleep_between_issues = "10s"
```
appilon, line 5: this can be a shorter sleep
```
sleep_between_issues = "5s"
```

aareet:
Thanks Alex, I've added this too -> no_comment_if_no_activity_for

.hashibot.hcl
```diff
@@ -0,0 +1,12 @@
+poll "closed_issue_locker" "locker" {
+    schedule = "0 10 5 * * *"
+    closed_for = "720h" # 30 days
+    no_comment_if_no_activity_for = "360h" # 15 days
```
appilon, line 4: I just realized that since you are locking another that's been closed for 30 days, that means likely EVERY issue being locked hasn't had activity for 15 days, meaning none of these issues will have comments. I would recommend setting the value for `no_comment_if_no_activity_for` to maybe 60 days `1440h`?

aareet: Oh that makes sense - I misunderstood the behaviour of that

Issue #1049, opened 2020-04-20
VM changes after 1.17.0 fail with error due to hardware version


stuart-c:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v0.12.24

### vSphere Provider Version

provider.vsphere 1.17.0 - wprks
provider.vsphere 1.17.1 (and later) - does not work

### Affected Resource(s)

vsphere_virtual_machine

### Debug Output

```xml
<Envelope xmlns="http://schemas.xmlsoap.org/soap/envelope/">
    <Body>
        <ReconfigVM_Task xmlns="urn:vim25">
            <_this type="VirtualMachine">vm-88664</_this>
            <spec>
                <name>metrics01-eqx</name>
                <version>vmx-9</version>
                <guestId>ubuntu64Guest</guestId>
                <annotation>Managed via Terraform</annotation>
                <tools>
                    <afterPowerOn>true</afterPowerOn>
                    <afterResume>true</afterResume>
                    <beforeGuestStandby>true</beforeGuestStandby>
                    <beforeGuestShutdown>true</beforeGuestShutdown>
                    <beforeGuestReboot>false</beforeGuestReboot>
                    <syncTimeWithHost>false</syncTimeWithHost>
                </tools>
```

### Actual Behavior

Error: The operation is not supported on the object.

Within vSphere the errors were:

* A specified parameter was not correct:
* The operation is not supported on the object.

### Steps to Reproduce

* Make a VM change (e.g. number of cpus)
* Run with version 1.17.0 - works
* Try name with version 1.17.1 or later
* Fails

Looking at the debug logs the call which fails (ReconfigVM_Task) the only difference between the two versions is the <version>vmx-9</version> tag is sent for 1.17.1 and later, which causes it to fail.

### Important Factoids

vSphere Client version 6.7.0.42000

### References

* #1020

stuart-c:
It looks like this causes it to always include the <version> tag:

[...]

PR #1055, opened 2020-04-26
Only include hardware version in CreateSpecs

bill-rich:
It seems that the vSphere API will return an error when some older hardware versions are included in the UpdateSpec. The hardware version is only needed in the CreateSpec, but the two share the Flatten function. Rather than splitting that or making the Flatten function determine if it is a Create or Update spec, I am removing the extre version info when updating.


bill-rich:
Fixes #1049

PR #1085, opened 2020-05-20
Specify cluster membership management

bill-rich:
Adding management arguments to `hosts` and `compute_clusters` so that the two do not conflict when both resources are in use.

### DIFF
--- a/vsphere/resource_vsphere_compute_cluster.go
+++ b/vsphere/resource_vsphere_compute_cluster.go
@@ -133,11 +133,12 @@ func resourceVSphereComputeCluster() *schema.Resource {
 				Description: "The managed object ID of the datacenter to put the cluster in.",
 			},
 			"host_system_ids": {
-				Type:        schema.TypeSet,
-				Optional:    true,
-				MaxItems:    64,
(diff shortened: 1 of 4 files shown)


vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -133,11 +133,12 @@ func resourceVSphereComputeCluster() *schema.Resource {
...
-				Optional:    true,
-				MaxItems:    64,
-				Description: "The managed object IDs of the hosts to put in the cluster.",
-				Elem:        &schema.Schema{Type: schema.TypeString},
+				Type:          schema.TypeSet,
+				Optional:      true,
+				MaxItems:      64,
+				Description:   "The managed object IDs of the hosts to put in the cluster.",
+				Elem:          &schema.Schema{Type: schema.TypeString},
+				ConflictsWith: []string{"host_managed"},
```
jrhouston, line 141: I actually didn't know about this option, _nice_!

bill-rich: It is certainly a nice one!

PR #1117, opened 2020-06-26
Updates imports for org move

bill-rich:


website/docs/r/virtual_machine.html.markdown.orig
```diff
@@ -0,0 +1,1637 @@
+---
```
dak1n1, line 1: Did you mean to include a file that ends with `.orig`?

bill-rich: I didn't. Thanks for catching that! Got it removed now.

bill-rich:
It was super easy thanks to your doc!

vsphere/internal/virtualdevice/virtual_machine_device_subresource.go.orig
```diff
@@ -0,0 +1,986 @@
+package virtualdevice
```
dak1n1, line 1: Another `.orig` file... I wonder if you meant to include this?

bill-rich: Thanks! Looks like that was the last one.

PR #1118, opened 2020-06-27
Support for SATA and IDE disks

bill-rich:
### Acceptance tests

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
@@ -547,48 +547,65 @@ func swapSCSIDevice(l object.VirtualDeviceList, device types.BaseVirtualSCSICont
 	return spec, nil
 }
 
-// NormalizeSCSIBus checks the SCSI controllers on the virtual machine and
+// NormalizeBus checks the SCSI controllers on the virtual machine and
 // either creates them if they don't exist, or migrates them to the specified
 // controller type. Devices are migrated to the new controller appropriately. A
 // spec slice is returned with the changes.
 //
 // The first number of slots specified by count are normalized by this
 // function. Any others are left unchanged.
-func NormalizeSCSIBus(l object.VirtualDeviceList, ct string, count int, st string) (object.VirtualDeviceList, []types.BaseVirtualDeviceConfigSpec, error) {
-	log.Printf("[DEBUG] NormalizeSCSIBus: Normalizing first %d controllers on SCSI bus to device type %s", count, ct)
+func NormalizeBus(l object.VirtualDeviceList, d *schema.ResourceData) (object.VirtualDeviceList, []types.BaseVirtualDeviceConfigSpec, error) {
+	scsiCount := d.Get("scsi_controller_count").(int)
+	scsiType := d.Get("scsi_type").(string)
+	scsiSharing := d.Get("scsi_bus_sharing").(string)
+	sataCount := d.Get("sata_controller_count").(int)
+	ideCount := d.Get("ide_controller_count").(int)
 	var spec []types.BaseVirtualDeviceConfigSpec
-	ctlrs := make([]types.BaseVirtualSCSIController, count)
+	scsiCtlrs := make([]types.BaseVirtualSCSIController, scsiCount)
+	sataCtlrs := make([]types.BaseVirtualSATAController, sataCount)
+	ideCtlrs := make([]*types.VirtualIDEController, ideCount)
 	// Don't worry about doing any fancy select stuff here, just go thru the
 	// VirtualDeviceList and populate the controllers.
-	iCount := int32(count)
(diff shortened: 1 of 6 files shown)


vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
```diff
@@ -547,57 +548,98 @@ func swapSCSIDevice(l object.VirtualDeviceList, device types.BaseVirtualSCSICont
...
+		case types.BaseVirtualSATAController:
+			if busNumber := ctlr.GetVirtualSATAController().BusNumber; busNumber < int32(sataCount) {
+				sataCtlrs[busNumber] = ctlr
+			}
+		case *types.VirtualIDEController:
+			if busNumber := ctlr.GetVirtualController().BusNumber; busNumber < int32(ideCount) {
+				ideCtlrs[busNumber] = ctlr
 			}
 		}
 	}
```
koikonom, line 588: Just wondering if it would make sense at this point to check if all three controller lists are full or not. Just a quick check that the device list contained all the devices we expected to find.

vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
```diff
@@ -547,57 +548,98 @@ func swapSCSIDevice(l object.VirtualDeviceList, device types.BaseVirtualSCSICont
...
 			return nil, nil, err
 		}
 		spec = append(spec, cspec...)
 		l = applyDeviceChange(l, cspec)
 		continue
 	}
-	log.Printf("[DEBUG] NormalizeSCSIBus: Outgoing device list: %s", DeviceListString(l))
-	log.Printf("[DEBUG] NormalizeSCSIBus: Outgoing device config spec: %s", DeviceChangeString(spec))
+	log.Printf("[DEBUG] NormalizeBus: Current SATA bus contents: %s", sataControllerListString(sataCtlrs))
+	// Now iterate over the controllers
```
koikonom, line 618: I think it would improve readability if the comment said which type of controller we are iterating

vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
```diff
@@ -617,6 +659,24 @@ func setSCSIBusSharing(l *object.VirtualDeviceList, ctlr types.BaseVirtualSCSICo
 	return cspec, nil
 }
 
+// createIDEController creates a new IDE controller.
+func createIDEController(l *object.VirtualDeviceList, bus int) ([]types.BaseVirtualDeviceConfigSpec, error) {
+	ide, _ := l.CreateIDEController()
+	cspec, err := object.VirtualDeviceList{ide}.ConfigSpec(types.VirtualDeviceConfigSpecOperationAdd)
+	*l = applyDeviceChange(*l, cspec)
```
koikonom, line 666: Idle thought: how much does this API request "cost" in terms of time ? Is there a way to gather all controller changes and apply them in a single go ?

vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1122,29 +1136,49 @@ func resourceVSphereVirtualMachineImport(d *schema.ResourceData, meta interface{
...
+			scsiBus[dev.GetVirtualSCSIController().BusNumber] = true
+		case types.BaseVirtualSATAController:
+			sataBus[dev.GetVirtualSATAController().BusNumber] = true
+		case *types.VirtualIDEController:
+			ideBus[dev.GetVirtualController().BusNumber] = true
 		}
-		scsiBus[sc.GetVirtualSCSIController().BusNumber] = true
 	}
-	var ctlrCnt int
+	var scsiCtlrCnt int
```
koikonom, line 1156: This loop for each bus list could be converted into a small function, since all three lists are of the same type

vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
bill-rich: The changes being applied at this point are actually all to a virtual set of devices, so there aren't any API calls to do these. After all the device changes are processed, a reconfig spec is generated and then finally applied to the real VM on vSphere here: https://github.com/hashicorp/terraform-provider-vsphere/blob/master/vsphere/resource_vsphere_virtual_machine.go#L748.

vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
bill-rich: Good call. Got that fixed.

vsphere/internal/virtualdevice/virtual_machine_device_subresource.go
bill-rich: It could help by allowing us to skip the next few loops, but it is ok to not have all the expected controllers found. In the next section of the function, any controllers that are `nil` are created. This is where new or missing controllers are found and created.

vsphere/resource_vsphere_virtual_machine.go
bill-rich: That does look much better.

PR #1120, opened 2020-07-01
Add import for vsphere_virtual_disk

skevir:
### Description
Add function to import a vsphere_virtual_disk resource via `terraform import`.

### Acceptance tests
- [X] Have you added an acceptance test for the functionality being added?
- [X] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)
        https://gist.github.com/skevir/1a5df485691d979a2e26b489a393e93f (only ran TestAccResourceVSphereVirtualDisk tests)

### References
https://github.com/hashicorp/terraform-provider-vsphere/issues/1113

### DIFF
--- a/vsphere/resource_vsphere_virtual_disk.go
+++ b/vsphere/resource_vsphere_virtual_disk.go
@@ -11,6 +11,7 @@ import (
 	"context"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/virtualdisk"
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
@@ -33,6 +34,9 @@ func resourceVSphereVirtualDisk() *schema.Resource {
 		Create: resourceVSphereVirtualDiskCreate,
 		Read:   resourceVSphereVirtualDiskRead,
 		Delete: resourceVSphereVirtualDiskDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceVSphereVirtualDiskImport,
+		},
 
 		Schema: map[string]*schema.Schema{
 			// Size in GB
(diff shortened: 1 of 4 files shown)


skevir:
Hi there,

Just wondering if you'd had a chance to look at this pull request yet? Any thoughts on whether it is likely to be accepted?

Thanks,

Richard.

vsphere/resource_vsphere_virtual_disk.go
```diff
@@ -465,3 +482,41 @@ func searchForDirectory(client *govmomi.Client, datacenter string, datastore str
...
+	client := meta.(*VSphereClient).vimClient
+	p := d.Id()
+	if !strings.HasPrefix(p, "/") {
+		return nil, errors.New("ID must start with a trailing slash")
+	}
+
+	// The path in the ID has the form `/<datacenter>/[<datastore>] path/to/vmdk`.
+	// Note that the values we care about in addrParts start at the first element as
+	// the zero-th element will be empty (that is, everything before the / prefix).
+	addrParts := strings.SplitN(p, "/", 3)
```
bill-rich, line 496: I think this should be increased to four as the expected path should be `/<dc-name>/datastore/<datastore-name>/path/to/datastore` to match the govc/vSphere convention.

skevir: Hi Bill,
Thanks for the suggestion. Perfectly happy to change to your suggested format, but just wanted to explain why I'd gone for the format that I had, just in case it changed your mind!
When I look in the vSphere web UI at the file associated with the virtual disk, the path that it provides is of the form I was suggesting (that is, `[<datastore>] path/to/vmdk`) so seemed like a sensible choice for what to use for the import. I did start with something closer to what you have suggested based on the format for other import commands, but changed to what is implemented currently after seeing the format in the UI. 
If I haven't convinced you, just let me know, and I'm happy to tweak the parameter parsing code accordingly.
Thanks,
Richard.

skevir: Hi Bill,
Looking at this a bit more today, I think the interface into govc (via govmomi's object.VirtualDiskManager) does use the format I am proposing, so I guess it becomes a choice between matching the govc/vSphere conventions or being more inline with the format for importing other resource types (or maybe those two things are actually the same, and it is just the format of an inventory path for a virtual disk (that is, a datastore path) is very different to the path for other resource types).
Would appreciate your thoughts here, following my comments. Are you happy for me to stick with `/<dc-name>/[<datastore-name>] /path/to/vmdk`, or would you still prefer me to change to `/dc-name>/datastore/<datastore-name>/path/to/vmdk` and convert from the latter into the former in the provider code?
Thanks,
Richard.

darracott:
Hiya,
Any updates on this?
Thanks!

bill-rich:
Sorry about the delay on this. I'll take another look today.

vsphere/resource_vsphere_virtual_disk.go
bill-rich: I guess I was mistaken on the govc side. Since thats the case, lets stick with the format you used initially. Thanks for the follow up!

PR #1142, opened 2020-07-24
Add host_thumbprint data source

bill-rich:
Needed for adding new hosts when the thumbprint is not known ahead of time.

### DIFF
--- a/vsphere/data_source_vsphere_host_thumbprint.go
+++ b/vsphere/data_source_vsphere_host_thumbprint.go
@@ -1,0 +1,49 @@
+package vsphere
+
+import (
+	"bytes"
+	"crypto/sha1"
+	"crypto/tls"
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceVSphereHostThumbprint() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceVSphereHostThumbprintRead,
+		Schema: map[string]*schema.Schema{
+			"address": {
+				Type:        schema.TypeString,
+				Required:    true,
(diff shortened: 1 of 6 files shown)


vsphere/data_source_vsphere_host_thumbprint.go
```diff
@@ -0,0 +1,49 @@
...
+				Default:     "443",
+				Description: "The port to connect to on the ESXi host.",
+			},
+		},
+	}
+}
+
+func dataSourceVSphereHostThumbprintRead(d *schema.ResourceData, meta interface{}) error {
+	config := &tls.Config{}
+	config.InsecureSkipVerify = true
```
koikonom, line 32: I am not sure this is always desired behaviour, better set it as an input with an optional value of `false`.

vsphere/data_source_vsphere_host_thumbprint.go
```diff
@@ -0,0 +1,49 @@
...
+	conn, err := tls.Dial("tcp", d.Get("address").(string)+":"+d.Get("port").(string), config)
+	if err != nil {
+		return err
+	}
+	cert := conn.ConnectionState().PeerCertificates[0]
```
koikonom, line 37: We should mention explicitly in the documentation that if the host is using a cert chain then we will only return the fingerprint of the first cert we find.

vsphere/data_source_vsphere_host_thumbprint.go
bill-rich: Thanks! I totally overlooked that when I was converting this from my original hacked together data source.

vsphere/data_source_vsphere_host_thumbprint.go
bill-rich: Good call. Added.

PR #1148, opened 2020-07-28
Added the label and unit_number fields to the disk-subresource schema.

arizvisa:
### Description
This PR allows the unit_number and label to be computable from the `vsphere_virtual_machine` that the disk is being cloned from. This removes the need to hardcode any of these fields using a variable or any other names as they're now exported by the disk resource's schema. This is useful if your disk naming scheme is actually imposed by the templates that are made available to you (which is how my environment is unfortunately).

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added? (No, but I can totally add them if there's interest in this PR)

### References
None that I know of...

### DIFF
--- a/vsphere/data_source_vsphere_virtual_machine.go
+++ b/vsphere/data_source_vsphere_virtual_machine.go
@@ -74,6 +74,10 @@ func dataSourceVSphereVirtualMachine() *schema.Resource {
 				Computed:    true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"label": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
 						"size": {
 							Type:     schema.TypeInt,
 							Computed: true,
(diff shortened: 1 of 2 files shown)


arizvisa:
If there's interest in this PR, I can try and work on some of the tests and such to meet your criteria for merging. I wasn't sure if there was some particular reason why these fields were not already exposed (hence my questions in this PR).

I also have another PR that also flattens the schema between the dataSource and the resource as it originally didn't make sense to me for both of these types to have different schemas especially when the logic appears the very same. If you guys are interested in that one, I can submit it as well and add its tests.

Ideally I'd prefer to have a lot more of the attributes exposed between both types (as well as made consistent where applicable since they're effectively the same), and would be willing to put the work in to make that happen if you guys are okay with it.

arizvisa:
Closing this PR as it's been superceded by PR #1157 which cleans up the code, adds tests, and updates the documentation.

Issue #1156, opened 2020-08-06
Unable to linked clone from a template due to no existing snapshots despite templates technically being a snapshot


arizvisa:
This issue is a placeholder for a pull request.

### Description
Making a linked-clone from a template is apparently not possible due to the assertion of it needing a snapshot. This assertion is re-iterated throughout the internet as when cloning from a non-template requires a snapshot. This makes sense for virtual machines because their disks would be in use which would interfere with other users of said disk. Templates, however, are essentially already snapshots and so therefore linked clones should be possible. Actually, technically, there's no real difference between templates and virtual machines other than a flag which tells VMware that it shouldn't tamper with the machine's disk. This technology is actually the same from VMWare's Workstation in which you're not allowed to modify the disks from a template hence being able to perform a full or linked clone from them.

The documentation at https://webcache.googleusercontent.com/search?q=cache:oVq53OL38-kJ:https://pubs.vmware.com/vsphere-50/topic/com.vmware.wssdk.pg.doc_50/PG_Ch11_VM_Manage.13.4.html+&cd=1&hl=en&ct=clnk&gl=us confirms this logic, and the available types are documented at https://vdc-download.vmware.com/vmwb-repository/dcr-public/b525fb12-61bb-4ede-b9e3-c4a1f8171510/99ba073a-60e9-4933-8690-149860ce8754/doc/vim.vm.RelocateSpec.DiskMoveOptions.html). 

According to the `vsphere/internal/vmworkflow/virtual_machine_clone_subresource.go`, Terraform's vSphere provider is using the `types.VirtualMachineRelocateDiskMoveOptionsCreateNewChildDiskBacking` to clone all virtual machines. However, if the `types.VirtualMachineRelocateDiskMoveOptionsMoveAllDiskBackingsAndAllowSharing` is used then the backing disk of the source template will be moved to the correct datastore and then shared with all virtual machines that use it. This would then allow performing linked-clones with a template or even without a snapsho.

It might be good to expose all of the available disk relocation types from the docs somehow, but I'd personally prefer the simpler interface that you guys have already implemented.

Cross-checking against the implementation of `vm.clone` from `govc` seems to include the following logic when a snapshot is not found. Again, as there's other types that are available in the API, I'm not sure if this is the correct logic to do for any virtual machine other than a template where you can guarantee that the disk is locked as read-only. But it seems to work nonetheless.
```
        if cmd.snapshot == "" {
                if cmd.link {
[...]

Issue #831, opened 2019-08-26
Linked clones depend on template_uuid instead of snapshot_id


rickhlx:
### Terraform Version

v0.12.3

### vSphere Provider Version

v1.12.0

### Affected Resource(s)

Please list the resources as a list, for example:
- `vsphere_virtual_machine`
- `vsphere_virtual_machine_snapshot`

If this issue appears to affect multiple resources, it may be an issue with
Terraform's core, so please mention this.

### Terraform Configuration Files

```hcl
resource "random_string" "master_password" {
  length  = 16
  special = true
}

resource "random_string" "clones_password" {
  count   = var.clone_count
  length  = 16
  special = true
}

resource "vsphere_virtual_machine" "rdsh_master" {
  name             = var.master_name
  datastore_id     = data.vsphere_datastore.ds.id
  num_cpus         = 4
  memory           = 4096
  guest_id         = data.vsphere_virtual_machine.master_template.guest_id
  resource_pool_id = data.vsphere_compute_cluster.cluster.resource_pool_id
  scsi_type        = data.vsphere_virtual_machine.master_template.scsi_type
  annotation       = "Created by Terraform"
  firmware         = data.vsphere_virtual_machine.master_template.firmware

  wait_for_guest_net_timeout = 0

  network_interface {
    network_id   = data.vsphere_network.network.id
    adapter_type = "vmxnet3"
  }

  disk {
    label            = "disk0"
    size             = var.clone_disk_size
    thin_provisioned = true
    eagerly_scrub    = false
  }

  clone {
    template_uuid = data.vsphere_virtual_machine.master_template.id

    customize {

      windows_options {
        computer_name     = "USRDSH-PC-1"
        organization_name = "Org"
        workgroup         = "WORKGROUP"
        time_zone         = 35
        admin_password    = random_string.master_password.result
      }
      network_interface {}

      timeout = 30
    }
  }

}

resource "vsphere_virtual_machine_snapshot" "vm_snapshot" {
  virtual_machine_uuid = vsphere_virtual_machine.rdsh_master.id
  snapshot_name        = "baseline"
  description          = "Managed Snapshot by Terraform for Linked Clones"
  memory               = "false"
  quiesce              = "false"
  remove_children      = "true"
  consolidate          = "true"
}

resource "vsphere_virtual_machine" "rdsh_clone" {
  name             = format("%s-%d", var.clone_prefix, count.index)
  count            = var.clone_count
[...]

PR #1158, opened 2020-08-06
Allow performing a linked-clone from a template without requiring snapshots

arizvisa:
### Description
This PR allows performing a linked clone from a template by using the "moveAllDiskBackingAndAllowSharing" option. When performing a clone, the source virtual machine is checked if it's marked as a template. If so, then it sets the option and proceeds with the clone. If it is not marked as a template, then it will proceed with the original logic which validates that a snapshot exists.

The documentation was updated to remove reference to the snapshot constraint.

### Acceptance tests
There were no tests for the cloning functionality, and so I wasn't sure how to proceed here.

### References
This closes issue #1156.

### DIFF
--- a/vsphere/internal/vmworkflow/virtual_machine_clone_subresource.go
+++ b/vsphere/internal/vmworkflow/virtual_machine_clone_subresource.go
@@ -140,6 +140,9 @@ func ValidateVirtualMachineClone(d *schema.ResourceDiff, c *govmomi.Client) erro
 // with no children, to make sure there is no ambiguity when selecting a
 // snapshot for linked clones.
 func validateCloneSnapshots(props *mo.VirtualMachine) error {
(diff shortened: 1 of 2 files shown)


arizvisa:
Added another commit to handle a case in `ValidateVirtualMachineClone` where I wasn't validating the machine being a template before it called `validateCloneSnapshots`. This wasn't caught before because I was initially using commit 7e3c4f38186fd7a4299658def6d9d9ac9dde0d8an (first one in my PR) in which I patched a way to break out of `validateCloneSnapshots` upon identification of the template flag.

Now, instead of tampering with the logic of `validateCloneSnapshots`, machines are checked for the template marking before it actually being called. This might help with the implementation of the solution to issue #831 where this function can instead be rewritten to validate a particular snapshot identifier and use that one.

PR #1161, opened 2020-08-07
Ignore thin_provisioned and eagerly_scrub during DiskPostCloneOperation

mtnbikenc:
### Description
In some environments the disk provisioning type can change silently when cloning due to vSphere storage policies.  Thin provisioning cannot be managed by Terraform and therefore should not attempt to set these value after cloning.

Error message:
```
ERROR                                              
ERROR Error: error reconfiguring virtual machine: error processing disk changes post-clone: disk.0: cannot change the value of "thin_provisioned" - (old: false new: true) 
ERROR                                              
ERROR   on ../../../../../tmp/openshift-install-068412821/master/main.tf line 1, in resource "vsphere_virtual_machine" "vm": 
ERROR    1: resource "vsphere_virtual_machine" "vm" { 
ERROR                                              
```

Attempting to ignore these changes have been unsuccessful using the following:
```
lifecycle {
    ignore_changes = [
      disk.0.eagerly_scrub,
      disk.0.thin_provisioned,
      ]
  }
```

### Acceptance tests

### References
https://github.com/hashicorp/terraform-provider-vsphere/issues/1028
https://github.com/hashicorp/terraform-provider-vsphere/pull/1075
https://github.com/hashicorp/terraform-provider-vsphere/issues/1125

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
@@ -1036,7 +1036,7 @@ func DiskPostCloneOperation(d *schema.ResourceData, c *govmomi.Client, l object.
 			//
 			// TODO: Remove "name" after 2.0.
 			switch k {
-			case "label", "path", "name", "datastore_id", "uuid":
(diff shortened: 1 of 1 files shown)


pryorda:
It appears this might of created a bug for us. Once we updated we ran into "failed to query disk type."

PR #1183, opened 2020-08-26
Content library advanced features

bill-rich:

Issue #1126, opened 2020-07-09
(Question) Clarification of how to use a content library, some process for referencing its items, and how to get at the template values of an item


arizvisa:
This is more of a question as it's not a bug or a feature request. There was a bug filed in issue #1012 which has been locked due to timeout. But I really couldn't quite understand how people are accomplish this. So I spent a lot of time trying to confirm how this thing is supposed to work which is this issue is sooo long, because it includes a huge explanation of my findings.

Disclaimer: I'm pretty newb at terraform and it took me a bit to find the "vsphere_content_library" and "vsphere_content_library_item" in the docs, so forgive me if it's just something that I've overlooked and there already is some way to reuse the attributes from a specific content library item.

## Description
So as I understand it, the attributes of a content library item aren't available due to the content library api being very minimalistic. Issue #1012 seems to describe a workaround (?) that says to deploy a template from the content library item into a datastore, and then use that vm as a data reference in terraform. This means that a resource is unable to use any attributes from the library item and will need to manually specify its specs.

## Confirming everything

The idea of the content library api being minimalistic is definitely true as its schema as exposed by govmomi ([govmomi/vapi/library#Library](https://godoc.org/github.com/vmware/govmomi/vapi/library#Library)) only gives ids, names, descriptions, and types. The schema for library items ([govmomi/vapi/library#Item](https://godoc.org/github.com/vmware/govmomi/vapi/library#Item)) also is pretty much the same.

However, from looking at documentation. It seems that the reason why the api is so minimalistic is because a content library is nothing more than a pub/sub service for monitoring changes in groups of files (an "item"). The vsphere automation sdk exposes a [com.vmware.content.library](https://vmware.github.io/vsphere-automation-sdk-python/vsphere/cloud/com.vmware.content.library.html#module-com.vmware.content.library.item_client) interface, but its functionality only provides a way for tracking changes to library items and creating a lock (DownloadSession or UploadSession) in order to read them or write to them (respectively) without interfering with any current users.

### Library Item "clients"

[...]

PR #1184, opened 2020-08-26
Add type to content library item data source

bill-rich:
Adding `type` to `content_library_item` so that different logic can be used based on the `type`.

Based on #1183. Will be size/S when that is merged.

#1126


arizvisa:
Thank you!

PR #1221, opened 2020-10-02
Fix ova deploy and tests

bill-rich:
Alloow `datastore_id` to be computed since it is not required for ovf deployment, and then run VM read at the end of creation.

### DIFF
--- a/vsphere/resource_vsphere_virtual_machine_test.go
+++ b/vsphere/resource_vsphere_virtual_machine_test.go
@@ -7178,11 +7178,17 @@ resource "vsphere_virtual_machine" "vm" {
   datacenter_id              = data.vsphere_datacenter.rootdc1.id
   host_system_id             = data.vsphere_host.roothost1.id
   wait_for_guest_net_timeout = 0
-  wait_for_guest_ip_timeout  = 1
+  wait_for_guest_ip_timeout  = 0
   num_cpus                   = 2
   ovf_deploy {
     remote_ovf_url = "%s"
   }
+	disk{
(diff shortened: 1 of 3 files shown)


vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -7178,11 +7178,17 @@ resource "vsphere_virtual_machine" "vm" {
   datacenter_id              = data.vsphere_datacenter.rootdc1.id
   host_system_id             = data.vsphere_host.roothost1.id
   wait_for_guest_net_timeout = 0
-  wait_for_guest_ip_timeout  = 1
+  wait_for_guest_ip_timeout  = 0
```
koikonom, line 7181: Why are we changing this? I'm concerned that it may surprise users depending on existing behaviour.

bill-rich: Thats just to help speed up the testing. Not all ovf/ova sources have vmware tools installed, so (along with speeding things up) I didn't want to get acc test failures based on that.

koikonom: You are right, I shouldn't be reviewing code this early :)

koikonom (approved):
LGTM

vsphere/resource_vsphere_virtual_machine_test.go
bill-rich: Ha! And I shouldn't be submitting a PR this late!

PR #1278, opened 2020-12-10
Make API timeout configurable when building VMs

koikonom:
Sometimes creating new VMs can take a while, especially when building
    bare VMs that have their OS provisioned by an external system. If that
    takes more than 5 minutes then the VM creation fails with a timeout.

    For this reason we are adding an `api_timeout` parameter in the
    provider's configuration schema that will allow users to override that
    timeout. The default value remains five (5) minutes.


bill-rich (approved):
LGTM

Issue #1300, opened 2021-01-05
REST API Keepalives Not Being Sent


happly:
### Terraform Version
v0.14.3
### vSphere Provider Version
v1.24.3

### Affected Resource(s)

resources that use the REST API. The REST session can timeout since REST Session keep alive POSTs are not occurring.

VM clones that take longer that the default vsphere REST API timeout will fail when the provider checks the tags of the newly cloned VM near the end of the clone operation.

### Debug Output

https://gist.github.com/happly/2ebb3cd0cb0058ae8b0f9c498e263961

First file in gist above shows behavior when cloning machine over network that takes longer that 45 minutes (our vsphere timeout); you can see that it does not attempt any keep alive POSTs. This deployment will successfully copy over the vm and then fail when it checks the tags at the end using the REST API.

The second file shows the output after running a patched version where I have enabled the keepalive functionality present in the govmomi backend. I will create a PR for this change but I wanted to present the issue here for a broader audience since I'm barely GO literate and not much of a programmer to boot.

### Panic Output

Don't have the exact error available right now, but basically 403 unauthorized to perform REST API Tag operation.

### Expected Behavior

Rest keepalives keep the REST session alive preventing a 403 when checking tags at end of a clone or other operation relying on the REST API that could occur after a period of time greater that the Vsphere REST timeout

### Actual Behavior

Timeouts for the CIS api are sent but not the REST api.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

tenthirtyam:
Resolved in #1301.

Marking this issue as closed. If this issue is reappears with the latest version of the provider, please create a new issue linking back to this one for added context.

**Ryan Johnson**
Staff II Solutions Architect
Cloud Infrastructure Business Group, VMware

happly:
Thank you for closing; I had forgotten to circle back after my PR was accepted!

PR #1301, opened 2021-01-05
Manual start of REST keepalive w/ cached session

happly:
### Description

#1300  contains most of the background for the issue this PR is resolving.

REST API session is not kept alive. I enabled the rest keepalive functionality from govmomi and tested this patched version in our environment and saw that it sends keepalives and successfully keeps the session open while Terraform is running. This fixed the issue I was having with low bandwidth causing template clones to fail when the provider checks the tags for the cloned VM using the Rest API. This probably also could affect Content Library resources as those looks to use the REST API as well.

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added? No
- [ ] Have you run the acceptance tests on this branch? No

Here is gist with pre/post patch logs for rest session:
https://gist.github.com/happly/2ebb3cd0cb0058ae8b0f9c498e263961

I will have to give this a shot tomorrow and edit with the results.

```

...
```

### Release Note

```release-note
NONE
...
```
### References

#1300  describes the issue in more detail and has a Gist of the error pre and post patch.

I reviewed the govmomi github project to understand how they were testing the REST API functionality and inserted that into this PR where relevant in the terraform-vpshere-provider context: https://github.com/vmware/govmomi/blob/87ef7a087817d7e6cddad4d52f6c9d0c7503a6a4/session/keepalive/handler_test.go#L112

### DIFF
--- a/vsphere/config.go
+++ b/vsphere/config.go
@@ -20,6 +20,7 @@ import (
 	"github.com/vmware/govmomi/pbm"
 	"github.com/vmware/govmomi/session"
 	"github.com/vmware/govmomi/session/cache"
+	"github.com/vmware/govmomi/session/keepalive"
 	"github.com/vmware/govmomi/vapi/tags"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/debug"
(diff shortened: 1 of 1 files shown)


koikonom:
Thanks for submitting this PR! LGTM!

PR #1311, opened 2021-01-13
Sorting permissions object on username/group name before storing.

### LINKED ISSUE #1293: vsphere_entity_permissions changed after every plan, even when specified in alphabetical order
### Terraform Version
Terraform v0.13.5
### vSphere Provider Version
registry.terraform.io/hashicorp/vsphere v1.24.3
vCenter: 6.5/6.7

### Affected Resource(s)

- `vsphere_entity_permissions`

permissions.tf:

 

> resource "vsphere_entity_permissions" cluster1 {
>    entity_id = data.module.cluster1_cluster.cluster_id
>    entity_type = "HostSystem" 
>  permissions {
>      user_or_group = "VSPHERE.LOCAL\\auser"
>      propagate = true
>      is_group = false
>      role_id = data.vsphere_role.administrator.id
>    }
>    permissions {
>      user_or_group = "VSPHERE.LOCAL\\buser"
>      propagate = true
>      is_group = false
>      role_id = data.vsphere_role.readonly.id
>    }

### Expected Behavior

Permissions are changed once. Further applies should not change the entity. Users/Groups have been placed in alphabetical order as requested.

### Actual Behavior
After every run, the following occurs. It appears the results pulled back from VC are not alphabetical so therefore terraform changes the permissions after every run. With large numbers of permissions per entity this becomes unmanageable

Terraform will perform the following actions:
 vsphere_entity_permissions.cluster1 will be updated in-place

 ```
 _~ resource "vsphere_entity_permissions" "cluster1" {
           entity_id   = "datacenter-969"
        entity_type = "HostSystem"
        id          = "datacenter-969"

      ~ permissions {
            is_group      = false
            propagate     = true
          ~ role_id       = "-2" -> "-1"
          ~ user_or_group = "**VSPHERE.LOCAL\\buser" -> "VSPHERE.LOCAL\\auser"**
        }
      ~ permissions {
            is_group      = false
            propagate     = true
          ~ role_id       = "-1" -> "-2"
          ~ user_or_group = "**VSPHERE.LOCAL\\auser" -> "VSPHERE.LOCAL\\buser"**
        }
    }_

```

### Steps to Reproduce

Create 2 users auser and buser. Apply to a cluster/folder etc using vsphere_entity_permissions. Re-run the plan and observe the plan being changed on each run. 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### PR DESCRIPTION
**Pull request by sumitAgrawal007**
### Description
The permissions objects received from server are not necessarily sorted on username/groupname. We are asking the users to keep it sorted in the TF file and this PR sorts the received objects alphabetically so that terraform doesn't show a diff while refreshing. 
### Acceptance tests

### Release Note

```release-note

- resource/vsphere_entity_permissions: Sorting permission objects on username/groupname before storing. (#1311)

```
### References
Closes #1293

### DIFF
--- a/vsphere/resource_vsphere_entity_permissions.go
+++ b/vsphere/resource_vsphere_entity_permissions.go
@@ -27,11 +27,12 @@ func resourceVsphereEntityPermissions() *schema.Resource {
 			Description: "The entity managed object type.",
 		},
 		"permissions": {
-			Type:        schema.TypeList,
-			Required:    true,
-			MinItems:    1,
-			Description: "Permissions to be given to the entity.",
-			Elem:        &schema.Resource{Schema: administrationroles.VspherePermissionSchema()},
+			Type:             schema.TypeList,
+			Required:         true,
+			MinItems:         1,
+			Description:      "Permissions to be given to the entity.",
+			Elem:             &schema.Resource{Schema: administrationroles.VspherePermissionSchema()},
+			DiffSuppressFunc: permissionsDiffSuppressFunc,
 		},
 	}
 
@@ -226,4 +227,30 @@ func resourceVSphereEntityPermissionsCustomizeDiff(d *schema.ResourceDiff, meta 
 		}
 	}
 	return nil
+}
+
+func permissionsDiffSuppressFunc(k, old, new string, d *schema.ResourceData) bool {
+	oldPermissions, newPermissions := d.GetChange("permissions")
+	oldPermissionsArr := oldPermissions.([]interface{})
+	newPermissionsArr := newPermissions.([]interface{})
+	if len(oldPermissionsArr) != len(newPermissionsArr) {
+		return false
+	}
+	sort.Slice(newPermissionsArr[:], func(i, j int) bool {
+		return strings.ToLower(newPermissionsArr[i].(map[string]interface{})["user_or_group"].(string)) <
+			strings.ToLower(newPermissionsArr[j].(map[string]interface{})["user_or_group"].(string))
+	})
+	for i := 0; i < len(oldPermissionsArr); i++ {
+		if strings.ToLower(oldPermissionsArr[i].(map[string]interface{})["user_or_group"].(string)) !=
+			strings.ToLower(newPermissionsArr[i].(map[string]interface{})["user_or_group"].(string)) ||
+			oldPermissionsArr[i].(map[string]interface{})["role_id"].(string) !=
+				newPermissionsArr[i].(map[string]interface{})["role_id"].(string) ||
+			oldPermissionsArr[i].(map[string]interface{})["is_group"].(bool) !=
+				newPermissionsArr[i].(map[string]interface{})["is_group"].(bool) ||
+			oldPermissionsArr[i].(map[string]interface{})["propagate"].(bool) !=
+				newPermissionsArr[i].(map[string]interface{})["propagate"].(bool) {
+			return false
+		}
+	}
+	return true
 }
--- a/vsphere/resource_vsphere_entity_permissions.go
+++ b/vsphere/resource_vsphere_entity_permissions.go
@@ -9,6 +9,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	"log"
+	"sort"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


koikonom:
Hey @sumitAgrawal007, I am trying to understand this change and I think I need some additional context. 

If I understand correctly the only thing this PR does it to sort the list of roles before setting it in the Read() function. Is there any actual effect on vSphere by this change?

What I'm concerned is that if users ignored the advice and did not sort their roles alphabetically, then once they run this version of the provider their plan will reports that changes are required, which may or may  not be true.

We could use a diff suppress func where we sort the list in the config and the list we get from vsphere and if they're equal then we just don't report back as a difference but, again, I need your help to understand the consequences of such a thing.

sumitAgrawal007:
No, there is no affect on vsphere by this change.

The reason of this change is that even if the user sorts his persmission in the configuration file, still the vsphere server returns them in unsorted/arbit way, and terraform plan reports that there are going to be changes. 
Earlier I had assumed that the server returns them in sorted way, which is true in many cases, but there was a bug raised, and he showed me thats not always the case.

So the best way to handle these issues is to suggest the user to sort them in .tf file and sort the server response as well.
We can use a diff surpress function to surpress diff if there is no change, but that wont show the diff correctly if there is a change. The unsorted permissions lists diff will be very hapazard.

So I am also going to add a diff surpress function here, to make it complete.

vsphere/resource_vsphere_entity_permissions.go
```diff
@@ -221,3 +228,29 @@ func resourceVSphereEntityPermissionsCustomizeDiff(d *schema.ResourceDiff, meta
...
+				newPermissionsArr[i].(map[string]interface{})["role_id"].(string) ||
+			oldPermissionsArr[i].(map[string]interface{})["is_group"].(bool) !=
+				newPermissionsArr[i].(map[string]interface{})["is_group"].(bool) ||
+			oldPermissionsArr[i].(map[string]interface{})["propagate"].(bool) !=
+				newPermissionsArr[i].(map[string]interface{})["propagate"].(bool) {
+			return false
+		}
+	}
+	return true
+}
```
koikonom, line 256: Thanks for adding this function. I think there is one more case we need to account for, let me know if this makes sense to you:

* User on the current version of the provider creates permissions but does not order them in their config.
* Once they upgrade we will sort the object and store it in the state.
* Next time a plan runs the diff suppress func will be evaluated. The "old" value is the unsorted one, and the "new" value is the sorted. This will produce a diff and ask the user to apply again.

My suggestion would be to sort both lists (old and new) and then use reflect.DeelEqual (https://golang.org/pkg/reflect/#DeepEqual) to decide if the two lists are equal or not.

What do you think?

koikonom (changes requested):

vsphere/resource_vsphere_entity_permissions.go
sumitAgrawal007: makes sense!! I will update the code.

vsphere/resource_vsphere_entity_permissions.go
```diff
@@ -119,6 +122,11 @@ func resourceEntityPermissionsRead(d *schema.ResourceData, meta interface{}) err
 		permissionObj["role_id"] = strconv.Itoa(int(permission.RoleId))
 		permissionObjs = append(permissionObjs, permissionObj)
 	}
+
+	sort.Slice(permissionObjs[:], func(i, j int) bool {
```
appilon, line 126: I noticed this pattern throughout, you are reslicing `slice[:]` and it's unclear to me the benefit (granted I'm not a slice ninja). This [go playground](https://play.golang.org/p/2BeLUXMWqts) demonstrates there is no difference. Curious on the reason.

appilon (approved):
Had a quick question about the slice trick I noticed? Other than that LGTM!

vsphere/resource_vsphere_entity_permissions.go
sumitAgrawal007: permissionObjs[:] It is converting array into slice.

sumitAgrawal007: Pardon me, I thought that was needed to convert array into slice, but turns out permissionObjs is already a slice.

PR #1310, opened 2021-01-12
Adding deployment option while deploying ovf/ova

sumitAgrawal007:
### Description

This allows users to specify any deployment option while deploying an OVA/OVF.
### Acceptance tests
- [x] Have you added an acceptance test for the functionality being added? Yes
- [ ] Have you run the acceptance tests on this branch? No

I have tested the code with NSX-T manager OVA.

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-vsphere/blob/master/CHANGELOG.md):

```release-note
resource/virtual_machine: Adding deployment option for OVA/OVF deployment (#1209)

```
### References
https://github.com/hashicorp/terraform-provider-vsphere/issues/1209

PR #1199, opened 2020-09-02
Allow vApp userConfigurable=false properties to be configured

adarobin:
### Description

This adds a new option under `ovf_deploy` named `enable_hidden_properties`.  This allows vApp properties that are marked with `userConfigurable=false` to be set. The most obvious use case for this functionality is directly deploying the vCenter Server Appliance OVA without using the installer.  The setting defaults to `false` which preserves the existing behavior.

The name is based off the name for an advanced option to `ovftool`.

```
$ ovftool --help debug | grep -A3 enableHidden 
  --X:enableHiddenProperties          : Enable source properties that are 
                                        marked as ovf:userConfigurable=false, 
                                        OVF Tool will set the value to true. 
                                        The default is false
```

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?
- [ ] Have you run the acceptance tests on this branch? (If so, please include the test log in a gist)

### References

* #394 
* https://www.virtuallyghetto.com/2016/10/how-to-deploy-the-vcenter-server-appliance-vcsa-6-5-running-on-vmware-fusion-workstation.html

PR #1339, opened 2021-02-11
OVF deployment options support 

koikonom:
The provider expects to be the only entity altering the settings of a
    virtual machine, which is not the case when deploying via OVF.

    The easy way out of this would be to make all attributes `Computed`
    which would make terraform to blidnly accept all changes into the state.
    Unfortunately this would also mean a loss of drift detection.

    The middle ground is to create a new datasource that extracts all the
    settings from the OVF template and expose it as outputs, so we can then
    pass it as attribute values to the `vsphere_virtual_machine` resource.

    The ideal solution would be to create a whole new resource to deal
    exclusively with OVF-built virtual machines but it would be a breaking
    change since users are already using the ovf subresource in
    `vsphere_virtual_machine`. This could potentially go into a v2.0 version
    of the provider (cc: @redeux).

    This commit creates the `vsphere_ovf_vm_template` resource and re-arranges
    some of the OVF code so it can be used from both `vsphere_virtual_machine`
    resource and the datasource.

### DIFF
--- a/vsphere/data_source_vsphere_ovf_vm_template.go
+++ b/vsphere/data_source_vsphere_ovf_vm_template.go
@@ -2,6 +2,7 @@ package vsphere
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/ovfdeploy"
 
@@ -81,6 +82,26 @@ func dataSourceVSphereOvfVMTemplate() *schema.Resource {
 			Type:        schema.TypeString,
 			Computed:    true,
 			Description: "The firmware interface to use on the virtual machine. Can be one of bios or EFI.",
+		},
+		"sata_controller_count": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "The number of SATA controllers that Terraform manages on this virtual machine. This directly affects the amount of disks you can add to the virtual machine and the maximum disk unit number. Note that lowering this value does not remove controllers.",
+		},
+		"ide_controller_count": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "The number of IDE controllers that Terraform manages on this virtual machine. This directly affects the amount of disks you can add to the virtual machine and the maximum disk unit number. Note that lowering this value does not remove controllers.",
+		},
+		"scsi_controller_count": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "The number of SCSI controllers that Terraform manages on this virtual machine. This directly affects the amount of disks you can add to the virtual machine and the maximum disk unit number. Note that lowering this value does not remove controllers.",
+		},
+		"scsi_type": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "The type of SCSI bus this virtual machine will have. Can be one of lsilogic, lsilogic-sas or pvscsi.",
 		},
 	}
 	s := map[string]*schema.Schema{
@@ -166,6 +187,47 @@ func dataSourceVSphereOvfVMTemplateRead(d *schema.ResourceData, meta interface{}
 	_ = d.Set("guest_id", vmConfigSpec.GuestId)
 	_ = d.Set("alternate_guest_name", vmConfigSpec.AlternateGuestName)
 	_ = d.Set("firmware", vmConfigSpec.Firmware)
+
+	controllers := map[string]int{}
+	var scsiType string
+
+	for _, dvc := range vmConfigSpec.DeviceChange {
+		dvcSpec := dvc.GetVirtualDeviceConfigSpec()
+
+		switch reflect.TypeOf(dvcSpec.Device) {
+		case reflect.TypeOf(&types.VirtualLsiLogicController{}):
+			if scsiType == "" {
+				scsiType = "lsilogic"
+			} else if scsiType != "lsilogic" {
+				return fmt.Errorf("multiple scsi controller types are not supported (found %s and %s)", scsiType, "lsilogic")
+			}
+			controllers["scsi"] = controllers["scsi"] + 1
+		case reflect.TypeOf(&types.VirtualLsiLogicSASController{}):
+			if scsiType == "" {
+				scsiType = "lsilogic-sas"
+			} else if scsiType != "lsilogic-sas" {
+				return fmt.Errorf("multiple scsi controller types are not supported (found %s and %s)", scsiType, "lsilogic-sas")
+			}
+			controllers["scsi"] = controllers["scsi"] + 1
+		case reflect.TypeOf(&types.ParaVirtualSCSIController{}):
+			if scsiType == "" {
+				scsiType = "pvscsi"
+			} else if scsiType != "pvscsi" {
+				return fmt.Errorf("multiple scsi controller types are not supported (found %s and %s)", scsiType, "pvsci")
+			}
+			controllers["scsi"] = controllers["scsi"] + 1
+		case reflect.TypeOf(&types.VirtualSATAController{}):
+			controllers["sata"] = controllers["sata"] + 1
+		case reflect.TypeOf(&types.VirtualIDEController{}):
+			controllers["ide"] = controllers["ide"] + 1
+		}
+	}
+
+	_ = d.Set("scsi_type", scsiType)
+	_ = d.Set("scsi_controller_count", controllers["scsi"])
+	_ = d.Set("sata_controller_count", controllers["sata"])
+	_ = d.Set("ide_controller_count", controllers["ide"])
+
 	d.SetId(d.Get("name").(string))
 
 	return nil
--- a/vsphere/data_source_vsphere_ovf_vm_template.go
+++ b/vsphere/data_source_vsphere_ovf_vm_template.go
@@ -103,7 +103,6 @@ func dataSourceVSphereOvfVMTemplate() *schema.Resource {
 		"datastore_id": {
 			Type:        schema.TypeString,
 			Optional:    true,
-			Computed:    true,
 			Description: "The ID of the virtual machine's datastore. The virtual machine configuration is placed here, along with any virtual disks that are created without datastores.",
 		},
 		"folder": {
--- a/vsphere/data_source_vsphere_ovf_vm_template.go
+++ b/vsphere/data_source_vsphere_ovf_vm_template.go
@@ -1,0 +1,173 @@
+package vsphere
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/ovfdeploy"
+
+	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/folder"
+
+	"github.com/vmware/govmomi/vim25/types"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/structure"
+	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/vmworkflow"
+)
+
+func dataSourceVSphereOvfVMTemplate() *schema.Resource {
+
+	vmConfigSpecSchema := map[string]*schema.Schema{
+		"num_cpus": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "The number of virtual processors to assign to this virtual machine.",
+		},
+		"num_cores_per_socket": {
+			Type:        schema.TypeInt,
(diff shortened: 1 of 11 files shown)


koikonom:
This extends the work done by @sumitAgrawal007 in #1310 .

/cc @djpbessems @aareet @redeux

koikonom:
I am posting this as a draft PR for now to get the conversation started. 

There's still tests and documentation to be done. I may also have an idea about how to better handle network interfaces.

djpbessems:
Man, can't wait to test this :D

koikonom:
You can test it now if you're on v0.14.X. 

Clone the repo, run `make` and then add this to `~/.terraformrc`:

```
provider_installation {
    dev_overrides {
        "hashicorp/vsphere" = "/path/to/your/gopath/bin"
    }
    direct {}
}```

Just remember to remove the setting from terraformrc once you're done, otherwise terraform will keep using that provider instead of the one in the registry.

**Comment by sumitAgrawal007** [comment]
This looks like a great improvement, I was always worried about the fact that the resources were not controlled by terraform, but rather by vsphere. This change will handle that limitation. I will try this out later today.

**Comment by djpbessems** [comment]
> You can test it now if you're on v0.14.X.
> [...]

Well I had some time today to look into this, and set up an environment with go/make/etc.; then I started looking at the syntax of the terraform templates. Am I correct in understanding that this will not work with deploying from Content Libraries like below (I don't see where I would define `deployment_option`)?
```
resource "vsphere_virtual_machine" "<vmname>" {
  [...]
  clone {
    template_uuid        = vsphere_content_library_item.<libraryitemname>.id
  }

  vapp {
    properties = {
      "<keyname>"     = "<value>"
    }
  }
}
```

**Comment by koikonom** [comment]
Hi @djpbessems you are correct, this would not work with Content Libraries. It is designed to with with URLs (or local paths) to OVA/OVF files.

**Comment by ccater** [comment]
I was testing this and I noticed that the scsi_type is not correctly pulled from the ovf, can we make this a computed attribute somehow?

**Comment by koikonom** [comment]
Hi @ccater, good call. I will give this one a try and see what we can do!

**Comment by adarobin** [comment]
@koikonom Would it make more sense for this to be a resource and not a datasource?  If this were a resource, it would only need to be read in at creation time and then persisted? That way you also wouldn't be reliant on the content existing at a certain url or file path for all time.

**Comment by koikonom** [comment]
Hi @adarobin, yes it does make sense and we do plan to do it for version 2 of this provider. Doing it now would cause confusion between people using the new resource and people using the existing `ovf_deployment` method.

**Comment by adarobin** [comment]
@koikonom I was thinking of it more in terms of being a "resource" that is basically read in once and then persisted. The "resource" wouldn't do anything different from the datasource other than not require it to be read in for every plan. The `ovf_deployment` method would still be used to actually deploy the machine. It would be more akin to a resource named something like `vsphere_ovf_vm_template_info` than `vsphere_ovf_vm_template`.

In any case, I built this the other night and its working great for me with a VCSA. I'm excited for this to be merged.

**Comment by adarobin** [comment]
I just noticed some interesting behavior when deploying the NSX-T 3.1 Universal Appliance.

It looks like the template doesn't declare values for `firmware` or `swap_placement_policy`
```
Error: expected firmware to be one of [bios efi], got

  on ../terraform-vsphere-nested-vsphere-lab/modules/nsx-t/main.tf line 35, in resource "vsphere_virtual_machine" "nsxt_manager":
  35: resource "vsphere_virtual_machine" "nsxt_manager" {



Error: expected swap_placement_policy to be one of [inherit vmDirectory hostLocal], got

  on ../terraform-vsphere-nested-vsphere-lab/modules/nsx-t/main.tf line 35, in resource "vsphere_virtual_machine" "nsxt_manager":
  35: resource "vsphere_virtual_machine" "nsxt_manager" {
```

I was able to work around this by not trying to set `firmware` or `swap_placement_policy` on the resource using the datasource.

The other thing I noticed is that `scsi_type` isn't exposed from the data source so the controller is getting swapped from "lsilogic" to "pvscsi" after the deployment.

**Comment by adarobin** [comment]
Does this actually _deploy_ the virtual machine with the deployment option like #1215 did, or just read the deployment option specs into the data source object?

I noticed that there were changes made to the machine _after_ I deployed the NSX-T universal appliance.  I used the "small" deployment option where as the default is "medium". After deployment I can see a reconfigure task in vCenter that sets the CPU and Memory values from the values provided by "medium" to the values provided by the data source.

Unfortunately, this doesn't cover all of the options that can be set at deploy time, specifically virtual disks. The NSX-T appliance has the same disk size for all of its deployment options so it doesn't really matter in this case. However, all of the deployment options available in the VCSA ova have different disk sizes.

It also looks the OVA is setting resource allocation options which also need to be available from the data source.

Here is the description of the reconfigure task I see.  I have cleaned up the formatting a bit for clarity:
```
Reconfigured nsxt on redacted.esxi.example.com in Homelab.
Modified:
config.flags.enableLogging: true -> false;
config.hardware.numCPU: 6 -> 4;
config.hardware.memoryMB: 24576 -> 16384;
config.cpuAllocation.reservation: 6000 -> 0;
config.cpuAllocation.shares.shares: 6000 -> 4000;
config.memoryAllocation.reservation: 24576 -> 0;
config.memoryAllocation.shares.shares: 245760 -> 163840;
Added: config.extraConfig("sched.cpu.latencySensitivity"): (key = "sched.cpu.latencySensitivity", value = "normal");
Deleted:
--
```

**Comment by djpbessems** [comment]
I've done a few more tests (now without using content libraries), and I'm getting the following error:
```
Error: error while creating vapp properties config vApp property with userConfigurable=false specified in vapp.properties: [<list of properties>]
```
These properties however (for example: property `dhcpconfig.startip`), are part of the deployment-option that I specified (and *not* the default deployment-option); see the code fragments below:
```
<Envelope [...]>
  [...]
  <DeploymentOptionSection>
    <Info>Deployment Type</Info>
    <Configuration ovf:id="primary">
      <Label>Primary (redundant deployment)</Label>
      <Description>Initial Domain Controller with 'PDC Emulator'-role</Description>
    </Configuration>
    <Configuration ovf:id="secondary">
      <Label>Secondary (redundant deployment)</Label>
      <Description>Additional Domain Controller</Description>
    </Configuration>
    <Configuration ovf:id="standalone">
      <Label>Stand-alone (non-redundant deployment)</Label>
      <Description>Single Domain Controller</Description>
    </Configuration>
  </DeploymentOptionSection>
  <VirtualSystem ovf:id="[...]">
    [...]
    <ProductSection>
      [...]
      <Category>1) Operating System</Category>
      <Property ovf:configuration="primary secondary standalone" ovf:key="guestinfo.hostname" [...]>
        <Label>Hostname*</Label>
      </Property>
      [...]
      <Category>3) Active Directory Domain Services</Category>
      <Property ovf:configuration="primary standalone" ovf:key="addsconfig.ntpserver" [...]>
        <Label>NTP Server*</Label>
      [...]
      <Category>4) DHCP default scope</Category>
      <Property ovf:configuration="secondary" ovf:key="dhcpconfig.startip" [...]>
        <Label>DNS server*</Label>
      </Property>
      [...]
      </Property>
    </ProductSection>
  </VirtualSystem>
</Envelope>
```
I used the following terraform template:
```
data "vsphere_ovf_vm_template" "ovf-dc" {
  name                   = "ovf-dc"
  resource_pool_id       = data.vsphere_resource_pool.pool.id
  datastore_id           = data.vsphere_datastore.vmdatastore.id
  host_system_id         = data.vsphere_host.host.id

  remote_ovf_url         = "https://<fqdn>/appliance.ova"
  deployment_option      = "standalone"

  ovf_network_map = {
    "LAN"                = data.vsphere_network.network.id
  }
}

resource "vsphere_virtual_machine" "DC01" {
  datacenter_id          = data.vsphere_datacenter.dc.id

  name                   = "DC01"
  folder                 = var.vsphere_vmfolder

  num_cpus               = data.vsphere_ovf_vm_template.ovf-dc.num_cpus
  memory                 = data.vsphere_ovf_vm_template.ovf-dc.memory

  guest_id               = data.vsphere_ovf_vm_template.ovf-dc.guest_id
  resource_pool_id       = data.vsphere_ovf_vm_template.ovf-dc.resource_pool_id
  datastore_id           = data.vsphere_ovf_vm_template.ovf-dc.datastore_id
  host_system_id         = data.vsphere_ovf_vm_template.ovf-dc.host_system_id

  ovf_deploy {
    ovf_network_map      = data.vsphere_ovf_vm_template.ovf-dc.ovf_network_map
    remote_ovf_url       = data.vsphere_ovf_vm_template.ovf-dc.remote_ovf_url
  }

  vapp {
    properties = {
      "guestinfo.hostname"         = "DC01"
      [...]
      "addsconfig.ntpserver"       = "0.pool.ntp.org,1.pool.ntp.org,2.pool.ntp.org"
      [...]
      "dhcpconfig.startip"         = "10.0.0.50"
      [...]
    }
  }
}
```

Either I'm misunderstanding where the vapp.properties need to be declared, or this resource does not actually select the deployment_option when deploying the .ova (which is why the listed properties are implicitly set to `userConfigurable: false`).

adarobin:
@djpbessems I think the piece that both you and I are missing is that you need to set ` deployment_option = data.vsphere_ovf_vm_template.ovf.deployment_option` in the `ovf_deploy` section of the resouce.

As far as the `userConfigurable: false` properties, I have #1199 still open and it cherry-picks cleanly to this branch if you want to try it out in conjunction with this.

adarobin:
I threw together a couple of modules using this new functionality. Thought it might be useful to someone for development/testing.

The ESXi one will download the OVA directly from VMware by default (or you can override it to use a local copy to save bandwidth).  The VCSA and NSX-T Manager ones require you to first get the ova file from VMware.  I used the latest versions available from VMUG Advantage in my testing.

https://github.com/adarobin/terraform-vsphere-nested-esxi
https://github.com/adarobin/terraform-vsphere-vcsa
https://github.com/adarobin/terraform-vsphere-nsx-t-manager

The only outstanding issues I see are:

- Resource allocation options not being presented from the data source.
- `firmware` and `swap_placement_policy` are not always being set.  Perhaps it would make sense for those to be set to the default value used in the virtual machine resource if they aren't set?
- I'm seeing some weirdness with the `enable_logging` option. Seems like the value gets forced to false no matter what I do.

djpbessems:
> 
> 
> @djpbessems I think the piece that both you and I are missing is that you need to set ` deployment_option = data.vsphere_ovf_vm_template.ovf.deployment_option` in the `ovf_deploy` section of the resouce.  

That was indeed all that was needed to get my .ova deployed correctly. **YAY!**  
  (still would *love* to have this be compatible with Content Libraries though, the deploy takes so loooong without :D)
> 
> As far as the `userConfigurable: false` properties, I have #1199 still open and it cherry-picks cleanly to this branch if you want to try it out in conjunction with this.  

That's not relevant for me, the properties that either are or are not part of a configuration are interpreted by vSphere as `userConfigurable` set to `true` or `false` respectively dynamically.

djpbessems:
> 
> 
> I threw together a couple of modules using this new functionality. Thought it might be useful to someone for development/testing.
> 

Since we're sharing; my usecase for `deployment_option` are (work-in-progress) appliances of ActiveDirectory domain controllers, certificate authorities, etc that can be deployed fully unattended:
- https://code.spamasaurus.com/djpbessems/Packer.Images
- https://code.spamasaurus.com/djpbessems/Terraform.VirtualMachines

appilon (approved):
Amazing!

djpbessems:
So, is this going to be merged soon? And will there be a new release of the provider?

I really don't want to sound ungrateful and impatient, but the last release was December 2020, and one other issue that has been haunting me (https://github.com/hashicorp/terraform-provider-vsphere/issues/1242) was solved and merged in January already.

I'm just very very eager to continue working with Terraform for our deployments, and want these fixes soon :)

koikonom:
Yes, this is going to be merged and released this week. Thank you for your help on testing this PR!

djpbessems:
Is there something that the community/users of this provider can do to make a release happen? I'm willing to learn stuff if it means I can help make stuff move again!

koikonom:
Hi @djpbessems, this PR is part of the latest vsphere provider release (https://github.com/hashicorp/terraform-provider-vsphere/blob/master/CHANGELOG.md#1250-march-17-2021).

djpbessems:
Ohhhh, very sorry for my comment then.

It's just that I am not seeing the new release when looking at the repo's landing page:
image

koikonom:
Hmm you're right, I'll fix that. 

In the mean time you can consult the registry page (https://registry.terraform.io/providers/hashicorp/vsphere/latest) since that's what terraform also does when it looks for the latest version.

Issue #1190, opened 2020-08-31
Add support for a flag to force replace of vsphere_virtual_machine


gmintoco:
### Description

I am using cloud-init with vsphere and terraform to automatically deploy VMs and bootstrap them. The issue I am facing is that when a modification is made to the cloud init metadata/config after deployment the VM is not replaced and the changes will not take effect (which makes sense as there is no way the provider knows that those 'guestinfo' parameters require a replacement)

You can see below an example configuration for passing data from 'extra_config' into the VM. I realise that this might not be the best way to do this - possibly puppet might work better, but in our use case it doesn't make sense to have the additional infrastructure for something like puppet.

My idea would be to have a boolean flag like 'replace_on_extra_config_modify: true/false' that would force a replacement if anything under extra_config changed. I am open to any other thoughts and suggestions. I think this might also be relevant for general OVF configuration (as only modification to the clone block will force replacement) along with vapp configuration through the vapp block possibly.

### Potential Terraform Configuration

```hcl
resource vsphere_virtual_machine "gitlab_runner" {
  name             = "gitlab-runner-${local.init_hash}.${local.domain_name}"
  resource_pool_id = data.vsphere_compute_cluster.this.resource_pool_id
  datastore_id     = data.vsphere_datastore.this.id

  enable_disk_uuid = true

  num_cpus = 2
  memory   = 4096
  guest_id = data.vsphere_virtual_machine.template.guest_id

  cdrom {
    client_device = true
  }

  network_interface {
    network_id = data.vsphere_network.management.id
  }

  network_interface {
    network_id = data.vsphere_network.aux.id
  }

  wait_for_guest_net_timeout = 0

  disk {
    label            = "disk0"
    size             = data.vsphere_virtual_machine.template.disks.0.size
    eagerly_scrub    = data.vsphere_virtual_machine.template.disks.0.eagerly_scrub
    thin_provisioned = data.vsphere_virtual_machine.template.disks.0.thin_provisioned
  }

  clone {
    template_uuid = data.vsphere_virtual_machine.template.id
  }

  extra_config = {
    "guestinfo.metadata" = base64encode(templatefile("./files/metadata.yaml.template", {
      management_ip      = var.Management_IP
      management_gateway = var.Management_Gateway
      management_subnet  = "24"
      aux_ip         = var.Aux_IP
      aux_gateway    = var.Aux_IP
      aux_subnet     = "24"
[...]

PR #1360, opened 2021-03-09
#1190 - Add replacement trigger to virtual machine resource

fredrikhgrelland:
### Description

This PR implements a new key `replace_trigger` to `vsphere_virtual_machine`. This adresses feature request #1190 which proposes a way to force the replacement of a virtual machine resource. I am linking in this issue as it matches our needs.

Why?
Currently there is no way of forcing a replace of a virtual machine created with cloud_init. In order to force replacements of a cloned vm, you must change a clone customization parameter and this does not work (and does not make sense) when using cloud_init.

Working implementation: https://registry.terraform.io/providers/fredrikhgrelland/vsphere/latest/docs/resources/virtual_machine?pollNotifications=true#replace_trigger

### Acceptance tests

I have not run acceptance tests using go tests, but have verified the functionality manually.
All we need to check is if changes to this key forces a replace, and it does as the schema specifically asks for it. 
This is core terraform functionality exposed.

```

...
```

### Release Note

```release-note
...
```


fredrikhgrelland:
@koikonom Do you accept PRs? This is a fundamental need for phoenix-builds.

redeux:
Hi @fredrikhgrelland.  Thank you for the PR.  

Terraform has a native capability to destroy and recreate a resource on the next apply using the [taint](https://www.terraform.io/docs/cli/commands/taint.html) command. Before we consider adding this kind of functionality to the vSphere provider, can you tell me whether you've tried to use taint and if so, why it doesn't work for you?  Thanks.

fredrikhgrelland:
@redeux Thank you for coming back to me on this issue.

Yes, we are aware that you may taint a resource, but tainting should not be a necessary step in order to make changes to a resource. The problem here arise from the fact that any changes to cloud-init will never trigger a meaningful change in the vm. Currently this provider will write the cloud-init file, restart the vm and consider it to be changed.
Changing parameters in cloud-init, eg. network definitions/hostname etc., requires a replace of the resource in order to be effectuated.
The wanted behaviour mirrors that of the clone-stanza, that will replace a resource if you change networking parameters.

The use case here is that we want the resource to be _replaced_ when we need it to.
- Installations is done in packer and put into a template.
- Configurations are done in cloud-init while instantiating the template. Including introduction of secrets.

Changes to the template will replace the resource, as this is considered a change.
Changes to networking, renewal or changing secrets/roles etc. will not replace the resource, as this is not considered a change.

This patch will allow for phoenix-builds and immutable infrastructure on vsphere along with automated procedures.

Does this clarify the problem sufficiently?

koikonom:
Hi @fredrikhgrelland 

Terraform 0.15.2 introduced a new `replace` parameter to plan and apply commands. Please take a look at https://www.terraform.io/docs/cli/commands/plan.html#replace-address, I think it will help in your case.

fredrikhgrelland:
Thanks @koikonom for pointing this out. It will surely be easier for ad-hoc replacements along with -target, but unfortunately it does not meet our requirements. With this switch we must create custom wrapper scripts and it would _always_ replace the resources. This trigger is still needed in order to automate the replacement based on relevant changes, while keeping it in place when there is no need to replace it.

koikonom (approved):

PR #1361, opened 2021-03-09
Fix compare for VM Config spec change

k-chu:
This PR fixes an issue reported in https://github.com/hashicorp/terraform-provider-vsphere/issues/1266. 

When a Virtual Machine is imported, and then a plan/apply is run without any changes to the template, the VM is started after apply. This started happening in **v1.17.4** and is still present in the latest provider.  This appears to be a side effect from a different fix https://github.com/hashicorp/terraform-provider-vsphere/pull/1055


ifwatts:
@hashicorp-cla is there any timeline for this to be merged?  There are customers experiencing this issue today and we want to resolve that as soon as possible.

ifwatts:
@koikonom is there some process needed to be done to have this merged?

koikonom (approved):
LGTM thanks!

PR #1376, opened 2021-03-30
Change vnic netstack property default value to 'defaultTcpipStack' as the 'default' value does not exist

frank-m:
### Description

Change default value to 'defaultTcpipStack' as the 'default' value does not exist

The test uses the correct value to test, which is why this probably went unnoticed.

### Acceptance tests

The current test already is valid, however it overwrites the default value and as such this has gone unnoticed.

No, tests need a test vsphere cluster which I currently do not have available. However, this makes the code work as specified in the documentation: 

https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/vnic

netstack - (Optional) TCP/IP stack setting for this interface. Possible values are 'defaultTcpipStack', 'vmotion', 'vSphereProvisioning'. Changing this will force the creation of a new interface since it's not possible to change the stack once it gets created. (Default: defaultTcpipStack)

### Release Note

```release-note
NONE
```

### DIFF
--- a/vsphere/resource_vsphere_vnic.go
+++ b/vsphere/resource_vsphere_vnic.go
@@ -3,15 +3,16 @@ package vsphere
 import (
 	"context"
 	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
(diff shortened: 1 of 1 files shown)


koikonom (approved):
LGTM. Thanks for fixing this!

PR #1381, opened 2021-04-08
Acceptance test simplification and related changes

koikonom:
This started as an effort to run tests against vsphere 7.0, but ended up fixing a few outstanding bugs and making the acceptance tests a bit simpler by reusing some resources instead of building new ones each time.

There are some outstanding issues with the tests, but these will addressed in separate PRs.

### DIFF
--- a/scripts/packet/main.tf.phase1
+++ b/scripts/packet/main.tf.phase1
@@ -30,6 +30,28 @@ variable "PACKET_PROJECT" {}
 
 variable "PRIV_KEY" {}
 
+variable "VCSA_DEPLOY_PATH" {}
+
+variable "ESXI_VERSION" {
+  default = "vmware_esxi_6_7"
+}
+
+variable "PACKET_FACILITY" {
+  default = "sjc1"
+}
+
+variable ESXI_PLAN {
+  default = "c3.medium.x86"
+}
+
+variable STORAGE_PLAN {
+  default = "c1.small.x86"
+}
+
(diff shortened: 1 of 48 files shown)


scripts/packet/main.tf.phase1
```diff
@@ -137,10 +179,21 @@ resource "local_file" "vcsa_template" {
   })
   filename = "./tmp/vcsa.json"
   provisioner "local-exec" {
-    command = "sleep 290; echo five more; sleep 290; TERM=xterm-256color ./tmp/mnt/vcsa-cli-installer/lin64/vcsa-deploy install --accept-eula --acknowledge-ceip --no-ssl-certificate-verification --verbose --skip-ovftool-verification ./tmp/vcsa.json"
+    command = "sleep 290; echo five more; sleep 290; TERM=xterm-256color ${var.VCSA_DEPLOY_PATH} install --accept-eula --acknowledge-ceip --no-ssl-certificate-verification --verbose --skip-ovftool-verification $(pwd)/tmp/vcsa.json"
```
appilon, line 182: I recall the sleep 290; echo five more as unnecessary when i set things up... but my memory is hazy

appilon (approved):
👏

scripts/packet/main.tf.phase1
koikonom: sometimes the server is not immediately available, which is why this long sleep was added.

PR #1431, opened 2021-06-23
Add new option for syncing time on vSphere 7 U1 and above

waquidvp:
### Description

Adds the ability to get a new option for syncing time which was introduced in vSphere 7 U1. IMO the name of the field is a bit miss leading, so if there any suggestions for a better name, let me know.

This new options `sync_time_with_host_allowed` determines if VMware tools should periodically sync time with the host. Before vSphere 7 U1, there was only one option `sync_time_with_host` which by default turned on periodic syncs. The vSphere 7 U1 and above, this option can be separately set. The following KB article explains it much better https://kb.vmware.com/s/article/1189.

Here is the options in the UI before vSphere 7 U1:
image

And here is the same options in the UI after vSphere 7 U1:
image

While testing, I have noticed that on vSphere 7 U1, just setting `sync_time_with_host` isn't enough for any of the tick boxes to activate, you have to have both `sync_time_with_host` and `sync_time_with_host_allowed` to be set to `true`, in which case both tick boxes will be activated. With that in mind, we could just use the value of `sync_time_with_host` and pass it onto the `SyncTimeWithHostAllowed` without introducing a new field in the Terraform module.

### Acceptance tests

### Release Note

```release-note
`resource/virtual_machine` - Add `sync_time_with_host_allowed` field to control whether to time should be synced periodically with hosts on vSphere 7 U1 and above.
```
### References

Here is a screenshot of the docs relating to the fields:
image

### DIFF
--- a/vsphere/virtual_machine_config_structure.go
+++ b/vsphere/virtual_machine_config_structure.go
@@ -145,6 +145,11 @@ func schemaVirtualMachineConfigSpec() map[string]*schema.Schema {
 			Optional:    true,
 			Description: "Enable guest clock synchronization with the host. Requires VMware tools to be installed.",
 		},
+		"sync_time_with_host_allowed": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Description: "Indicates whether VMware tools is allowed to synchronize the guest clock with the host. On vSphere 7 U1 and above, this must be set to true for clock synchronization. Requires VMware tools to be installed.",
+		},
 		"run_tools_scripts_after_power_on": {
 			Type:        schema.TypeBool,
 			Optional:    true,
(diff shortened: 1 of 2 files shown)


koikonom:
Hi @waquidvp thanks for this PR! I have a quick question if you don't mind. Are there any cases where someone on vsphere 7.0.1 and later would choose to enable one option but not the other?

waquidvp:
I retested all the scenarios, here are the findings:

- Setting only `sync_time_with_host_allowed` ticks only the first box - So this setting controls the "Synchronize at startup and resume (recommended)" box
- Setting only `sync_time_with_host` doesn't do anything - I believe this controls the "Synchronize time periodically" tick box. The vSphere docs do mention that, setting this to true is useless unless `sync_time_with_host_allowed` is also set to true, so this behavior makes sense.
- Setting both `sync_time_with_host_allowed` and `sync_time_with_host` follows the expected behavior with both boxes ticked.

With that in mind, I think there is a use case for the ability to control these separately. Eg. someone wants to synchronize at startup but not periodically.

Also, maybe the following names make more sense with the above behavior in mind:

- `sync_time_with_host_allowed` should be `sync_time_with_host`
- `sync_time_with_host` should be `sync_time_with_host_periodically`

However that will probably break the API and users not yet on vSphere 7.0.1.

koikonom:
I think you're correct on the naming front.

As for breaking backward compatibility, you can set this only if a  version check succeeds, like this:
https://github.com/hashicorp/terraform-provider-vsphere/pull/1430/files#diff-0f666e3e0a18ec30c8b0ce30625fbaaa649f031053f41843159c3470d6baf51eR382-R383

waquidvp:
I have updated the names to reflect what we mentioned above. It's now using the new `version.AtLeast` function to check and behave differently on vSphere 7 U1 and above. Let me know if it makes sense now.

koikonom (changes requested):
Thanks for updating this. There is one issue to fix though, see suggested changes :)

koikonom:
Can you also update the docs? The same text you used for the `Description` field is fine to go in the docs too!

waquidvp:
Suggested changes applied and docs updated (sorry, I assumed the docs were somehow being generated from the code)

koikonom:
Newer providers use https://github.com/hashicorp/terraform-plugin-docs, but this provider hasn't migrated to that tool yet.

koikonom (approved):
Thank you very much for this! I will be merging this tomorrow and it will go out on the next release.

Issue #1205, opened 2020-09-08
`vsan_disk_group` in `r/vsphere_compute_cluster` should be `TypeSet` and not `TypeList`


adarobin:
### Terraform Version
0.13.1

### vSphere Provider Version
1.23.0

### Affected Resource(s)

- `vsphere_compute_cluster`

### Terraform Configuration Files

```hcl
resource "vsphere_compute_cluster" "cluster" {
  provider = vsphere.nested

  name          = var.nested_cluster_name
  datacenter_id = vsphere_datacenter.datacenter.moid

  host_system_ids = [ for k,v in var.esxi_hostname_ip_map : vsphere_host.nested_esxi[k].id ]

  drs_enabled                 = true
  drs_automation_level        = "fullyAutomated"
  ha_enabled                  = true
  ha_admission_control_policy = "disabled"
  force_evacuate_on_destroy   = true
  
  vsan_enabled = true

  dynamic "vsan_disk_group" {
    for_each = var.esxi_hostname_ip_map
    iterator = host

    content {
      cache   = var.vsan_cache_disks[host.key][0]
      storage = var.vsan_capacity_disks[host.key]
    }
  }
}
```

### Expected Behavior

`terraform apply` is successful.  No changes are needed after an initial apply.

### Actual Behavior

`terraform apply` is successful.  The vSAN is created, but future `terraform plan` runs show changes needed due to the order of elements in vsan_disk_group.

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.nested-esxi-lab-inner.vsphere_compute_cluster.cluster will be updated in-place
  ~ resource "vsphere_compute_cluster" "cluster" {
        custom_attributes                                     = {}
        ** snipped a bunch of lines here for clarity **
        vsan_enabled                                          = true

      ~ vsan_disk_group {
          ~ cache   = "naa.6000c29f402bc1b9b4bff84f5f4c5bc4" -> "naa.6000c2964873546c663d4fa6e64faa2e"
          ~ storage = [
              + "naa.6000c2906103371d4b758bc6557aa2e0",
              - "naa.6000c2929de009565ef1efdb3077ee9a",
            ]
        }
        vsan_disk_group {
            cache   = "naa.6000c29758b6ba406b8ebfb672056810"
            storage = [
                "naa.6000c29128e01ac02b774e94f953555a",
            ]
        }
[...]

PR #1432, opened 2021-06-23
`vsan_disk_group` should be of type `TypeSet`

waquidvp:
### Description

Change the type of `vsan_disk_group` from `typelist` to `typeset`. When it is a list, the order of it is random every time `terraform plan` is called and causes a change in the vSAN disk groups, resulting in disks being deleted and re-added into the disk group every time you run `terraform apply`. This issue has already been raised here: https://github.com/hashicorp/terraform-provider-vsphere/issues/1205.

### Acceptance tests

### Release Note

```release-note
`resource/compute_cluster` - `vsan_disk_group` is now a set and the order will be preserved between `terraform` runs
```
### References

* https://github.com/hashicorp/terraform-provider-vsphere/issues/1205.

### DIFF
--- a/vsphere/resource_vsphere_compute_cluster.go
+++ b/vsphere/resource_vsphere_compute_cluster.go
@@ -478,7 +478,7 @@ func resourceVSphereComputeCluster() *schema.Resource {
 				Description: "Whether the VSAN service is enabled for the cluster.",
 			},
 			"vsan_disk_group": {
-				Type:        schema.TypeList,
+				Type:        schema.TypeSet,
 				Optional:    true,
 				Computed:    true,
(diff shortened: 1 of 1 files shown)


waquidvp:
@koikonom Any chance you could have a look at this PR?

koikonom (approved):
Good catch!

appilon:
Unfortunately we consider this type change a breaking change, furthermore the change caused a crash as type assertions around the attribute need to be tweaked with `TypeSet` (it's `.(*schema.Set).List()` VS `.([]interface)`

We will be reverting this change and re-opening #1205 . Once we are ready for a `v3.0.0` we will re-merge this change.

PR #1451, opened 2021-07-24
Update content_library.html.markdown

nvibert:
### Description

The current wording for the description of the `on_demand` attribute is confusing. When set to true, "on_demand" will not download all library content immediately (it's the opposite).


appilon (approved):

tenthirtyam:
Other detailed updates to the same content are included in #1502.

Ryan

PR #1486, opened 2021-09-30
force reconfigure due to race condition

brenthc:
### Description

This change removes the conditional logic whereby if the datastore is known, that value is used. In some cases the previously obtained datastore value is incorrect due to an immediate vmotion event that occurs between the initial clone and the post update operations.

### Acceptance tests
This issue was observed in production and has been resistant to reproduction due to the numerous factors involved in inducing an immediate vmotion fast enough.

### Release Note

```release-note
Don't assume the datastore is the same between create and post updates
```


appilon (approved):
I ran some existing tests and it doesn't seem like this change breaks anything so we will give it a shot

PR #1522, opened 2021-11-29
Update `vsphere_virtual_machine` resource docs with reboot conditions

AliAllomani:
### Description

Updated the documentation for vsphere_virtual_machine resource with the conditions that causes the virtual machine to reboot on update.

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?  Not needed
- [ ] Have you run the acceptance tests on this branch? Not needed

### Release Note

```release-note
`resource/vsphere_virtual_machine`:  Updated the documentation with the conditions that causes the virtual machine to reboot on update
```
### References

Closes https://github.com/hashicorp/terraform-provider-vsphere/issues/1004

### DIFF
--- a/website/docs/r/virtual_machine.html.markdown
+++ b/website/docs/r/virtual_machine.html.markdown
@@ -1602,41 +1602,41 @@ management of such disks would render their configuration unstable.
 
 ## Virtual Machine Reboot
 
-The virtual machine will be rebooted if any of the following parameters changed:
+The virtual machine will be rebooted if any of the following parameters are changed:
 
-* `extra_config`
-* `guest_id`
 * `alternate_guest_name`
-* `hardware_version`
-* `pci_device_id`
-* `memory` -  When reducing the memory size, or when increasing the memory size and `memory_hot_add_enabled` is set to `false`
-* `memory_hot_add_enabled`
 * `cpu_hot_add_enabled`
 * `cpu_hot_remove_enabled`
-* `num_cores_per_socket`
-* `swap_placement_policy`
-* `firmware`
-* `nested_hv_enabled`
 * `cpu_performance_counters_enabled`
-* `efi_secure_boot_enabled`
-* `enable_disk_uuid`
-* `hv_mode`
-* `ept_rvi_mode`
-* `enable_logging`
-* `vbs_enabled`
-* `vvtd_enabled`
-* `run_tools_scripts_after_power_on`
-* `run_tools_scripts_after_resume`
-* `run_tools_scripts_before_guest_standby`
-* `run_tools_scripts_before_guest_shutdown`
-* `run_tools_scripts_before_guest_reboot`
 * `disk.controller_type`
 * `disk.unit_number`
 * `disk.disk_mode`
 * `disk.write_through`
 * `disk.disk_sharing`
+* `efi_secure_boot_enabled`
+* `ept_rvi_mode`
+* `enable_disk_uuid`
+* `enable_logging`
+* `extra_config`
+* `firmware`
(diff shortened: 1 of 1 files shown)


website/docs/r/virtual_machine.html.markdown
```diff
@@ -1600,6 +1600,44 @@ management of such disks would render their configuration unstable.
 
 [tf-vsphere-virtual-disk]: /docs/providers/vsphere/r/virtual_disk.html
 
+## Virtual Machine Reboot
+
+The virtual machine will be rebooted if any of the following parameters changed:
```
tenthirtyam, line 1605: "...are changed:"

website/docs/r/virtual_machine.html.markdown
```diff
@@ -1600,6 +1600,44 @@ management of such disks would render their configuration unstable.
...
+
+* `extra_config`
+* `guest_id`
+* `alternate_guest_name`
+* `hardware_version`
+* `pci_device_id`
+* `memory` -  When reducing the memory size, or when increasing the memory size and `memory_hot_add_enabled` is set to `false`
```
tenthirtyam, line 1612: * `memory` 
   When reducing or when increasing the memory size and when `memory_hot_add_enabled` is set to `false`.

website/docs/r/virtual_machine.html.markdown
```diff
@@ -1600,6 +1600,44 @@ management of such disks would render their configuration unstable.
...
+* `run_tools_scripts_after_resume`
+* `run_tools_scripts_before_guest_standby`
+* `run_tools_scripts_before_guest_shutdown`
+* `run_tools_scripts_before_guest_reboot`
+* `disk.controller_type`
+* `disk.unit_number`
+* `disk.disk_mode`
+* `disk.write_through`
+* `disk.disk_sharing`
+* `network_interface` - When deleting a network interface and VMware Tools is not running.
```
tenthirtyam, line 1638: * `network_interface`
    When deleting a network interface and VMware Tools is not running.
* `network_interface.adapter_type`
    When VMware Tools is not running.

website/docs/r/virtual_machine.html.markdown
```diff
@@ -1600,6 +1600,44 @@ management of such disks would render their configuration unstable.
 
 [tf-vsphere-virtual-disk]: /docs/providers/vsphere/r/virtual_disk.html
 
+## Virtual Machine Reboot
```
tenthirtyam, line 1603: It may be a good idea to consider including these in alphabetical order.

website/docs/r/virtual_machine.html.markdown
AliAllomani: Done

website/docs/r/virtual_machine.html.markdown
AliAllomani: I think it's more organised to keep them in one line as the rest of the document, what do you think ?

website/docs/r/virtual_machine.html.markdown
AliAllomani: I think my statement is more correct here as reducing the memory size even if hot add is enabled will cause reboot, but increase the memory size will cause reboot only if hot add is not enabled

website/docs/r/virtual_machine.html.markdown
tenthirtyam: Either way is good, but readability may be impacted with wrapping.

Will deter to the maintainers.

tenthirtyam (approved):
LGTM! 🚀

Issue #1394, opened 2021-04-21
Add support for shutdown VM state


benfab:
### Description

As an infrastructure operator, I would like to keep some of my vSphere VMs in shutdown state, so they are not consuming compute capacity.  

### Potential Terraform Configuration

This could be defined in the Terraform configuration files by introducing the "desired_status" argument, as shown below:

```hcl
resource "vsphere_virtual_machine" "vm" {
  name             = "terraform-test"

  [...]

  desired_status = "SHUTDOWN"
}
```

### References

This feature is available with the Google provider:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#desired_status

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

aishwarya128:
Hi @koikonom ,

I would like to contribute on this issue, so picking up this if its not already worked upon

tenthirtyam:
PR: https://github.com/hashicorp/terraform-provider-vsphere/pull/1533

tenthirtyam:
After some deliberations with the maintainers, we've decided that at this time this capability will not be added to the provider as it tends to blur the lines between [mutable vs. immutable infrastructure](https://www.hashicorp.com/resources/what-is-mutable-vs-immutable-infrastructure).

If you take a look back to #327, the same enhancement was requested and the maintainer at the time [shared the reasoning](https://github.com/hashicorp/terraform-provider-vsphere/issues/327#issuecomment-352559036) at the time for not including the capability. 

**Ryan Johnson**
Staff II Solutions Architect | VMware, Inc.

PR #1407, opened 2021-05-15
Add check for `vsphere_virtual_machine` power state

sumitmaggo:
### Description

Added a computed parameter (power_state) for checking the power state of VM and to show the difference in case VM is not in running state.
Related to https://github.com/hashicorp/terraform-provider-vsphere/issues/916

When VM is not running and we run terraform plan, it will show the following diff:
This is tested with Terraform 0.13.3 version.

```console
$ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

local_file.meta-data[0]: Refreshing state... [id=221754ffcbea5c01e45fcf0cc40be37aeaa33fda]

vsphere_virtual_machine.example-vm[0]: Refreshing state... [id=564d7840-81a1-72c1-463b-0cc137eed6ea]

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

vsphere_virtual_machine.example-vm[0] will be updated in-place
  ~ resource "vsphere_virtual_machine" "example-vm" {
        boot_delay                              = 0
        boot_retry_delay                        = 10000
        boot_retry_enabled                      = false
        change_version                          = "2020-12-28T06:39:57.251199Z"
        cpu_hot_add_enabled                     = false
        cpu_hot_remove_enabled                  = false
        cpu_limit                               = -1
        cpu_performance_counters_enabled        = false
        cpu_reservation                         = 0
        cpu_share_count                         = 1000
        cpu_share_level                         = "normal"
        datastore_id                            = "5bd0b7d5-02d4a3bb-154a-ac162d77f63c"
        efi_secure_boot_enabled                 = false
        enable_disk_uuid                        = false
        enable_logging                          = false
        ept_rvi_mode                            = "automatic"
        extra_config                            = {}
        firmware                                = "bios"
        force_power_off                         = true
        guest_id                                = "centos8_64Guest"
        guest_ip_addresses                      = []
        hardware_version                        = 14
        host_system_id                          = "ha-host"
        hv_mode                                 = "hvAuto"
        id                                      = "564d7840-81a1-72c1-463b-0cc137eed6ea"
        ide_controller_count                    = 2
        latency_sensitivity                     = "normal"
        memory                                  = 1024
        memory_hot_add_enabled                  = false
        memory_limit                            = -1
        memory_reservation                      = 0
        memory_share_count                      = 10240
[...]

Issue #327, opened 2017-12-18
Add option to control VM power states


Matty9191:
Howdy,

I finally got a chance to play with the vsphere provider. The provider rocks and I want to thank you for all the hard work you did on i!!! We use cobbler to provision systems and I bumped into an issue this morning. Cobbler uses the mac address to connect a given system with a profile during the PXE boot process. I would like to be able to use the mac address assigned to the VM when it is created in my cobbler resource. The mac  would be referenced similar to this:

mac_address = "${vsphere_virtual_machine.kub1_vm.network_interface.0.mac_address}"

Due to the way the graphs are created the cobbler profile is created after the VM is provisioned. That is causing PXE boots to fail. Would it be possible to add VM power control options to control the power state when the VM is provisioned and to power it on and off? That would allow me to create the VM in a powered down state, grab the MAC, create the cobbler profile with the MAC, power the machine up and have the entire provisioning process automated. I spent the past couple of days reading through the documentation and I couldn't find a way to do this. If there is a solution to this I sure would appreciate a pointer to the pertinent docs. If not, maybe you will consider adding power control support in the future?

Thanks again for your efforts!
- Ryan

### Terraform Version

$ /usr/local/bin/terraform -v
Terraform v0.11.1
+ provider.cobbler v1.0.0
+ provider.vsphere v1.1.1

### vSphere Provider Version

$ /usr/local/bin/terraform -v
Terraform v0.11.1
+ provider.cobbler v1.0.0
+ provider.vsphere v1.1.1

### Affected Resource(s)

`vsphere_virtual_machine`

### Terraform Configuration Files

```
provider "vsphere" {
  user           = "${var.vsphere_user}"
  password       = "${var.vsphere_password}"
  vsphere_server = "${var.vsphere_server}"
  allow_unverified_ssl = true
}

provider "cobbler" {
  username = "${var.cobbler_username}"
  password = "${var.cobbler_password}"
  url      = "${var.cobbler_url}"
}

data "vsphere_datacenter" "dc" {
  name = "DC"
}

data "vsphere_datastore" "datastore" {
  name          = "san01"
  datacenter_id = "${data.vsphere_datacenter.dc.id}"
}

data "vsphere_network" "network" {
  name          = "public"
  datacenter_id = "${data.vsphere_datacenter.dc.id}"
}

data "vsphere_resource_pool" "pool" {
  name          = "ProdCluster/Resources"
[...]

PR #1537, opened 2021-12-15
Add VM Shutdown

aishwarya128:
### Description
This feature helps to shutdown/power off a running virtual machine in the environment. User can import the deployed virtual machines using terraform configuration file for the existing infra and he/she can pass this parameter and power off the state.

### Release Note
NONE

### References
Closes: #1394


tenthirtyam:
Closes: #1394

aishwarya128:
> Hi @aishwarya128
> 
> You may want to add "Closes: #1394" to the description to link the issue and pull request.
> 
> Ryan

Thanks @tenthirtyam ,have updated the description.

tenthirtyam:
I think it might be good to consider how to align this one with #1407 and the proposed computed value for `power_state` so that these are grouped (`power_`) and that they each support all three power states of `PoweredOn`, `PoweredOff`, and `Suspended` that would be seen via API/PowerCLI.

For example:
- power_state : the desired state/status
- power_status : the current status

@appilon - what do you think? It's the a similar precedent in other providers?

If so, I can work with the both @aishwarya128 and @sumitmaggo to align these changes.

Ryan

appilon:
I will have to think about these issues more, but I don't think there should be 2 variables, I think if we want to add the ability to control the power state of the VM it would be 1 optional attribute, Terraform is designed to have state match upstream, if there is mismatch the user would be notified of the drift. IE if you change a hypothetical new attribute `power_state` to some value, terraform expects to apply that change and then immediately read that state back form upstream and have it match what was set. I'll try and look at this feature request further.

tenthirtyam:
After reviewing this with a colleague today, we were very much thinking the same and have some thoughts on how it could be addressed.

tenthirtyam:
HI @aishwarya128 👋🏻 

We appreciate your contribution. Unfortunately, after some deliberations with the maintainers, we've decided that at this time this capability will not be added to the provider as it tends to blur the lines between [mutable vs. immutable infrastructure](https://www.hashicorp.com/resources/what-is-mutable-vs-immutable-infrastructure).

If you take a look back to #327, the same enhancement was requested and the maintainer at the time [shared the reasoning](https://github.com/hashicorp/terraform-provider-vsphere/issues/327#issuecomment-352559036) at the time for not including the capability. 

We'd like to encourage you to help contribute to other issues (bugs or enhancements). You can look at the [upcoming version milestones](https://github.com/hashicorp/terraform-provider-vsphere/milestones) to see if there are any in which you'd like to assist.

**Ryan Johnson**
Staff II Solutions Architect | VMware, Inc.

Issue #1438, opened 2021-06-30
ha_datastore_apd_response_delay units issue


BradCalvertHCA:
It appears for the setting ha_datastore_apd_response_delay in vsphere_compute_cluster, the vSphere UI wants this setting in minutes, and this provider wants the setting in minutes, but vSphere actually saves it in seconds behind the scenes (the setting as exposed in PowerCLI is in seconds and is called VmTerminateDelayForAPDSec). So on a default settings brand new cluster, when the provider retrieves the setting during a TF plan, it comes back as 180 seconds instead of 3 minutes and triggers a reconfigure that would result in being set to 3 seconds instead of 180.

### Terraform Version
1.0.0

### vSphere Provider Version
2.0.1

### Affected Resource(s)
vsphere_compute_cluster

### Terraform Configuration Files

```hcl
resource "vsphere_compute_cluster" "clu" {
  for_each = { for obj in local.cluster_inv : "${obj.name}" => obj }

  datacenter_id = each.value.dc_moid
  name          = each.value.name
  host_managed  = true

  ha_enabled = true
  ha_host_monitoring = "enabled"
  ha_vm_component_protection = "enabled"
  ha_datastore_apd_response_delay = 3
}
```
### Expected Behavior
setting ```ha_datastore_apd_response_delay = 3``` (which is the vSphere and TF default should produce no action on a vSphere cluster set to the default.

### Actual Behavior
TF tries to set ```ha_datastore_apd_response_delay``` from 180 to 3 on a fresh cluster

```
# vsphere_compute_cluster.clu["CLUSTERNAME"] will be updated in-place
  ~ resource "vsphere_compute_cluster" "clu" {
      ~ ha_datastore_apd_response_delay                       = 180 -> 3
      ~ ha_vm_component_protection                            = "disabled" -> "enabled"
        id                                                    = "domain-c95409"
        name                                                  = "CLUSTERNAME"
        tags                                                  = []
        # (47 unchanged attributes hidden)
    }
```

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

BradCalvertHCA:
[...]

Issue #1386, opened 2021-04-12
Documentation for vsphere_compute_cluster ha_datastore_apd_response_delay incorrect


jamiegrandfield:
The documentation states that the value for **ha_datastore_apd_response_delay** is in _minutes_.  The value is actually in _seconds_.  Also, the default value of 3 minutes should state 180 seconds.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1542, opened 2021-12-20
Update `ha_datastore_apd_response_delay` defaults and docs

tenthirtyam:
### Description

- Updates `ha_datastore_apd_response_delay` to the API default for `vmTerminateDelayForAPDSec`.
- Updates the `compute_cluster` resource documentation to seconds versus minutes for `ha_datastore_apd_response_delay`.
- Updates the `compute_cluster` resource documentation to use the same structure for describing time-based arguments .

### Release Note

```release-note
Updates `ha_datastore_apd_response_delay` in the `compute_cluster` resource to the API default (180) for `vmTerminateDelayForAPDSec`.
```
### References

- Closes:  #1438
- Closes:  #1386 
- API: [vmTerminateDelayForAPDSec](https://developer.vmware.com/apis/361/doc/vim.cluster.VmComponentProtectionSettings.html#vmTerminateDelayForAPDSec)

### DIFF
--- a/vsphere/resource_vsphere_compute_cluster.go
+++ b/vsphere/resource_vsphere_compute_cluster.go
@@ -308,7 +308,7 @@ func resourceVSphereComputeCluster() *schema.Resource {
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     180,
-				Description: "When ha_vm_component_protection is enabled, controls the delay in minutes to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response.",
+				Description: "When ha_vm_component_protection is enabled, controls the delay in seconds to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response.",
 			},
 			// VM monitoring
 			"ha_vm_monitoring": {
--- a/vsphere/resource_vsphere_compute_cluster.go
+++ b/vsphere/resource_vsphere_compute_cluster.go
@@ -307,7 +307,7 @@ func resourceVSphereComputeCluster() *schema.Resource {
 			"ha_datastore_apd_response_delay": {
 				Type:        schema.TypeInt,
 				Optional:    true,
-				Default:     3,
+				Default:     180,
 				Description: "When ha_vm_component_protection is enabled, controls the delay in minutes to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response.",
 			},
 			// VM monitoring
--- a/vsphere/resource_vsphere_ha_vm_override.go
+++ b/vsphere/resource_vsphere_ha_vm_override.go
@@ -129,7 +129,7 @@ func resourceVSphereHAVMOverride() *schema.Resource {
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     -1,
-				Description: "Controls the delay in minutes to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response. Specify -1 to use the cluster setting.",
+				Description: "Controls the delay in seconds to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response. Specify -1 to use the cluster setting.",
 			},
 			// VM monitoring
 			"ha_vm_monitoring_use_cluster_defaults": {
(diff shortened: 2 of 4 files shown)


vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -307,7 +307,7 @@ func resourceVSphereComputeCluster() *schema.Resource {
 			"ha_datastore_apd_response_delay": {
 				Type:        schema.TypeInt,
 				Optional:    true,
-				Default:     3,
+				Default:     180,
 				Description: "When ha_vm_component_protection is enabled, controls the delay in minutes to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response.",
```
appilon, line 311:
```suggestion
				Description: "When ha_vm_component_protection is enabled, controls the delay in seconds to wait after an APD timeout event to execute the response action defined in ha_datastore_apd_response.",
```

appilon (changes requested):
Normally changing defaults typically could be interpreted as a breaking change if it significantly changes the behavior of the system, but considering we are adjusting a delay, and it appears the old value was erroneously calculated in minutes when the code uses it as seconds, this change is fine, just update the description on in the Go code.

Maybe one day we will generate the docs from the descriptions so they don't need to be double maintained... sigh

appilon:
Actually... might have to do a state migration to make this non-breaking... ugh

tenthirtyam:
Thanks for taking a look into this one, I've committed your changes in the review.

I didn't think about the possible impact to state changes.😬

Ryan

appilon:
In practical terms this should be an easy change, but the fact that all attributes are committed to state it could cause a bad user experience/diff for the users, might be able to migrate their state from 3 to 180 if their config was left empty (implying they were relying on the default). The problem is the old SDK (which this provider uses) doesn't do a good job of providing config vs state...

tenthirtyam:
I suppose the changes in GH-1460 for the TF SDK don't help with that concern?

appilon:
Unfortunately no the new SDK I refer to is this one https://github.com/hashicorp/terraform-plugin-framework. And it won't be feasible to migrate to it for this provider, not to get sidetracked. Lemme see what I can repro in an isolated example.

appilon:
Okay so I ran a few scenarios in an isolated environment.

1) In the case where state is 180 but the default of 3 causes a diff (as described in #1438) the change in this PR fixes the situation completely. It's odd though, I traced the code and I would NOT expect this to be the scenario must users are in, a default of 3 should have been sent to the API, Create/Update call Read to sync state, where I would expect 3 to be read back (not 180).

2) This is the case that I expect to see, where 3 is currently in state, if we switch the default to 180, then the user will be prompted with a state drift.
```
Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply":

  # bug_example.test has changed
  ~ resource "bug_example" "test" {
        id    = "example"
      ~ thing = 3 -> 180
    }

```
If the user applies, then there are no subsequent messages/problems.

Given that, I think it's safe to merge this PR without any state migrations, we will make note in the changelog that a onetime drift message is possible. I did notice there is another piece of schema that I guess mirrors this one but for overrides?

@tenthirtyam Can you perform the doc updates [here](https://github.com/hashicorp/terraform-provider-vsphere/blob/c8905137b4142e0a374e1b6ec34dcf7d3052bde9/vsphere/resource_vsphere_ha_vm_override.go#L132) and [here](https://github.com/hashicorp/terraform-provider-vsphere/blob/c8905137b4142e0a374e1b6ec34dcf7d3052bde9/website/docs/r/ha_vm_override.html.markdown?plain=1#L160) as well?

tenthirtyam:
@appilon I'll make these needed changes and update the PR very soon. Thanks!

Ryan

tenthirtyam:
Committed the changes for the `ha_datastore_apd_response_delay` to denote the time interval from minutes to seconds.

appilon (approved):

tenthirtyam:
Thanks for the support @appilon!

Issue #1454, opened 2021-07-27
Changing `datastore_cluster_id` on  `vsphere_virtual_machine` resource does not move virtual machine


doumhfr:
### Terraform Version
1.0.2
### vSphere Provider Version
2.0.2

### Affected Resource(s)

- `vsphere_virtual_machine`

### Terraform Configuration Files

```hcl
resource "vsphere_virtual_machine" "tst1"{
  name  = "tst1"
  resource_pool_id  = data.vsphere_compute_cluster.cluster.resource_pool_id
  datastore_cluster_id = data.vsphere_datastore_cluster.clusterdatastore.id
```

### Debug Output

```console
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # vsphere_virtual_machine.tst1 will be updated in-place
  ~ resource "vsphere_virtual_machine" "tst1" {
      ~ datastore_id                            = "datastore-2061" -> (known after apply)
        id                                      = "422802a0-ce16-f312-70e1-7944870303b7"
        name                                    = "tst1"
        # (67 unchanged attributes hidden)
        # (4 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

vsphere_virtual_machine.tst1: Modifying... [id=422802a0-ce16-f312-70e1-7944870303b7]
vsphere_virtual_machine.tst1: Still modifying... [id=422802a0-ce16-f312-70e1-7944870303b7, 10s elapsed]
vsphere_virtual_machine.tst1: Modifications complete after 10s [id=422802a0-ce16-f312-70e1-7944870303b7]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

### Expected Behavior

virtual machine is moved to new datastore

### Actual Behavior

nothing happens.
If you re run terraform apply you still have the same modificatn

### Steps to Reproduce

create a vm to a datastore cluster
after creation, change the code to change the datastore cluster

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
[...]

Issue #1268, opened 2020-12-04
Updating the `datastore_id` on the `r/virtual_machine` does not apply to disk sub-resources


jweigand:
### Terraform Version
13.5

### vSphere Provider Version
1.24.2

### Affected Resource(s)
- `vsphere_virtual_machine`

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    vsphere = {
      source  = "hashicorp/vsphere"
      version = "~> 1.24.2"
    }
  }
}

provider "vsphere" {
  vsphere_server       = var.vcenter
  user                 = var.VSPHERE_USER
  password             = var.VSPHERE_PASSWORD
  allow_unverified_ssl = true
}

data "vsphere_datacenter" "dc" {
  name = var.datacenter
}

data "vsphere_compute_cluster" "cluster" {
  name          = var.cluster
  datacenter_id = data.vsphere_datacenter.dc.id
}

data "vsphere_datastore" "manual" {
  name          = var.datastore
  datacenter_id = data.vsphere_datacenter.dc.id
}

data "vsphere_network" "network" {
  name          = var.network
  datacenter_id = data.vsphere_datacenter.dc.id
}

resource "vsphere_virtual_machine" "vm1" {
  name                 = var.vmname
  resource_pool_id     = data.vsphere_compute_cluster.cluster.resource_pool_id 
  datastore_id = data.vsphere_datastore.manual.id
  num_cpus = 1
  memory   = 2048
  guest_id = "other3xLinux64Guest"
  wait_for_guest_net_timeout = 0

  network_interface {
    network_id   = data.vsphere_network.network.id
    adapter_type = "vmxnet3"
  }

  disk {
    label            = "disk0"
    size             = var.c_size
    eagerly_scrub    = false
    thin_provisioned = true
    unit_number      = 0
  }
}
```

### Debug Output
https://gist.github.com/jweigand/174beffb17846c1cee70b5c804f59db7
### Expected Behavior

`datastore_id` should be changed for both the global value on the VM resource, and the disk subresource, causing a storage migration for both the general VM files, the disk VMDK file. 

### Actual Behavior

`datastore_id` only changes for the global value, and only the general VM files are storage migrated. The disk subresource `datastore_id` value does not think it should change, and the disk VMDK files are left on the original datastore. 

```
  # vsphere_virtual_machine.vm1 will be updated in-place
  ~ resource "vsphere_virtual_machine" "vm1" {
[...]

PR #1546, opened 2022-01-03
Fix disk migration when `datastore_cluster_id` changes

pascal-hofmann:
### Description

When changing the `datastore_cluster_id` of a `vsphere_virtual_machine` the disks are not migrated to datastores of the new datastore cluster because no storage migration is triggered. This PR fixes this.

### Acceptance tests
  
   No. There used to be acceptance tests for this functionality that have been disabled with #1090 and removed with #1416.

### Release Note

`resource/virtual_machine`: Fix migration of all disks and configuration files when the `datastore_cluster_id` is changed on the resource. GH-1546

### References

Closes https://github.com/hashicorp/terraform-provider-vsphere/issues/1454

- This fixes the issue described in my comment on https://github.com/hashicorp/terraform-provider-vsphere/issues/1268#issuecomment-980014586, which is different from #1268 itself.
- I think storage migration broke due to https://github.com/hashicorp/terraform-plugin-sdk/issues/165

### DIFF
--- a/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
+++ b/vsphere/internal/virtualdevice/virtual_machine_disk_subresource.go
@@ -816,7 +816,9 @@ func DiskMigrateRelocateOperation(d *schema.ResourceData, c *govmomi.Client, l o
 	ods, nds := d.GetChange(subresourceTypeDisk)
 
 	var relocators []types.VirtualMachineRelocateSpecDiskLocator
-	var relocateOK bool
+
+	// We definitely need to relocate if the datastore cluster changed
+	relocateOK := d.HasChange("datastore_cluster_id")
 
 	// We are only concerned with resources that would normally be updated, as
 	// incoming or outgoing disks obviously won't need migrating. Hence, this is


tenthirtyam:
Closes #1454

appilon:
@tenthirtyam Can this be manualy reproduced and verified to be fixed?

tenthirtyam:
@appilon, yes! I can test this with some NFS-backed datastore clusters to confirm and will report back the results.

Stay tuned!

tenthirtyam:
**Status**: _Currently running some change of `datastore_cluster_id` tests based on the change and some regression testing on `datastore_id` to ensure no implications._

tenthirtyam:
I've reviewed this change in my testbed under vSphere v7.0.3c with two NFS-backed datastore clusters.

I can confirm that it **resolves** the linked issue.

- ✅ When changing `datastore_cluster_id` at the top-most level of a `vsphere_virtual_machine` resource, all disks are relocated to the updated `datastore_cluster_id`.  
- ✅ No regression impact to changing to `datastore_id` at the top-most level of a `vsphere_virtual_machine` resource or at the disk level. 

**Notes**:
-  🧹 While changes to `datastore_cluster_id` at the top-most will move all disks, for `datastore_id` it will only move the VM configuration (e.g. `.vmx`) files. This is not a regression, but it is the current behavior. It is overcome by adding `datastore_id` to each disk.
- 🛠️ Changes between `datastore_cluster_id` to `datastore_id` (_and visa-versa_) will not move the resource's disks between the two. While this is not a regression and is the current behavior, it is not ideal.

tenthirtyam (approved):
See https://github.com/hashicorp/terraform-provider-vsphere/pull/1546#issuecomment-1030398158.

appilon (approved):

PR #1613, opened 2022-03-05
Update packet usage to metal

appilon:
This is the first step in a few in cleaning up the test setup, packet is now metal


tenthirtyam (approved):
🚀 Excellent to see this. Let me know if you need any help reviewing the outcome, etc.

appilon:
This was the easy part... but every step helps

PR #1615, opened 2022-03-08
Revert type change of vsan_disk_group

appilon:
Reverts change from #1432 as it was an accidental breaking change from a configuration perspective, it also caused the provider to panic since the type assertions were not updated correctly. We have re-opened #1205 and will address it for `v3.0.0` of the provider.


tenthirtyam (approved):
Thank you, Alex!

Issue #1282, opened 2020-12-14
Add automated provider acceptance tests


sumitAgrawal007:
Recently there has been many bug reports about OVA deployment failing with error, **runtime error: index out of range [0] with length 0.**
Given that it is a very widely used feature, can we run acceptance tests for every commit and make sure the test cases work.

Recently reported bugs on OVA deployment
I think all of these issues are the same:

https://github.com/hashicorp/terraform-provider-vsphere/issues/1252
https://github.com/hashicorp/terraform-provider-vsphere/issues/1257

There all relate to one commit that broke the OVA deployment flow.

tenthirtyam:
Updated to an enhancement for the project to cover the tech debt that Alex and I have been discussing.

PR #1626, opened 2022-03-18
Rewrite configs to point to new TinyCore based template

appilon:
This is another baby step towards #1282. It would appear after running every single acceptance test that two major problems still exist I plan on addressing in followup PRs

1. There is a panic in the provider regarding an attribute `nvmetcp_maximum_mbit`, hopefully a simple fix
2. The TinyCore VM I created for testing doesn't seem to have reachable internet out of the box, I can't confirm it yet, but it would appear basically all the VM resource tests fail due to network timeout, I am assuming it's due to this. Will dig in.


tenthirtyam (approved):
LGTM! 🚀

PR #1637, opened 2022-03-28
Fix tests to pass even without DHCP

appilon:
Makes tests in environments without DHCP (don't wait for VMs to receive IP address). Also switch to https://github.com/luischanu/TinyVM as the test VM since it comes with OpenVM Tools and a few other goodies.

The status of the tests are:
```
- skipped [29]
- failed  [70]
- passed  [263]
```

The next priority is figuring out what needs to be passed to the VMFS Datastore resource tests as whatever was captured in the setup config doesn't seem to port across deployments.


tenthirtyam (approved):
LGTM, @appilon! 🚀

PR #1643, opened 2022-03-30
Generate VMFS disk IDs for tests

appilon:
The VMFS datastore tests weren't broken but was previously hardcoded, all of them pass with the right values set.


tenthirtyam (approved):
Looks great, @appilon!

PR #1687, opened 2022-06-24
fix: Update `docscheck.sh` to check data source docs

tenthirtyam:
### Description

In the v2.2.0 release, we moved all data source documentation into subcategories. This change updates the `./scripts/docscheck.sh` script to check each resource and data source documentation item for a subcategory to help avoid any misses when accepting new documentation.


appilon (approved):
I am unsure of the automation we have in place for linting/checking at this point... but if you fixed something that's great.

PR #1729, opened 2022-08-10
chore: remove legacy docs `.erb` config

tenthirtyam:
### Description

Removes legacy docs .erb configuration file used for terraform.io documentation index. 

Documentation displayed on the Terraform Registry does not use the ERB and instead builds the navigation dynamically based on the directory and YAML frontmatter of a file.

Signed-off-by: Ryan Johnson <johnsonryan@vmware.com>

### References

Terraform Registry Transition: #888


appilon (approved):
If something breaks we can just revert

PR #1811, opened 2023-01-10
chore: Remove CircleCI

appilon:


tenthirtyam (commented):
Looks like the unit test [is failing](https://github.com/hashicorp/terraform-provider-vsphere/actions/runs/3887146306) - I think it's on the linting change (which itself looks correct.)

appilon:
Very odd, locally I'm using go1.19.4 which is what spotted the gofmt issue, something must have changed between 1.18 and 1.19 because when I run it locally against 1.18 there's no issue, please ignore and merge for now, I'll bump the project to 1.19 anyways as a part of getting things back up to par.

tenthirtyam (approved):
LGTM! 🚀.

PR #1841, opened 2023-02-15
fix: user and password from environment variables

plaes:
Currently, documentation mentions that it's possible to use `VSPHERE_USER` and `VSPHERE_PASSWORD` environment variables to provide the credentials to the server.

Unfortunately this doesn't work, as `user` and `password` are marked as required, so environment variable is never looked up.

### Acceptance tests

```

...
```

### Release Note

```release-note
* Make `user` and `password` optional variables for provider, so these could be filled from environment variables as stated in documentation.
```

### DIFF
--- a/vsphere/provider.go
+++ b/vsphere/provider.go
@@ -21,14 +21,14 @@ func Provider() *schema.Provider {
 		Schema: map[string]*schema.Schema{
 			"user": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("VSPHERE_USER", nil),
 				Description: "The user name for vSphere API operations.",
 			},
 
 			"password": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("VSPHERE_PASSWORD", nil),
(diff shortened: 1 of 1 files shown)


appilon:
It's funny I was having a problem with this too but from my testing it was an issue between v1.2 of Terraform and v1.3. Of course now I can't reproduce the issue, but I am investigating right now.

Unfortunately we cannot make these fields optional so I will be closing this PR as it's the only method of authenticating the client. If you have a config you could post as a repro that would be helpful, please also printout the version of Terraform and the provider with
```
$ terraform version
```

appilon:
I'm not really sure what error I was running into in the past, my test configuration seems to work fine on Terraform v1.3.0 and the latest v1.3.9. I successfully loaded the username and password through environment variables. @plaes Are you certain you exported your env vars properly?

plaes:
> I'm not really sure what error I was running into in the past, my test configuration seems to work fine on Terraform v1.3.0 and the latest v1.3.9. I successfully loaded the username and password through environment variables. @plaes Are you certain you exported your env vars properly?

Apologies for wasting Your time
I can confirm that with a clean example environment variable lookup works! Just need to figure out what is wrong with our internal TF modules...

appilon:
@plaes Never a waste of time, appreciate your interest in the provider.

PR #1874, opened 2023-03-29
feat: add vSAN ESA support

zxinyu08:
### Description

Feat: Add vSAN ESA support

This feature is to support configuring vSAN ESA cluster. We added a new field vsan_esa_enabled in compute cluster resource. Unit tests for vSAN ESA service is also added.

**Testing Details:**
We did end to end test from 3 standalone hosts based on the vSphere 8.0. Test cases are as follows:

1. create cluster with vsan and vsan esa enabled.
2. update cluster to vsan enabled but vsan esa disabled -> failed as vsan esa could not be set separately.
3. update cluster to both vsan and vsan esa disabled.
4. update cluster to only vsan enabled.
5. update cluster to vsan enabled and vsan esa enabled -> failed as vsan esa could not be set separately.

For example, a cluster could be created with vSAN ESA enabled by applying .tf file which contains the following settings.
```
resource "vsphere_compute_cluster" "compute_cluster" {
  name            = "terraform-compute-cluster-test"
  datacenter_id   = "${data.vsphere_datacenter.datacenter.id}"
  vsan_enabled = true
  vsan_esa_enabled = true
}
```

### Acceptance tests
- [y] Have you added an acceptance test for the functionality being added?
- [y] Have you run the acceptance tests on this branch?

```
$make testacc TESTARGS="-run=TestAccResourceVSphereComputeCluster_vsanEsaEnabled"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccResourceVSphereComputeCluster_vsanEsaEnabled -timeout 240m
?       github.com/hashicorp/terraform-provider-vsphere [no test files]
=== RUN   TestAccResourceVSphereComputeCluster_vsanEsaEnabled
PASS: TestAccResourceVSphereComputeCluster_vsanEsaEnabled (320.63s)
PASS
ok      github.com/hashicorp/terraform-provider-vsphere/vsphere 320.654s
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/administrationroles    [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/clustercomputeresource  [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/computeresource [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/contentlibrary  [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/customattribute [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/datacenter      [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/datastore       [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/dvportgroup     [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/envbrowse       [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/folder  [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/hostsystem      [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/network [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/nsx     [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/ovfdeploy       [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/provider        [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/resourcepool    [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/spbm    [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/storagepod      [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/structure       [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/testhelper      [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/utils   [no test files]
?       github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/vappcontainer   [no test files]
testing: warning: no tests to run
PASS
...
```

### Release Note

```release-note
...
```


tenthirtyam (commented):
Initial thought is that there should be a version check for 8.0 or later.

tenthirtyam:
@zxinyu08 and @SlimYang - could you open an enhancement issue to link to this pull request?

tenthirtyam (dismissed):
@zxinyu08 and @SlimYang - could you open an enhancement issue to link to this pull request?

Also, since vSAN ESA is only for 8.0 and later there a version check should be considered. Something like:

```go
if !d.Get("vsan_esa_enabled").(bool) {
    if version.Older(viapi.VSphereVersion{Product: version.Product, Major: 8, Minor: 0}) {
    return fmt.Errorf("vsan_esa_enabled is only supported on vSphere 8.0 and higher")
    }
}
```

Marking as draft.

website/docs/r/compute_cluster.html.markdown
```diff
@@ -462,6 +462,9 @@ details, see the referenced link in the above paragraph.
 ### vSAN Settings
 
 * `vsan_enabled` - (Optional) Enables vSAN on the cluster.
+* `vsan_esa_enabled` - (Optional) Enables vSAN ESA on the cluster.
+  `vsan_unmap_enabled` will be set as `true` automatically when `vsan_esa_enabled` is `true`.
```
tenthirtyam, line 466:
```suggestion
  `vsan_unmap_enabled` - will be set as `true` automatically when `vsan_esa_enabled` is `true`.
```

website/docs/r/compute_cluster.html.markdown
```diff
@@ -462,6 +462,9 @@ details, see the referenced link in the above paragraph.
...
+  Disabling unmap is not allowed in vSAN ESA cluster.
 * `vsan_dedup_enabled` - (Optional) Enables vSAN deduplication on the cluster.
   Cannot be independently set to true. When vSAN deduplication is enabled, vSAN
```
tenthirtyam, line 469:
```suggestion
  Cannot be independently set to `true`. When vSAN deduplication is enabled, vSAN
```

zxinyu08:
> @zxinyu08 and @SlimYang - could you open an enhancement issue to link to this pull request?
> 
> Also, since vSAN ESA is only for 8.0 and later there a version check should be considered. Something like:
> 
> ```go
> if !d.Get("vsan_esa_enabled").(bool) {
>     if version.Older(viapi.VSphereVersion{Product: version.Product, Major: 8, Minor: 0}) {
>     return fmt.Errorf("vsan_esa_enabled is only supported on vSphere 8.0 and higher")
>     }
> }
> ```
> 
> Marking as draft.

Thank you @tenthirtyam @appilon for reviewing and it's good to know there will be an enhancement for version checking. We're considering if we could get this merged first with above version checking added as a temp workaround. When the new design is finished, we'll make change accordingly, if this is not a hard dependency. We'll also create an issue to keep this tracked. Please help review if there's any concern? Thanks.

tenthirtyam:
Marked as **Ready for Review**.

zxinyu08:
Hi folks, I just patch a fix in latest commit. The issue is -
terraform apply fails when `vsan_esa_enabled` is set as true and `vsan_unmap_enabled` is unset. It occurs when we enable vSAN ESA for the first time. The root casue is that vSAN unmap service is by default enabled when vSAN ESA is enabled.

However, the value of `vsan_unmap_enabled` will be computed after apply, which means that the value will be null, if this field is not set manually. And null value of bool type will be seen as false. So the pre-check in `resourceVSphereComputeClusterApplyVsanConfig()` will return error directly when `vsan_esa_enabled` is set as true and `vsan_unmap_enabled` is unset. 

Resolution:
Add `GetOkExists()` to check whether `vsan_unmap_enabled` is set manually in .tf file or is computed after apply. If it's null for the first time apply, we'll set it to true by default.

tenthirtyam:
Hi @zxinyu08 👋 -

Could we ask that you [open an `enhancement` issue](https://github.com/hashicorp/terraform-provider-vsphere/issues/new?assignees=&labels=enhancement&template=enhancement.yml&title=Add+support+for+%5BREPLACE+WITH+SHORT+DESCRIPTION%5D)? We can then link the pull request to the enhancement for tracking purposes.

**Ryan Johnson**
Senior Staff Solutions Architect | Product Engineering @ VMware, Inc.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -531,7 +540,7 @@ func resourceVSphereComputeCluster() *schema.Resource {
 			"vsan_unmap_enabled": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     false,
+				Computed:    true,
```
appilon, line 543: Is this computed? does it get automatically toggled under certain conditions by vsphere? If so, what restrictions would be in place for the user setting a discrete value (setting it false but vsphere wants it to be true?)

zxinyu08: Yes, this feature is related with vSAN ESA switch, as mentioned in doc file, vSAN unmap will be by default enabled along with vSAN ESA. Disabling unmap separately is not allowed in vSAN ESA cluster.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1478,15 +1507,8 @@ func resourceVSphereComputeClusterApplyVsanConfig(d *schema.ResourceData, meta i
...
-	dedupEnabled := d.Get("vsan_dedup_enabled").(bool)
-	compressionEnabled := d.Get("vsan_compression_enabled").(bool)
-	if dedupEnabled && !compressionEnabled {
-		return fmt.Errorf("vsan compression must be enabled if vsan dedup is enabled")
-	}
-
-	conf.DataEfficiencyConfig = &vsantypes.VsanDataEfficiencyConfig{
-		DedupEnabled:       dedupEnabled,
-		CompressionEnabled: &compressionEnabled,
+	if version.AtLeast(viapi.VSphereVersion{Product: version.Product, Major: 8, Minor: 0}) {
```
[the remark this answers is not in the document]
appilon: I won't hold the PR up on this though if things are working since this has been a prioritized issue, let me have a short window to see if there's SOMETHING else that could be done to version gate this.

vsphere/resource_vsphere_compute_cluster.go
appilon: I'm going to have to wrap my head around this subtle change. Historically there exists a problem around Optional + Computed and Booleans, and I'm unsure if this would present a subtle breaking change.

Current behavior:

* User specifies nothing OR false
    * Defaults false (or explicitly false) --> unmap is disabled
* User specifies true
    * unmap is enabled
* User DELETES the attribute
   * Because of the default, semantically this is the same as setting back to false --> unmap is disabled

New behavior:

* User specifies nothing
    * If vSAN ESA is disabled --> unmap is also disabled
    * If vSAN ESA is enabled --> unmap is also enabled
* User specified true
    * If vSAN ESA is disabled --> no problem, unmap is now enabled
    * If vSAN ESA is enabled ---> no problem, redundant, unmap remains enabled
* User specified false
    * If vSAN ESA is disabled --> no problem, unmap is disabled
    * If vSAN ESA is enabled --> PROBLEM, I assume the API request to set this to false will error? Safest bet would be to prevent this in the Create/Update codepath, unfortunately with the legacy provider SDK, cross-attribute value validation has to be done at apply time (can't be done in Validate funcs/plantime).
* User DELETES the attribute
    * Terraform takes this as set to `nil`, however since Computed is true, Terraform just forwards along whatever was previously set in State, therefore no change is triggered (no longer gets set back to false).
    
This is a very nuanced situation, I'd argue it's probably not such a hard deviation of behavior that it would a dealbreaker. A user upset could explicitly try and set it to false, and probably experience an error (either from the vSphere API itself, or if we codify this restriction into the implementation).

We could allow this schema change and deal with any "fallout" if there's an edgecase I failed to capture (Boolean + Optional + Computed has historically created extremely frustrating scenarios so there could be one I'm forgetting).

Or we could not switch to Computed and simply error during apply if the user has it unset/set to false IF vSAN ESA is enabled.

What are your thoughts? I actually lean towards my latter suggestion, I have always hated the semantic ambiguity of unset + computed potentially resulting in `True` on the backend, I much prefer `nil` be a semantic `False` assumption throughout codebases.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1463,6 +1473,25 @@ func resourceVSphereComputeClusterApplyVsanConfig(d *schema.ResourceData, meta i
...
+		if !d.Get("vsan_enabled").(bool) && d.Get("vsan_esa_enabled").(bool) {
+			return fmt.Errorf("vSAN ESA service cannot be enabled on cluster due to vSAN is disabled: %s", d.Get("name").(string))
+		}
+		if !d.HasChange("vsan_enabled") && d.HasChange("vsan_esa_enabled") {
+			return fmt.Errorf("vSAN ESA service must be configured along with vSAN service: %s", d.Get("name").(string))
+		}
+		if d.Get("vsan_esa_enabled").(bool) {
+			// need to be revised if GetOkExists() is deprecated in the future
+			unmapEnabled, unmapOK := d.GetOkExists("vsan_unmap_enabled")
```
appilon, line 1486: I would not rely on GetOkExists at all. Like i wrote up in my longwinded comment above. I think it's safer to not have this attribute be computed, just use `Get` which means `nil/unset` should be resolved to us as `False` and then error if it's False when they are trying to enable vSAN ESA.

vsphere/resource_vsphere_compute_cluster.go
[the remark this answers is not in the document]
appilon: I will allow the old school version gating for vSphere 8 for now to not be a blocker.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1478,15 +1507,8 @@ func resourceVSphereComputeClusterApplyVsanConfig(d *schema.ResourceData, meta i
...
-	compressionEnabled := d.Get("vsan_compression_enabled").(bool)
-	if dedupEnabled && !compressionEnabled {
-		return fmt.Errorf("vsan compression must be enabled if vsan dedup is enabled")
-	}
-
-	conf.DataEfficiencyConfig = &vsantypes.VsanDataEfficiencyConfig{
-		DedupEnabled:       dedupEnabled,
-		CompressionEnabled: &compressionEnabled,
+	if version.AtLeast(viapi.VSphereVersion{Product: version.Product, Major: 8, Minor: 0}) {
+		conf.VsanClusterConfig.(*vsantypes.VsanClusterConfigInfo).VsanEsaEnabled = structure.GetBool(d, "vsan_esa_enabled")
```
appilon, line 1511: If we opt for the non-nil implementation I've described you wouldn't need to use `GetBool` you would simply pass a pointer with an explicitly valued Boolean (see `compressionEnabled`)

appilon (dismissed):
I have some comments around the nuance of how we want to handle the Boolean logic. ~~However the "READ" side of the implemenation is missing. Please see the `ReadFunc` and ensure vSphere API request is copied back into state (this is how Terraform ensures what exists on the backend matches when the user specified).~~ I apologize I missed the d.Set call.

vsphere/resource_vsphere_compute_cluster_test.go
```diff
@@ -304,6 +304,28 @@ func TestAccResourceVSphereComputeCluster_vsanDITEncryption(t *testing.T) {
 	})
 }
 
+func TestAccResourceVSphereComputeCluster_vsanEsaEnabled(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
```
appilon, line 309: We can perform a pre-check to skip the test if vSphere is not >= 8.0. Or we could just have the test error in our CI (which wouldn't be the end of the world). Once CI is running against 8.0 it will pass/be worth looking into when it fails.

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: > I would not rely on GetOkExists at all. Like i wrote up in my longwinded comment above. I think it's safer to not have this attribute be computed, just use `Get` which means `nil/unset` should be resolved to us as `False` and then error if it's False when they are trying to enable vSAN ESA.

that means user must type "vsan_unmap_enabled = true" along with "vsan_esa_enabled = true" in .tf file, otherwise provider throw error. It avoids such verbosity if-else check, let me change to this way and explain this restrictions clearly in documentation.

vsphere/resource_vsphere_compute_cluster_test.go
zxinyu08: > We can perform a pre-check to skip the test if vSphere is not >= 8.0. Or we could just have the test error in our CI (which wouldn't be the end of the world). Once CI is running against 8.0 it will pass/be worth looking into when it fails.

If we could ignore its result, I prefer to keep it here.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1499,6 +1521,26 @@ func resourceVSphereComputeClusterApplyVsanConfig(d *schema.ResourceData, meta i
 		return fmt.Errorf("cannot apply vsan service on cluster '%s': %s", d.Get("name").(string), err)
 	}
 
+	if !d.Get("vsan_esa_enabled").(bool) {
```
zxinyu08, line 1524: Found a bug here, we separate dedup and compression in another reconfigure call, due to these two fields are conflict with vSAN ESA, they cannot be set in same spec along with vsan esa, but if we only check whether vsan esa is false, the reconfigure will also be called when we try to disable vsan esa, it's definitely wrong. The correct check is only vsan is true **AND** vsan esa is false, these two fields could be added to conf.

vasilsatanasov (commented):
I see that you have one comment not resolved, would you mind addressing it ?
https://github.com/hashicorp/terraform-provider-vsphere/pull/1874/files#r1298744443
IMHO it will fail constantly on the  environment against the acceptance tests are executed on github. AFAIK vSphere 7 is used there. The acceptance tests are not 100% stable but I don’t think we need to add more instability. So please add the check that Alex asked. I don’t think I have any other considerations, but I think  Ryan should have the final word here

vsphere/resource_vsphere_compute_cluster.go
vasilsatanasov: Regarding this `vsan_unmap_enabled ` property, since it has to be enabled on ESA cluster and cannot be disabled until ESA is enabled, should we introduce it to the user at all ? Is there a case when the configuration might be `vsan_esa_enabled =false` and `vsan_unmap_enabled=true` ? If so would not it be better to introduce a flag attribute how the `vsan_unmap_enabled` should be handled if `vsan_esa_enabled` is turned off, somethng like  `turn off` | `keep as is`? 

Considering this as a solution to above comments because the feature is vSAN ESA and this is what we want to introduce to the users and I am not sure about the part of the `vsan_unmap_enabled`. If it is important feature like we want to give the users control over the `vsan_unmap_enabled` attribute and It may create a combination with the `vsan_esa_enabled`  which will not work for the user, will make terraform end up with improper state or sth probably a validation should be added in the beginning of create and update methods of the resource to notify the user that the created configuration is not correct.

zxinyu08: The constraints between vsan esa and vsan unmap are:
1. vsan unmap will be enabled by default when users try to enable vsan esa.
2. vsan unmap could not be disabled separately with vsan esa enabled.
3. vsan unmap could be set as true standalone only with vsan enabled, that's the case `vsan_esa_enabled =false` and `vsan_unmap_enabled=true`
In latest commit, we modify vsan unmap from computed true to false back, and make it default value as false. Then only one invalid combination `vsan_esa_enabled =true` and `vsan_unmap_enabled=false`is forbidden and terraform will directly output err msg. It also means users should explicitly configure both vsan unmap and vsan esa as true if they're going to enable vsan esa, this constraint is added in Line 1484. We could have an offline discussion if there're any other concerns, thanks.

vsphere/resource_vsphere_compute_cluster_test.go
vasilsatanasov: Resolved with https://github.com/hashicorp/terraform-provider-vsphere/pull/1874/commits/f6adc75c6454cb760f2ecd27072c54a4fa37aecb

tenthirtyam (approved):
LGTM!

tenthirtyam (approved):
LGTM! 🚀 

I'll add in the `CHANGELOG.md` update prior to approval from @iBrandyJackson.

tenthirtyam:
@iBrandyJackson - the `CHANGELOG.md` has been update and is ready for your approval.

PR #1885, opened 2023-04-17
feat: add vsan stretched cluster support

### LINKED ISSUE #1900: Add support for vSAN stretched cluster enablement
### Community Guidelines

- [X] I have read and agree to the [HashiCorp Community Guidelines ](https://www.hashicorp.com/community-guidelines).
- [X] Vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue initial description to help the maintainers prioritize.
- [X] Do not leave "+1" or other comments that do not add relevant information or questions.
- [X] If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

Add support for enabling/disabling vSAN stretched cluster.

### Use Case(s)

Enable control of vSAN stretched cluster.

### Potential Terraform Provider Configuration

```
data "vsphere_host" "faultdomain1_hosts" {
  count         = length(var.faultdomain1_hosts)
  name          = var.faultdomain1_hosts[count.index]
}

data "vsphere_host" "faultdomain2_hosts" {
  count         = length(var.faultdomain2_hosts)
  name          = var.faultdomain2_hosts[count.index]
}

data "vsphere_host" "witness_node" {
  name          = "10.78.86.87"
}

resource "vsphere_compute_cluster" "cluster" {
  name          = "cluster-test"
  drs_enabled              = false
  drs_automation_level     = "fullyAutomated"
  ha_enabled               = false

  vsan_enabled       = true
  vsan_stretched_cluster_enabled = true
  vsan_stretched_cluster_conf {
    fault_domain1_host_ids = "${data.vsphere_host.faultdomain1_hosts.*.id}"
    fault_domain2_host_ids = "${data.vsphere_host.faultdomain2_hosts.*.id}"
    witness_node = "${data.vsphere_host.witness_node.id}"
  }
```

### References

https://github.com/hashicorp/terraform-provider-vsphere/pull/1885

zxinyu08:
### Description

This feature is to support enabling and disabling vSAN stretched cluster. We added a new field `vsan_stretched_cluster` in compute cluster resource for its configurations.

- `preferred_fault_domain_host_ids` - (Mandatory) The managed object IDs of the hosts to put in the first fault domain.
- `secondary_fault_domain_host_ids` - (Mandatory) The managed object IDs of the hosts to put in the second fault domain.
- `witness_node` - (Mandatory) The managed object IDs of the host selected as witness node when enable stretched cluster.
- `preferred_fault_domain_name` - (Optional) The name of first fault domain. Default is Preferred.
- `secondary_fault_domain_name` - (Optional) The name of second fault domain. Default is Secondary.

Unit test for vSAN Stretched cluster enable/disable is added.

Testing Details:

We did end to end test based on the vSphere 7.0 and 8.0. A vSAN cluster could be set up from 3 standalone hosts and be converted to vSAN stretched cluster later by applying .tf file which contains the following settings, witness node is under datacenter but not in cluster.

```icl
data "vsphere_host" "faultdomain1_hosts" {
  count         = length(var.faultdomain1_hosts)
  name          = var.faultdomain1_hosts[count.index]
}

data "vsphere_host" "faultdomain2_hosts" {
  count         = length(var.faultdomain2_hosts)
  name          = var.faultdomain2_hosts[count.index]
}

data "vsphere_host" "witness_node" {
  name          = "10.78.86.87"
}

resource "vsphere_compute_cluster" "cluster" {
  name          = "cluster-test"
  drs_enabled              = false
  drs_automation_level     = "fullyAutomated"
  ha_enabled               = false

  vsan_enabled       = true
  vsan_stretched_cluster {
    preferred_fault_domain_host_ids = [data.vsphere_host.roothost1.id]
    secondary_fault_domain_host_ids = [data.vsphere_host.roothost2.id]
    witness_node = data.vsphere_host.roothost3.id
   }
}
```

To disable vSAN stretched cluster, we need remove `vsan_stretched_cluster` block. Some Negative E2E cases are passed: lack of mandatory fields, duplicately enable/disable.

### Acceptance Tests

```bash
$make testacc TESTARGS="-run=TestAccResourceVSphereComputeCluster_vsanStretchedCluster"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccResourceVSphereComputeCluster_vsanStretchedCluster -timeout 360m
?       github.com/hashicorp/terraform-provider-vsphere [no test files]
=== RUN   TestAccResourceVSphereComputeCluster_vsanStretchedCluster
--- PASS: TestAccResourceVSphereComputeCluster_vsanStretchedCluster (429.86s)
PASS
ok      github.com/hashicorp/terraform-provider-vsphere/vsphere 430.357s
...
```

### Release Note

```release-note
`r/vsphere_compute_cluster`: Added support for vSAN stretched clusters.
```

### References

Closes #1900


tenthirtyam:
Hi @zxinyu08 👋 -

Could we ask that you [open an `enhancement` issue](https://github.com/hashicorp/terraform-provider-vsphere/issues/new?assignees=&labels=enhancement&template=enhancement.yml&title=Add+support+for+%5BREPLACE+WITH+SHORT+DESCRIPTION%5D)? We can then link the pull request to the enhancement for tracking purposes.

**Ryan Johnson**
Senior Staff Solutions Architect | Product Engineering @ VMware, Inc.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -577,6 +578,51 @@ func resourceVSphereComputeCluster() *schema.Resource {
...
+						"witness_node": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "The managed object IDs of the host selected as witness node when enable stretched cluster.",
+						},
+						"preferred_fault_domain_name": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Description: "The name of prepferred fault domain.",
+							Default:     "Preferred",
```
appilon, line 615: I assume these are sane/expected default values vSphere typically uses/assumes? Or does it really not matter.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -577,6 +578,51 @@ func resourceVSphereComputeCluster() *schema.Resource {
 					},
 				},
 			},
+			"vsan_stretched_cluster_enabled": {
```
appilon, line 581: Any chance we can combine semantics of the 2 attributes? Like if there are zero entries in the typeset is it disabled? If there ARE entries in the TypeSet enable it?

This would probably mean the "conf" attribute can't be Computed so Terraform knows when you remove the config we want to explicitly disable this feature.

Is that possible or is the feature more nuanced?

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -577,6 +578,51 @@ func resourceVSphereComputeCluster() *schema.Resource {
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Whether the vSAN stretched cluster is configured for the cluster.",
+			},
+			"vsan_stretched_cluster_conf": {
+				Type:        schema.TypeSet,
```
appilon, line 588: It seems like lower down you are expecting a single entry in the Set. This is a common practice for creating that nice "nested block" of config for the user. We typically do `schema.TypeList` and `MaxItems: 1` to create that. Which will allow you to not have to convert around the TypeSet in the rest of the implementation.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
 	return conf, nil
 }
 
+func buildVsanStretchedClusterReq(d *schema.ResourceData, cluster types.ManagedObjectReference) (*vsantypes.VSANVcConvertToStretchedCluster, error) {
+	log.Printf("[DEBUG] building vsan stretched cluster request...")
+	conf := d.Get("vsan_stretched_cluster_conf").(*schema.Set).List()[0].(map[string]interface{})
+
+	if conf["preferred_fault_domain_host_ids"] == "" {
```
appilon, line 1512: I'm not sure how this runs without panicking, this is a TypeSet so needs to go through with the conversions. I think a simpler design though is to make it `Required: true` and not `Optional` that should enforce that the Set is non-empty (at least I think so anyways, can easily be verified).

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+		return nil, fmt.Errorf("preferred_fault_domain_host_ids field is required in vsan_stretched_cluster_conf when enabling stretched cluster!")
+	}
+
+	if conf["secondary_fault_domain_host_ids"] == "" {
```
appilon, line 1516: Same as above

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+		return nil, fmt.Errorf("secondary_fault_domain_host_ids field is required in vsan_stretched_cluster_conf when enabling stretched cluster!")
+	}
+
+	if conf["witness_node"] == "" {
```
appilon, line 1520: Same as above, but for sure in this case just make the schema `Required: true`

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+		return nil, fmt.Errorf("witness_node field is required in vsan_stretched_cluster_conf when enabling stretched cluster!")
+	}
+
+	witness_node := []string{}
```
appilon, line 1524: please camelCase variables

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+	witness_node = append(witness_node, conf["witness_node"].(string))
+	witness := structure.SliceStringsToManagedObjectReferences(witness_node, "HostSystem")
```
appilon, line 1526: This can be a oneliner:
```suggestion
	witness := structure.SliceStringsToManagedObjectReferences([]string{conf["witness_node"].(string)}, "HostSystem")
```

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+
+	fault_domain_config := vsantypes.VimClusterVSANStretchedClusterFaultDomainConfig{
```
appilon, line 1528: As mentioned please camelCase

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+		SecondFdHosts: structure.SliceInterfacesToManagedObjectReferences(conf["secondary_fault_domain_host_ids"].(*schema.Set).List(), "HostSystem"),
+	}
+
+	// TODO: make diskmapping configurable.
+	return &vsantypes.VSANVcConvertToStretchedCluster{
+		This:              vsan.VsanVcStretchedClusterSystem,
+		Cluster:           cluster.Reference(),
+		FaultDomainConfig: fault_domain_config,
+		WitnessHost:       witness[0],
+		PreferredFd:       conf["preferred_fault_domain_name"].(string),
```
appilon, line 1541: Is this piece of data required on both parts of the structure? (seems like it's here at the top level and inside `FaultDomainConfig`?

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1455,6 +1505,60 @@ func expandVsanDatastoreConfig(d *schema.ResourceData, meta interface{}) (*vsant
...
+		DiskMapping:       nil,
```
appilon, line 1542: I assume this is a pointer type, so no need to explicitly set to `nil` it defaults as `nil`.

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1692,6 +1823,52 @@ func flattenVsanDisks(d *schema.ResourceData, cluster *object.ClusterComputeReso
...
+	if err != nil {
+		return err
+	}
+
+	if res.Returnval == nil {
+		return d.Set("vsan_stretched_cluster_enabled", false)
+	}
+
+	if res.Returnval[0].UnicastAgentAddr != "" {
+		conf := []interface{}{}
```
appilon, line 1837: you don't need to initilized conf, using `append` will handle that for you:
```suggestion
		var conf []interface{}
```

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1692,6 +1823,52 @@ func flattenVsanDisks(d *schema.ResourceData, cluster *object.ClusterComputeReso
...
+			preferredFaultDomainName := witnessHost.PreferredFdName
+			var secondaryFaultDomainName string
+			preferredFaultDomainHostIds := []string{}
+			secondaryFaultDomainHostIds := []string{}
+			for _, hostConf := range obj.VsanHostConfig {
+				name := hostConf.FaultDomainInfo.Name
+				if name == preferredFaultDomainName {
+					preferredFaultDomainHostIds = append(preferredFaultDomainHostIds, hostConf.HostSystem.Value)
+				} else {
+					if len(secondaryFaultDomainName) == 0 {
```
appilon, line 1849:
```suggestion
					if secondaryFaultDomainName == "" {
```

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1692,6 +1823,52 @@ func flattenVsanDisks(d *schema.ResourceData, cluster *object.ClusterComputeReso
...
+		return d.Set("vsan_stretched_cluster_enabled", false)
+	}
+
+	if res.Returnval[0].UnicastAgentAddr != "" {
+		conf := []interface{}{}
+
+		for _, witnessHost := range res.Returnval {
+			preferredFaultDomainName := witnessHost.PreferredFdName
+			var secondaryFaultDomainName string
+			preferredFaultDomainHostIds := []string{}
```
appilon, line 1842:
```suggestion
			var preferredFaultDomainHostIds []string
```

vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1692,6 +1823,52 @@ func flattenVsanDisks(d *schema.ResourceData, cluster *object.ClusterComputeReso
...
+			secondaryFaultDomainHostIds := []string{}
```
appilon, line 1843:
```suggestion
			var secondaryFaultDomainHostIds []string
```

vsphere/resource_vsphere_compute_cluster_test.go
```diff
@@ -304,6 +304,34 @@ func TestAccResourceVSphereComputeCluster_vsanDITEncryption(t *testing.T) {
...
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccResourceVSphereComputeClusterCheckExists(false),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceVSphereComputeClusterStretchedClusterEnabled(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccResourceVSphereComputeClusterCheckExists(true),
+					resource.TestCheckResourceAttr("vsphere_compute_cluster.compute_cluster", "vsan_stretched_cluster_enabled", "true"),
+				),
```
appilon, line 322: Would be good to check the conf values match what you set in config see: https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/teststep#check-functions

vsphere/resource_vsphere_compute_cluster_test.go
```diff
@@ -982,6 +1010,60 @@ resource "vsphere_compute_cluster" "compute_cluster" {
...
 }
 
+func testAccResourceVSphereComputeClusterStretchedClusterEnabled() string {
+	return fmt.Sprintf(`
+%s
+
+resource "vsphere_compute_cluster" "compute_cluster" {
+  name                        = "testacc-compute-cluster"
+  datacenter_id               = data.vsphere_datacenter.rootdc1.id
+  host_system_ids             = [data.vsphere_host.roothost1.id, data.vsphere_host.roothost2.id]
```
appilon, line 1020: I'm unfamiliar with vSAN in general, but for the record the CI and expected testing environment is setup as follows:

`roothost1` Physical/main ESXi host expected to be in a default cluster we created on CI.
`roothost2` nested ESXi host expected to be in a default cluster we created on CI.
`roothost3` nested ESXi host you can feel free to move into a test specific cluster such as this one.
`roothost4` same as 3.

Probably you don't want to put roothost1 in this cluster for sure, we can DM offline to get a sense for what you need for these tests.

appilon: Also host 2,3,4 are only on a private network, 1 is on both public and private. Doesn't seem to affect the other vSAN tests in the past but thought I should add that context.

appilon (changes requested):
Need some changes but I think this PR shouldn't take too long to get ready for merging.

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: > I assume these are sane/expected default values vSphere typically uses/assumes? Or does it really not matter.

Yes, these two names are default values, but the name of fault domains are mandatory.

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: Yes, we only need a nested block of config. Could you please help provide an example of using the schema.TypeList, so that we could follow the current coding specifications?

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: Yes, it's mandatory here, I checked the types of `FaultDomainConfig` in govmomi:
```
type VimClusterVSANStretchedClusterFaultDomainConfig struct {
	types.DynamicData

	FirstFdName   string                         `xml:"firstFdName"`
	FirstFdHosts  []types.ManagedObjectReference `xml:"firstFdHosts"`
	SecondFdName  string                         `xml:"secondFdName"`
	SecondFdHosts []types.ManagedObjectReference `xml:"secondFdHosts"`
}
```
It only stores info of two fault domains, but during enabling stretched cluster, we must choose one of them as preferred fault domain, that is the second fd could also be set as preferred fd.

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: Yes, it's pointer indeed, will remove this line.

vsphere/resource_vsphere_compute_cluster_test.go
zxinyu08: cool, that's really helpful, we're also looking for some examples of validating attribute set, seems we could use [TestCheckTypeSetElemNestedAttrs(name string, key string, values map[string]string)](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#TestCheckTypeSetElemNestedAttrs)
Will try to add more checks.

vsphere/resource_vsphere_compute_cluster_test.go
zxinyu08: We also have the setup of mixed nested and physical ESXis, but the performance might be unstable, so I'll replace hosts with roothost2/3/4 in case, and let's have a try to see if anything else needed.

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: The reason we split to two fields is that besides basic enable/disable stretched cluster, there're other workflows like reconfigure, users could move hosts from one fd to another, and witness node could also be replaced, these workflows use different APIs. This part would be implemented in future change.
If we combine these 2 attributes, I'm thinking how to distinguish if it's the first time we enable stretched cluster or reconfigure needed actually. That's why we create a switch for that.

vsphere/resource_vsphere_compute_cluster.go
vasilsatanasov: Example in the provider code :
https://github.com/hashicorp/terraform-provider-vsphere/blob/main/vsphere/resource_vsphere_vnic.go#L235

HCL example from the provider documentation

```
resource "vsphere_vnic" "v1" {
  host                    = data.vsphere_host.h1.id
  distributed_switch_port = vsphere_distributed_virtual_switch.d1.id
  distributed_port_group  = vsphere_distributed_port_group.p1.id
  ipv4 {
    dhcp = true
  }
  netstack = "vmotion"
}
```

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: have combined 2 attributes in latest commit.

vsphere/resource_vsphere_compute_cluster_test.go
zxinyu08: To test stretched cluster, we could use roothost3 and roothost4 in primary and secondary fault domains, but we still need one more host outside of any cluster to be witness node, roothost1 and 2 doesn't meet this condition, so maybe this acceptance test will be failed as expected in CI, do we need keep code here and skip it? Or just let it failed and ignore its result.

vsphere/resource_vsphere_compute_cluster_test.go
zxinyu08: Have updated with the check functions

vsphere/resource_vsphere_compute_cluster_test.go
tenthirtyam: I suggest the failure should be expected.

vasilsatanasov: IMHO this vSAN case requires specific testing infrastructure which is not supported by the Acceptance environment. Looks we should consider skipping the test instead of accepting that it fails. What about  introducing some env variables like TF_VSPHERE_VSAN_HOST_1, TF_VSPHERE_VSAN_HOST_2 and TF_VSPHERE_VSAN_WITNESS_HOST which will be the minimal setup for the stretched cluster
and add some info in the skip message or in the function which checks if they are present and set to describe what are the requirements for those hosts ?

vasilsatanasov (approved):
The functionality looks OK for me, just need to add code to skip the e2e test until we have a infra for it on the Acc environment. Several more environment variables should be introduced

tenthirtyam:
> The functionality looks OK for me, just need to add code to skip the e2e test until we have a infra for it on the Acc environment. Several more environment variables should be introduced 

Agreed!

tenthirtyam (approved):
LGTM! 🚀 

Post merge, follow-up on [Vasil's comment](https://github.com/hashicorp/terraform-provider-vsphere/pull/1885#pullrequestreview-1721947476) is required regarding setting it to skip the e2e test until the CI can support it and the vars that need to be introduced.

PR #1955, opened 2023-07-19
fix: set vSAN client with the current version of the service en…

zxinyu08:
### Description
Currently the vSAN client created in provider uses reserved version, but it's a best practice to set it as the current version of the service endpoint. Since whatever version of customer's env, 7.0 or 8.0, the compatibility of each vSAN API should be guaranteed by provider side.

### Acceptance tests

```

...
```

### Release Note

```release-note
...
```
### References
https://github.com/hashicorp/terraform-provider-vsphere/issues/1933

### DIFF
--- a/vsphere/config.go
+++ b/vsphere/config.go
@@ -205,7 +205,9 @@ func (c *Config) Client() (*Client, error) {
 	}
 
 	if isEligibleVSANEndpoint(client.vimClient) {
-		client.vimClient.UseServiceVersion("vsan")
+		if err := client.vimClient.UseServiceVersion("vsan"); err != nil {
+			return nil, err
+		}
 		vc, err := vsan.NewClient(ctx, client.vimClient.Client)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


appilon:
I dunno that I could test the implications of this change, but it seems low risk to merge.

vsphere/config.go
```diff
@@ -205,10 +205,12 @@ func (c *Config) Client() (*Client, error) {
 	}
 
 	if isEligibleVSANEndpoint(client.vimClient) {
+		client.vimClient.UseServiceVersion("vsan")
```
appilon, line 208: This function can error, please check it and return error if not nil

appilon (changes requested):

vsphere/config.go
zxinyu08: Thanks for review, have addressed the comment.

appilon (approved):

PR #1966, opened 2023-08-01
Fix for https://github.com/hashicorp/terraform-provider-vsphere/issue…

nklinkachevAtVMware:
…s/1939

### Description

Supported guest OS are dependent on the VM hardware version. Changed querying the OS Family when deploying a VM from template to read the template hardware version instead of using the server default version.

### Release Note

```release-note
...
```

### DIFF
--- a/vsphere/internal/helper/computeresource/compute_resource_helper.go
+++ b/vsphere/internal/helper/computeresource/compute_resource_helper.go
@@ -169,14 +169,14 @@ func DefaultDevicesFromReference(client *govmomi.Client, ref types.ManagedObject
 
 // OSFamily uses the compute resource's environment browser to get the OS family
 // for a specific guest ID.
-func OSFamily(client *govmomi.Client, ref types.ManagedObjectReference, guest string) (string, error) {
+func OSFamily(client *govmomi.Client, ref types.ManagedObjectReference, guest string, hardwareVersion string) (string, error) {
 	b, err := EnvironmentBrowserFromReference(client, ref)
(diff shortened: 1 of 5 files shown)


appilon (approved):
Seems like a low risk change. Can we upgrade the acceptance tests in anyway to test for this?

appilon:
@nklinkachevAtVMware let's just run it in CI, in the future please add a changlog line I can paste into the changelog that matches the format: https://github.com/hashicorp/terraform-provider-vsphere/blob/main/CHANGELOG.md

appilon:
This does not seem to provide a fix we can accept. The code unfortunately fails at runtime due to a type conversion, I verified this locally and fixed the type issue, however now a previously passing test fails with the following:

```
=== RUN   TestAccResourceVSphereVirtualMachine_cloneCustomizeWithNewResourcePool
    resource_vsphere_virtual_machine_test.go:1538: Step 1/1 error: Error running apply: exit status 1

        Error: cannot find OS family for guest ID "other3xLinuxGuest": ServerFaultCode: A specified parameter was not correct: key

          with vsphere_virtual_machine.vm,
          on terraform_plugin_test.tf line 71, in resource "vsphere_virtual_machine" "vm":
          71: resource "vsphere_virtual_machine" "vm" {

--- FAIL: TestAccResourceVSphereVirtualMachine_cloneCustomizeWithNewResourcePool (30.08s)
```

I'm not familiar with how this "key" works with regards to the api? It was previously unset

PR #1969, opened 2023-08-01
feat: add vsan fault domain support

zxinyu08:
### Description
Feat: add support for vSAN Fault Domain Service.

This feature is to support configuring vSAN Fault Domain. We added a new field `vsan_fault_domains` with`fault_domain`, including `name` and `host_ids` in `compute_cluster` resource. To remove all vSAN fault domains info, we need remove `vsan_fault_domains` whole block directly.
**Testing Details:**
We did end to end test from 2 standalone hosts based on the vSphere 7.0 and 8.0.
Following is an example for .tf file:
```
data "vsphere_host" "faultdomain1_hosts" {
  count         = length(var.faultdomain1_hosts)
  name          = var.faultdomain1_hosts[count.index]
}

data "vsphere_host" "faultdomain2_hosts" {
  count         = length(var.faultdomain2_hosts)
  name          = var.faultdomain2_hosts[count.index]
}

data "vsphere_datacenter" "datacenter" {
  name = "DataCenter"
}

data "vsphere_host" "hosts" {
  count         = length(var.hosts)
  name          = var.hosts[count.index]
}
resource "vsphere_compute_cluster" "cluster" {
  name          = "test"

  vsan_enabled       = true
  vsan_fault_domains {
    fault_domain {
      name = "fd1"
      host_ids = "${data.vsphere_host.faultdomain1_hosts.*.id}"
    }
    fault_domain {
      name = "fd2"
      host_ids = "${data.vsphere_host.faultdomain2_hosts.*.id}"
    }
  }
}
```
### Acceptance tests

```
make testacc TESTARGS="-run=TestAccResourceVSphereComputeCluster_faultDomain"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccResourceVSphereComputeCluster_faultDomain -timeout 360m
...
?   	github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/vmworkflow	[no test files]
=== RUN   TestAccResourceVSphereComputeCluster_faultDomain
--- PASS: TestAccResourceVSphereComputeCluster_faultDomain (253.47s)
PASS
ok  	github.com/hashicorp/terraform-provider-vsphere/vsphere	255.387s
testing: warning: no tests to run
...
PASS
ok  	github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/virtualdevice	1.603s [no tests to run]
```

### Release Note

```release-note
...
```
### References

https://github.com/hashicorp/terraform-provider-vsphere/issues/1967


vsphere/resource_vsphere_compute_cluster.go
```diff
@@ -1415,7 +1448,59 @@ func expandClusterConfigSpecEx(d *schema.ResourceData, version viapi.VSphereVers
 		obj.ProactiveDrsConfig = expandClusterProactiveDrsConfigInfo(d)
 	}
 
-	return obj
+	return obj, nil
+}
+
+func expandVsanHostConfig(d *schema.ResourceData, obj []types.VsanHostConfigInfo) []types.VsanHostConfigInfo {
```
appilon, line 1454: it feels strange to me to be taking in []types.VsanHostConfigInfo and also return the same type. Could you add a code comment (maybe even a few sentences in length) somewhere describing the rationale of this in general. I'm curious to now why we are performing the "joining" of the sets?

appilon: Are we not able to just pass along whatever the user has specified?

vsphere/resource_vsphere_compute_cluster_test.go
```diff
@@ -304,6 +304,27 @@ func TestAccResourceVSphereComputeCluster_vsanDITEncryption(t *testing.T) {
...
+			testAccResourceVSphereComputeClusterPreCheck(t)
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccResourceVSphereComputeClusterCheckExists(false),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceVSphereComputeClusterConfigFaultDomains(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccResourceVSphereComputeClusterCheckExists(true),
+					testAccResourceVSphereComputeClusterCheckFaultDomain("fd1", 1, "fd2", 1),
```
appilon, line 321: See if we can use the TypeSet checks (CTRL+F TypeSet on this doc: https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/teststep#check-functions) to check into the TypeSets. If this advise does not apply here we can stick with the custom func you've written.

appilon (commented):
I have some general questions, but this PR seems overall pretty simple and shouldn't take much to get merged.

vsphere/resource_vsphere_compute_cluster.go
zxinyu08: putting hosts into fault domain is not very straight forward, this should be done by inserting one field of `VsanHostConfig`, which content is:
```
type VsanHostConfigInfo struct {
	DynamicData

	Enabled         *bool                          `xml:"enabled"`
	HostSystem      *ManagedObjectReference        `xml:"hostSystem,omitempty"`
	ClusterInfo     *VsanHostConfigInfoClusterInfo `xml:"clusterInfo,omitempty"`
	StorageInfo     *VsanHostConfigInfoStorageInfo `xml:"storageInfo,omitempty"`
	NetworkInfo     *VsanHostConfigInfoNetworkInfo `xml:"networkInfo,omitempty"`
	FaultDomainInfo *VsanHostFaultDomainInfo       `xml:"faultDomainInfo,omitempty"`
	VsanEsaEnabled  *bool                          `xml:"vsanEsaEnabled"`
}
```
After the cluster created, there're other info tracked in this data structure, we should keep them and just update the `FaultDomainInfo`, that's why we pass and return same type. Because the changes we make is something like insertion. We'll explain this in comment.

vasilsatanasov (approved):

tenthirtyam:
I'll be reviewing this one today.

tenthirtyam (commented):
@zxinyu08 - a couple questions:

- Should this be singular `fault_domain` versus the plural? 

```hcl
vsan_fault_domain {
    name = "rack-1"
    host_ids = [data.vsphere_host.rack1.*.id]
  }
vsan_ fault_domain {
    name = "rack-2"
    host_ids = [data.vsphere_host.rack2.*.id]
  }
```

If plural, I would see this more like:

```hcl
vsan_fault_domains {
    fault_domain = {
        name = "rack-1"
        host_ids = [data.vsphere_host.rack1.*.id]
    }
    fault_domain = {
        name = "rack-2"
        host_ids = [data.vsphere_host.rack2.*.id]
    }
    fault_domain = {
        name = "rack-3"
        host_ids = [data.vsphere_host.rack3.*.id]
    }
}
```

- This configuration is specific to vSAN. Should this be `vsan_fault_domain`?
- Can you confirm that renaming of the fault domains was tested successfully.?
- Can you confirm that add, remove, reassign of hosts to fault domains was tested successfully?
- Perhaps I don't see if, but is there a dependency set that `vsan_enabled` must be `true` to allow `fault_domains` to be used?
- If `vsan_enabled` is set to false and the fault domains still exist, what would occur?
- If the `[data.vsphere_host_foo.*.id]` and  `[data.vsphere_host_foo.*.id]` returns an ID for the same host I assume the API will throw and error. Or should that be caught by the provider first?

tenthirtyam:
Marking as draft whilst the prior open comments are answered. 😃

tenthirtyam:
Hi @zxinyu08 - could you take a look at my questions from the review? Let me know if you have any questions!

Reference: https://github.com/hashicorp/terraform-provider-vsphere/pull/1969#pullrequestreview-1688045938

zxinyu08:
Hi @tenthirtyam, we've addressed the main comment by modifying the field structure to plural way. And for other questions:
1. This configuration is specific to vSAN. Should this be vsan_fault_domain? 
A: Done

2. Can you confirm that renaming of the fault domains was tested successfully?
A: tested in e2e test, passed

3. Can you confirm that add, remove, reassign of hosts to fault domains was tested successfully?
A: tested in e2e test, passed

4. Perhaps I don't see if, but is there a dependency set that vsan_enabled must be true to allow fault_domains to be used?
5. If vsan_enabled is set to false and the fault domains still exist, what would occur?
A: yes, should be, it's a good point, since other configurations are also lack of such check, we plan to fix it in another PR due to it's a common issue for vSAN confs. If users set vsan ks to false, vsan will be disabled directly, other sub-configurations will be disabled automatically, but tfstate file won't be changed if other sub-configurations are not changed in .tf file, it still tracks the former status. We'll fix by adding a vsan check in a new PR.

6. If the [data.vsphere_host_foo.*.id] and [data.vsphere_host_foo.*.id] returns an ID for the same host I assume the API will throw and error. Or should that be caught by the provider first?
A: It's caught by provider now and FD configuration will be blocked.

7. If the number of hosts in [data.vsphere_host_foo.*.id] and [data.vsphere_host_foo.*.id] are not equal sized, should it at least be logged? This would not be a recommended configuration.
8. The more I think about it, I like the mural example above, because then you could consider expecting a minimum of 3 fault domains as best practice and then logging if less than? Just an idea.
A: Since these two are recommended not must have, we added check for them and WARNING msg will be logged.

tenthirtyam:
The functionality looks OK for me, just need to add code to skip the e2e test until we have a infra for it on the Acc environment. Several more environment variables would need be introduced.

Agree @vasilsatanasov?

tenthirtyam (approved):
LGTM! 🚀

Post merge, follow-up is required regarding setting it to skip the e2e test until the CI can support it and the vars that will need to be introduced, if added to e2e.

tenthirtyam:
@zxinyu08 - could you rebase from `origin/main` and resolve the conflicts?

zxinyu08:
> LGTM! 🚀
> 
> Post merge, follow-up is required regarding setting it to skip the e2e test until the CI can support it and the vars that will need to be introduced, if added to e2e.

Hi @tenthirtyam @vasilsatanasov, I think acc test in this PR for fault domain could be supported based on current CI setup. No extra host outside of cluster needed, and fault domain is supported in 7.0, it should be okay as expected.

tenthirtyam:
Thanks! I'll get this merged today and we can prepare for a release next week with the current payload.

tenthirtyam (approved):

PR #1971, opened 2023-08-04
chore: CI add small script to find potential regressions from previous day

appilon:
small JS script to check log against previous dates log (this won't work until after the first day a log is uploaded with the date in the name).


tenthirtyam (approved):
lgtm!

PR #1985, opened 2023-08-09
Restore regression check

appilon:
The previous artifact downloader action didn't support downloads across workflows. Using a popular and HCP approved alternative that does. Added exit code to comparison scripts to flag regressions in CI.


tenthirtyam (approved):
LGTM!

PR #1986, opened 2023-08-09
chore(ci): summarize test output

appilon:
Small script to summarize sums of pass/skip/fail. Will error if pass count is 0. Tweak regression check to flag new failures that passed in previous log (subtle difference).


tenthirtyam (approved):
LGTM!

PR #1989, opened 2023-08-11
chore(ci): ignore workflow status in log download

appilon:
The GH action that downloads artifacts only downloads successful workflow artifacts by default. Setting to empty string allegedly removes that restriction.


tenthirtyam (approved):
LGTM!

PR #2213, opened 2024-06-13
chore: version tracking

tenthirtyam:
### Description

Adds `// Minimum Supported Version: x.y.z` where there is a version restriction.

This is preparation for removing checks and to support only 7.0 and higher per the product lifecycle.


vsphere/virtual_machine_config_structure.go
```diff
@@ -436,6 +441,8 @@ func flattenVirtualMachineFlagInfo(d *schema.ResourceData, obj *types.VirtualMac
 	_ = d.Set("enable_logging", obj.EnableLogging)
 
 	version := viapi.ParseVersionFromClient(client)
+
+	// Minimum Supported Version: 6..0
```
spacegospod, line 445: you probably meant 6.0.0 or 6.X.0

spacegospod (approved):

vsphere/virtual_machine_config_structure.go
tenthirtyam: Updated!

PR #2220, opened 2024-06-17
docs: remove interpolation syntax

tenthirtyam:
### Description

Removes deprecated interpolation syntax where it is no longer required.

**Example**

From:

```hcl
data "vsphere_host" "hosts" {
  count         = "${length(var.hosts)}"
  name          = "${var.hosts[count.index]}"
  datacenter_id = "${data.vsphere_datacenter.dc.id}"
}
```

To:

```hcl
data "vsphere_host" "hosts" {
  count         = length(var.hosts)
  name          = var.hosts[count.index]
}
```

### DIFF
--- a/website/docs/r/compute_cluster_host_group.html.markdown
+++ b/website/docs/r/compute_cluster_host_group.html.markdown
@@ -39,30 +39,30 @@ cluster.
 
 ```hcl
 variable "datacenter" {
-  default = "dc1"
+  default = "dc-01"
 }
 
 variable "hosts" {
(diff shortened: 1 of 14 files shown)


spacegospod (approved):
Nice!
We should probably start cleaning up the syntax in the tests too

tenthirtyam:
> Nice! We should probably start cleaning up the syntax in the tests too

Yes - 100%! That was the next on the list for me - quite extensive. 😄 💯

PR #2230, opened 2024-06-26
bugfix: Remove default values for ept_rvi_mode and hv_mode

spacegospod:
### Description

I'm removing the default values for `ept_rvi_mode` and `hv_mode` from the virtual machine configuration.
These properties are overwritten during the "read" phase and cause in uplace updates on the VM even if the configuration hasn't changed.

We should allow the vCenter to control the defaults instead of setting them ourselves

This is my second take at attempting to fix this problem.
[My first attempt](https://github.com/hashicorp/terraform-provider-vsphere/pull/2172) was unsuccessful and was reverted. It [caused in-place updates to VMs](https://github.com/hashicorp/terraform-provider-vsphere/issues/2193) managed by configurations created with older versions of the provider.

The reason for this was that I only removed the default values for these attributes. This meant that configurations created with previous versions of the provider had actual values and upgrading to v2.8.0 meant that Terraform was attempting to unset those (since we no longer had the defaults).

The proper way to fix this problem is to not only remove the defaults but mark the attributes as `Computed`.
This way Terraform ignores the previously assigned defaults and we can ensure backwards compatibility.

### Acceptance tests

- [X] Have you run the acceptance tests on this branch?

Ran `TestAccResourceVSphereVirtualMachine_basic` and confirmed that even though the scenario is successful the test fails because the plan is not empty. This is due to the default values for `ept_rvi_mode` and `hv_mode`.

This problem is reproducible on most VM tests. Some even have `ignore_changes` blocks to work around this.
The problem is visible on the nightly runs https://github.com/hashicorp/terraform-provider-vsphere/actions/runs/8954207241/job/24593567973#step:10:241

I modified the resoruce by removing the defaults and marking the attributes as computed.
Ran `TestAccResourceVSphereVirtualMachine_basic` again and did not observe the problem.

I copied the exact same configuration from the test, installed v2.7.0, ran `terraform apply` and then `terraform plan`.
I observed that the issue is reproducible. I configured a local override for the provider, ran `terraform plan` on the same configuration but with the binary which contains the fix. I observed that the issue is not reproducible.

I manually set values to the two attributes, different than the defaults, ran `terraform apply` and verified that the operation is successful.

### Release Note

```
NONE
```

### References

https://github.com/hashicorp/terraform-provider-vsphere/issues/1902


tenthirtyam (approved):
Awesome - looks good and will be great to have this resolved!

PR #2244, opened 2024-08-09
feat: allow resizing of virtual disks

### LINKED ISSUE #851: Add support to resize `r/vsphere_virtual_disk`
Right now the vsphere_virtual_disk resource does not permit any changes, but we need to increase disk size on occasion. It is unreasonable to expect us to find a manual workaround to copy data from an old disk to a new larger disk. I would understand requiring a new disk when decreasing the size since that is generally not possible on most systems. During our tests we found the 'eagerZeroedThick' type does not allow resizing by vSphere but 'lazy' should permit increases in size.

> NOTE: All fields in the vsphere_virtual_disk resource are currently immutable and force a new resource if changed.

### Terraform Version

0.12.x

### vSphere Provider Version

1.12.0

### Affected Resource(s)

Please list the resources as a list, for example:
- `vsphere_virtual_disk `

### Expected Behavior

We expect to be able to increase the size of a type='lazy' vsphere_virtual_disk.

### Actual Behavior

Disk wants to be recreated and we will lose all data.

spacegospod:
### Description

Increasing the size of a disk will no longer recreate it but modify it in-place.

Reductions in size are **not supported** by vCenter and will not be allowed.

The rest of the properties for `r/virtual_disk` are still marked with `ForceNew` and will continue to recreate the disk when changed.

### Acceptance tests

- [X] Have you added an acceptance test for the functionality being added?
- [X] Have you run the acceptance tests on this branch?

Added a new test case for this scenario.

### Release Note

`r/virtual_disk`: Allow resizing of virtual disks

### References

Closes #851

### DIFF
--- a/vsphere/resource_vsphere_virtual_disk.go
+++ b/vsphere/resource_vsphere_virtual_disk.go
@@ -338,7 +338,11 @@ func resourceVSphereVirtualDiskUpdate(d *schema.ResourceData, meta interface{}) 
 	log.Printf("[INFO] Updating Virtual Disk")
 	client := meta.(*Client).vimClient
 
-	// TODO - only allow increase
+	oldSize, newSize := d.GetChange("size")
+	if newSize.(int) < oldSize.(int) {
+		return fmt.Errorf("shrinking a virtual disk is not supported")
+	}
+
 	vDisk := virtualDisk{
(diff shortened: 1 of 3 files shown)


tenthirtyam (approved):
Very nice addition for a long-time enhancement need!

PR #2349, opened 2025-02-26
feat: support retry in `d/vsphere_network`

annakhm:
### Description

When nsxt and vsphere configurations are applied in one go, we need a way to wait for nsxt network to be realized on VC, otherwise network data source will fail.

This change adds two parameters to network data source - `retry_timeout` and `retry_interval`. If `retry_timeout` is unset, there would be no change to the data source behavior. Otherwise data source will retry according to the specified parameters. Retry will only occur for `NotFound` error.

### Acceptance tests

```
$ make testacc TESTARGS='-run=TestAccDataSourceVSphereNetwork_'

=== RUN   TestAccDataSourceVSphereNetwork_dvsPortgroup
--- PASS: TestAccDataSourceVSphereNetwork_dvsPortgroup (106.63s)
=== RUN   TestAccDataSourceVSphereNetwork_withTimeout
--- PASS: TestAccDataSourceVSphereNetwork_withTimeout (103.90s)
```

### Release Note

```release-note
`d/vsphere_network`: add `retry_timeout` and `retry_interval` attributes that enable retry if network is not found
```

### DIFF
--- a/vsphere/data_source_vsphere_network.go
+++ b/vsphere/data_source_vsphere_network.go
@@ -5,11 +5,21 @@ package vsphere
 
 import (
 	"fmt"
+	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
 	"github.com/hashicorp/terraform-provider-vsphere/vsphere/internal/helper/network"
+
 	"github.com/vmware/govmomi/object"
+)
+
+const (
+	waitForNetworkPending   = "waitForNetworkPending"
+	waitForNetworkCompleted = "waitForNetworkCompleted"
+	waitForNetworkError     = "waitForNetworkError"
 )
 
 func dataSourceVSphereNetwork() *schema.Resource {
@@ -52,6 +62,18 @@ func dataSourceVSphereNetwork() *schema.Resource {
 					},
 				},
 			},
+			"retry_timeout": {
+				Type:        schema.TypeInt,
+				Description: "Timeout (in seconds) if network is not present yet",
+				Optional:    true,
+				Default:     0,
+			},
+			"retry_interval": {
+				Type:        schema.TypeInt,
+				Description: "Retry interval (in milliseconds) when probing the network",
+				Optional:    true,
+				Default:     500,
+			},
 		},
 	}
 }
@@ -69,8 +91,6 @@ func dataSourceVSphereNetworkRead(d *schema.ResourceData, meta interface{}) erro
 			return fmt.Errorf("cannot locate datacenter: %s", err)
 		}
 	}
-	var net object.NetworkReference
-	var err error
 
 	vimClient := client.Client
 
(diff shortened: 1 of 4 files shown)


tenthirtyam:
>[!NOTE]
> Authored by Anna Khmelnitsky, Broadcom.

website/docs/d/network.html.markdown
```diff
@@ -62,6 +62,8 @@ The following arguments are supported:
   the distributed virtual switch ID.
 * `filter` - (Optional) Apply a filter for the discovered network.
   * `network_type`: This is required if you have multiple port groups with the same name. This will be one of `DistributedVirtualPortgroup` for distributed port groups, `Network` for standard (host-based) port groups, or `OpaqueNetwork` for networks managed externally, such as those managed by NSX.
+* `retry_timeout` - (Optional) If specified, the data source will retry read until timeout has elapsed. Timeout is specified in seconds.
```
tenthirtyam, line 65:
```suggestion
* `retry_timeout` - (Optional) The timeout duration in seconds for the data source to retry read operations.
```

Minor nit.

website/docs/d/network.html.markdown
```diff
@@ -62,6 +62,8 @@ The following arguments are supported:
...
+* `retry_interval` - (Optional) Interval to retry the read in case `retry_timeout` is set. Specified in milliseconds, default is 500.
```
tenthirtyam, line 66:
```suggestion
* `retry_interval` - (Optional) The interval in milliseconds to retry the read operation if `retry_timeout` is set. Default: 500.
```

Minor nit.

tenthirtyam (approved):
LGTM

vsphere/data_source_vsphere_network.go
```diff
@@ -85,19 +105,65 @@ func dataSourceVSphereNetworkRead(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	if dvSwitchUUID != "" {
-		// Handle distributed virtual switch port group
-		net, err = network.FromNameAndDVSUuid(client, name, dc, dvSwitchUUID)
-		if err != nil {
-			return fmt.Errorf("error fetching DVS network: %s", err)
+	readRetryFunc := func() (interface{}, string, error) {
```
spacegospod, line 108: Make this function's signature strongly typed like so `(*object.NetworkReference, string, error)`

This will allow you to replace the `struct{}{}` returns below with `nil` and you won't have to perform a type assertion on the result

spacegospod (changes requested):
Looks good, I have 1 request before I approve

vsphere/data_source_vsphere_network.go
annakhm: Hi @spacegospod, `retryFunc` should be of the type `retry.StateRefreshFunc`, since it is assigned to `Refresh` field of `resource.StateChangeConf`. If I'm missing some trick in golang to make your suggestion work, please let me know!

tenthirtyam: @spacegospod I think that Anna is correct above.

annakhm: @spacegospod is there still a change you would like me to do regarding this one? thanks!

tenthirtyam: I asked Dimitar to look at it as well since Stoyan is on PTO, but I'm good with it as-is.

spacegospod (approved):

vsphere/data_source_vsphere_network.go
spacegospod: @tenthirtyam @annakhm I had missed the part where StateChangeConf needs a refresh function that accepts `interface{}` as its first argument.

Still the code on our side can be improved like so.
(just a suggestion)
```
	readRetryFunc := func() (interface{}, string, error) {
		var net object.NetworkReference
		var err error
		if dvSwitchUUID != "" {
			// Handle distributed virtual switch port group
			net, err = network.FromNameAndDVSUuid(client, name, dc, dvSwitchUUID)
			if err != nil {
				if _, ok := err.(network.NetworkNotFoundError); ok {
					return net, waitForNetworkPending, nil
				}

				return net, waitForNetworkError, err
			}
			return net, waitForNetworkCompleted, nil
		}
		// Handle standard switch port group
		net, err = network.FromName(vimClient, name, dc, filters) // Pass the *vim25.Client
		if err != nil {
			if _, ok := err.(network.NetworkNotFoundError); ok {
				return net, waitForNetworkPending, nil
			}
			return net, waitForNetworkError, err
		}
		return net, waitForNetworkCompleted, nil
	}
```

PR #2361, opened 2025-03-17
fix: `r/vsphere_virtual_machine` skip updating attributes if no diff

root-expert:
### Description

`GetDiffProcessorIfAttributesDefined` function is missing a `HasChange` check which is causing the provider to always update custom attributes regardless if there is a diff or not.

This may cause problems in some cases, such as when you want to restrict the permissions the terraform account has on vsphere. If the permissions are locked down you'll end up getting a permission denied error.

In our case we were not setting the `custom_attributes` parameter at all but the provider still tried to update the remote object which was causing the error:
```
[ERROR] vertex vsphere_virtual_machine.vm[0]" error: error setting attributes for object ID "vm-181897": ServerFaultCode: Permission to perform this operation was denied.
```

### Acceptance tests

### Release Note

```release-note
BUG FIX:
* `r/vsphere_virtual_machine` skip updating custom attributes if there is no diff
```

### References

None


mmadhur-cops:
+1

tenthirtyam (approved):

PR #2366, opened 2025-03-28
Fix handling simultaneous attempts to create parent directories when creating a vsphere_virtual_disk resource

skevir:
### Description
This PR fixes the functionality that attempts to cope with a directory already existing when creating the parent directories for a `vsphere_virtual_disk` resource. For VSAN datastores it has been observed that multiple simultaneous attempts to create the same parent directories can result in vSphere reporting errors.

A previous attempt to fix this (under https://github.com/hashicorp/terraform-provider-vsphere/pull/639) appears to be flawed, in that the error message the functionality is triggered off is actually slightly different to what the code is written to expect. Additionally, there is different error that is sometimes hit in the same scenario, and the original fix didn't handle that case at all.

To capture both of these cases, as well as any other errors that might be seen in the future, I have changed the implementation to ignore the error hit when attempting to create the parent directories and use the existing functionality that performs a post-check that the directory now exists to determine if the creation was successful.

### Testing
I have installed a fixed version of the provider and tested the following. In all case I have performed the test both in the case when the directory doesn't already exist, as well as when it does:
- creating multiple vsphere_virtual_disk resources in the same directory using separate definitions in the Terraform config file.
- creating multiple vsphere_virtual_disk resources using the `count` field on a single resource definition in the Terraform config file.

In the cases that would have previously failed (that is, when the directory didn't already exist) I have confirmed, by looking at the debug logs, that errors were reported when trying to create the directories for some of the vsphere_virtual_disk resources to verify that the bug fix was being invoked. 

### Acceptance tests

- [X] Have you added an acceptance test for the functionality being added? I do not believe any new acceptance tests are required due to the nature of the bug being fixed.
- [X] Have you run the acceptance tests on this branch? No - the code change seems contained enough that the live testing I have performed feels adequate.

```
Not run
```

### Release Note

```release-note
BUG FIX:
`resource/vsphere_virtual_disk`: Fix handling of the creation of parent directories hitting "already exists" errors.
```

### References
Issue that this PR addresses: https://github.com/hashicorp/terraform-provider-vsphere/issues/2309
Issue associated with previous fix attempt: https://github.com/hashicorp/terraform-provider-vsphere/issues/633


tenthirtyam:
Please ammend using [contentional commits](https://www.conventionalcommits.org/en/v1.0.0/) for the commits and PR title.

Marking as draft.

tenthirtyam:
Additionally, if this PR closes an existing issue, please mark is as "Closes #xxx" in the References to link to the PR.

tenthirtyam (dismissed):
Hi, @skevir! 👋🏻  The HashiCorp CLA needs to be signed before this pull request can proceed.

skevir:
> Hi, @skevir! 👋🏻 The HashiCorp CLA needs to be signed before this pull request can proceed.

Thanks - I'm just confirming with my company's legal department that they are happy for me to sign. Once I have that, I will remove the "Draft" status for this PR so that it can proceed.

tenthirtyam:
> I'm just confirming with my company's legal department that they are happy for me to sign. 

Understood! Note that right now the commit is signed-off with your no-reply that does not designate an employer. So be mindful if you need to agree to the CLA as an individual contrinutor or through your employer.

```shell
Merge: 7a288618 9ad0f32c
Author: skevir <67648942+skevir@users.noreply.github.com>
Date:   Mon Mar 31 17:03:54 2025 +0100
```

vsphere/resource_vsphere_virtual_disk.go
```diff
@@ -173,16 +174,27 @@ func resourceVSphereVirtualDiskCreate(d *schema.ResourceData, meta interface{})
 		if directoryPathIndex > 0 {
 			vmdkPath := vDisk.vmdkPath[0:directoryPathIndex]
```
spacegospod, line 175: A predefined sleep timer will probably work in most cases but is not sufficiently robust IMO.
You can declare a mutex (called `m` in the example below) and lock the time-sensitive code

```suggestion
			m.Lock()
			vmdkPath := vDisk.vmdkPath[0:directoryPathIndex]
			log.Printf("[DEBUG] Creating parent directories: %v", ds.Path(vmdkPath))
			err = fm.MakeDirectory(context.TODO(), ds.Path(vmdkPath), dc, true)
			...
			m.Unlock()
```

Terraform invokes the provider over RPC and all disk resources effectively share the same runtime during a `terraform apply`

skevir: If I'm understanding what you are suggesting correctly, the mutex would basically ensure that only one `MakeDirectory` operation is allowed at a time, irrespective of the directory being created. I just wanted to check there isn't any concern about performance in the case multiple different directories need to be created (across multiple disks) where currently those would be created in parallel, but changing to using the mutex would result in them being created in series. I expect in the overall scheme of the time it takes to create a VM, the additional time resulting from creating them in series would be negligible (so using a mutex for its other advantages would be beneficial), but just wanted to get a second opinion. Thanks.

spacegospod (commented):
@skevir the diff looks good!
~~Can you please sign your last commit and mark the PR as ready (non-draft)~~

tenthirtyam (approved):
lgtm

spacegospod (approved):

PR #2438, opened 2025-04-27
refactor: host thumbprint insecure option

tenthirtyam:
### Description

Simplifies the logic for determining the `InsecureSkipVerify` configuration in the `getHostThumbprint` function of `vsphere/resource_vsphere_host.go`.

Removed nested conditionals and redundant checks for the `allow_unverified_ssl` attribute, streamlining the logic for setting `config.InsecureSkipVerify`. The updated code directly assigns the value of `allow_unverified_ssl` if it exists and is of type `bool`.

### DIFF
--- a/vsphere/resource_vsphere_host.go
+++ b/vsphere/resource_vsphere_host.go
@@ -891,18 +891,12 @@ func getHostThumbprint(d *schema.ResourceData) (string, error) {
 	// Otherwise, use the default value of false.
 	if thumbprint, ok := d.Get("thumbprint").(string); ok && thumbprint != "" {
 		return thumbprint, nil
-	} else {
-		if insecure, ok := d.GetOk("allow_unverified_ssl"); ok {
-			if insecureBool, ok := insecure.(bool); ok {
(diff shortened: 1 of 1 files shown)


spacegospod:
You already did this change in https://github.com/hashicorp/terraform-provider-vsphere/pull/2423

vsphere/resource_vsphere_host.go
```diff
@@ -891,18 +891,12 @@ func getHostThumbprint(d *schema.ResourceData) (string, error) {
...
-				}
-			} else {
-				config.InsecureSkipVerify = false
-			}
-		} else {
-			config.InsecureSkipVerify = false
-		}
+	}
+
+	config.InsecureSkipVerify = false
```
spacegospod, line 896: You already did this in https://github.com/hashicorp/terraform-provider-vsphere/pull/2423

tenthirtyam: Doh! It's what happens when you have a bunch you're working on!

PR #2527, opened 2025-05-21
refactor: replace type assertion on errors

tenthirtyam:
### Summary

Replaces type assertion on errors with a call to `errors.As`.

The preferred way for checking for a specific error type is to use the `errors.As` function from the standard library as this function traverses the chain of the wrapped errors while checking for a specific error type.

### Type

- [ ] `fix`: Bug Fix
- [ ] `feat`: Feature or Enhancement
- [ ] `docs`: Documentation
- [x] `refactor`: Refactoring
- [ ] `chore`: Build, Dependencies, Workflows, etc.
- [ ] `other`: Other (Please describe.)

### Breaking Changes?

- [ ] Yes, there are breaking changes.
- [x] No, there are no breaking changes.

### Tests

- [ ] Tests have been added or updated.
- [ ] Tests have been completed.

Output:

### Documentation

- [ ] Documentation has been added or updated.

### DIFF
--- a/vsphere/data_source_vsphere_network.go
+++ b/vsphere/data_source_vsphere_network.go
@@ -5,6 +5,7 @@
 package vsphere
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
@@ -111,7 +112,8 @@ func dataSourceVSphereNetworkRead(d *schema.ResourceData, meta interface{}) erro
 			// Handle distributed virtual switch port group
 			net, err = network.FromNameAndDVSUuid(client, name, dc, dvSwitchUUID)
 			if err != nil {
-				if _, ok := err.(network.NotFoundError); ok {
+				var notFoundError *network.NotFoundError
+				if errors.As(err, &notFoundError) {
 					return struct{}{}, waitForNetworkPending, nil
 				}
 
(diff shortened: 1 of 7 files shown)


vsphere/data_source_vsphere_network.go
```diff
@@ -111,7 +112,8 @@ func dataSourceVSphereNetworkRead(d *schema.ResourceData, meta interface{}) erro
 			// Handle distributed virtual switch port group
 			net, err = network.FromNameAndDVSUuid(client, name, dc, dvSwitchUUID)
 			if err != nil {
-				if _, ok := err.(network.NotFoundError); ok {
+				var notFoundError network.NotFoundError
```
spacegospod, line 115: Can you please double-check if this works with a non-pointer target?

I may be missing something, but when I tried this myself I got `target must be interface or implement error`

spacegospod (approved):

vsphere/data_source_vsphere_network.go
tenthirtyam: Updated all in https://github.com/vmware/terraform-provider-vsphere/commit/5721ff0ccee8b303cab1c3c993f11529baaa89b0 to use a pointer type.

PR #2730, opened 2026-06-08
feat: add `datastore_path` for VM metadata sub-folder placement

### LINKED ISSUE #2729: Allow configuring datastore sub-folder for VM metadata files on r/virtual_machine
### Description

Currently, when `vsphere_virtual_machine` creates a VM, the metadata files (`.vmx`, `.nvram`, log files, etc.) are placed at the **root of the target datastore** in a folder named after the VM. There is no way to influence the directory the VMX is written to via the provider — the `Files.VmPathName` field of the `VirtualMachineConfigSpec` is left to vSphere's default behavior.

This makes it difficult to organize VMs into sub-folders on a shared datastore (for example by environment, tenant, or owning team) directly from Terraform. Users currently have to either:

- pre-create folders out-of-band and accept that Terraform will still drop new VMs at the root, or
- post-process by Storage vMotion'ing the VM into a sub-folder.

### Proposal

Add a new optional, ForceNew string attribute to `resource "vsphere_virtual_machine"`:

```hcl
  name         = "example"
  datastore_id = data.vsphere_datastore.ds.id
  datastore_path = "production/team-a"
  # ...
}
```

When set, the provider should build the VMX path as:

```
[datastore] production/team-a/example/example.vmx
```

instead of the current:

```
[datastore] example/example.vmx
```

### Requirements

- Optional — empty/unset preserves today's behavior (files at datastore root).
- Path is `/`-joined, relative to the datastore root; leading/trailing slashes should be normalized.
- Must work for:
  - bare VM creation on a single datastore (`datastore_id`),
  - bare VM creation under a Storage DRS pod (`datastore_cluster_id`), where the chosen datastore is only known after the SDRS recommendation,
  - VMs created inside a vApp container.
- ForceNew — this only affects initial placement; the provider should not attempt to relocate existing VM files when the value changes.
- Does not affect virtual disk placement (disks continue to follow their own `datastore_id` / SDRS rules).
- Documentation update on `docs/resources/virtual_machine.md`.

### Use case

Organizing dozens to hundreds of VMs that share a small number of large datastores into a meaningful directory layout, without manual datastore browser intervention or post-creation Storage vMotion.

I have a working implementation and will open a PR shortly.

aklakina:
Closes #2729.

### Summary

Adds a new optional, ForceNew `datastore_path` attribute on `resource "vsphere_virtual_machine"` that lets users place the VM metadata files (VMX, NVRAM, logs, …) into a `/`-joined sub-folder of the target datastore instead of dropping them at the datastore root.

```hcl
  name           = "example"
  datastore_id   = data.vsphere_datastore.ds.id
  datastore_path = "production/team-a"
  # ...
}
```

Resulting VMX path:

```
[datastore] production/team-a/example/example.vmx
```

When unset (default), behavior is unchanged — files land at the datastore root.

### Changes

- **`vsphere/resource_vsphere_virtual_machine.go`**
  - New schema attribute `datastore_path` (Optional, ForceNew, String).
  - New `buildVMPathName` helper that normalizes leading/trailing slashes and composes `[datastore] <path>/<name>/<name>.vmx`.
  - Bare standard create path now sets `Files.VmPathName` via the helper.
  - Bare SDRS create path forwards the value to `storagepod.CreateVM`.
- **`vsphere/internal/helper/storagepod/storage_pod_helper.go`**
  - `CreateVM` gains a `dsPath string` parameter; when set, pre-populates `spec.Files.VmPathName = "[] <path>/"` so SDRS-selected datastores honor the sub-folder.
  - vApp-container creation path (`createVAppVMFromSPS`) also receives the path and rebuilds the VMX path with the exported `BuildVMPathName` once the SDRS recommendation reveals the chosen datastore.
- **`docs/resources/virtual_machine.md`** — documents the new attribute, with example and caveats.
- **`CHANGELOG.md`** — `Unreleased` → `FEATURES` entry.

### Behavior / scope

- Optional; unset preserves existing behavior.
- ForceNew — only affects initial placement, no relocation of existing files.
- Does not affect virtual disk placement; disks continue to follow `datastore_id` / SDRS rules.
- Works for single-datastore, datastore-cluster (SDRS), and vApp-container creation paths.

### Testing

- `go build ./...` clean.
- Manual `terraform apply` against a vCenter with and without `datastore_path` set, on both a regular datastore and a Storage DRS cluster, produced the expected VMX paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # <!-- markdownlint-disable first-line-h1 no-inline-html -->
 
+## Unreleased
+
+FEATURES:
+
+- `r/virtual_machine`: Added a new optional `datastore_path` attribute that lets users place virtual machine metadata files (`.vmx`, `.nvram`, logs, etc.) into a `/`-joined sub-folder of the selected datastore instead of the datastore root. Works for both standard datastore and `datastore_cluster_id` (Storage DRS) deployments.
+
 ## v2.16.0
 
 > Release Date: 2026-05-12
--- a/docs/resources/virtual_machine.md
+++ b/docs/resources/virtual_machine.md
@@ -648,6 +648,8 @@ The following options are general virtual machine and provider workflow options:
 
   ~> **NOTE:** When used together with `datastore_cluster_id`, the path is applied to the datastore that Storage DRS selects for the virtual machine.
 
+  ~> **NOTE:** This property is ignored when deploying from OVF/OVA or a content library item.
+
 
 
 * `datacenter_id` - (Optional) The datacenter ID. Required only when deploying an OVF/OVA template.
--- a/docs/resources/virtual_machine.md
+++ b/docs/resources/virtual_machine.md
@@ -640,6 +640,16 @@ The following options are general virtual machine and provider workflow options:
 
 ~> **NOTE:** The `datastore_cluster_id` setting applies to the entire virtual machine resource. You cannot assign individual individual disks to datastore clusters. In addition, you cannot use the [`attach`](#attach) setting to attach external disks on virtual machines that are assigned to datastore clusters.
 
+* `datastore_path` - (Optional) A `/`-joined relative path within the selected datastore (or datastore cluster member) in which to store the virtual machine metadata files (`.vmx`, `.nvram`, log files, etc.). When omitted or empty, the files are placed at the datastore root (the default vSphere behavior, typically a folder named after the virtual machine). Leading and trailing slashes are tolerated and ignored.
+
+  Example: `datastore_path = "tenants/acme/web-tier"` produces a VMX path of `[datastore-name] tenants/acme/web-tier/<vm-name>.vmx`.
+
+  ~> **NOTE:** This setting only affects where the VM configuration/metadata files are created at VM creation time. It does not move files for an existing virtual machine and does not control the placement of virtual disks (use the `datastore_id` setting on a `disk` sub-resource for that). Any intermediate folders must either already exist or be creatable by the user under which the provider is authenticated.
+
+  ~> **NOTE:** When used together with `datastore_cluster_id`, the path is applied to the datastore that Storage DRS selects for the virtual machine.
+
+
+
 * `datacenter_id` - (Optional) The datacenter ID. Required only when deploying an OVF/OVA template.
 
 * `disk` - (Required) A specification for a virtual disk device on the virtual machine. See [disk options](#disk-options) for more information.
--- a/vsphere/internal/helper/storagepod/storage_pod_helper.go
+++ b/vsphere/internal/helper/storagepod/storage_pod_helper.go
@@ -563,4 +563,3 @@ func BuildVMPathName(dsName, dsPath string) string {
 	}
 	return fmt.Sprintf("[%s] %s/", dsName, dsPath)
 }
-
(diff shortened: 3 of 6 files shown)


tenthirtyam (changes requested):
gofmt needs running on the following files:

- ./vsphere/resource_vsphere_virtual_machine.go
- ./vsphere/internal/helper/storagepod/storage_pod_helper.go

spacegospod:
Hey @aklakina the diff looks good overall, I've got a few notes below

* Please add your test configurations as test cases in `resource_vsphere_virtual_machine_test.go`. Let me know if you need help with running the acceptance tests.

* You mentioned in the PR description that the new attribute is flagged as "ForceNew" but it doesn't appear to be.
What do you intend to happen if you modify the value on an existing VM? I don't see any handling in the "update" codepath.

* Do you intend to support this new attribute for clone operations?

aklakina:
Hey @spacegospod

I have added the test cases.
As for the update, sadly the vsphere API does not seem to provide a way to do this without reregistering the virtual machine so I have to stay true to the description and flag it with force new (I just forgot to add the flag)

And as for the clone operations, I found where to place this update for regular clone operations, but I can't seem to tackle where to put this for a content library item clone deployment.
Can you please help me with this one?

And as for running the acceptance tests, i will need help in executing them.

spacegospod:
@aklakina I ran all VM tests on your branch. All green except the new test case, I think you need to add `wait_for_guest_net_timeout = 0`  to one of the new configuration functions.

And here is where the clone from library item happens
https://github.com/vmware/terraform-provider-vsphere/blob/main/vsphere/resource_vsphere_virtual_machine.go#L1568

aklakina:
@spacegospod 
Thank you for running them, i have added the wait disable.

As for the ovf/ova deployment, yes, i found that codepath, but downstream it calls these helpers:
https://github.com/vmware/terraform-provider-vsphere/blob/6e5d67ea56c66dd13531c8b70c180bf35c823d32/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go#L565-L622
And while the helpers expect this object:
https://github.com/vmware/terraform-provider-vsphere/blob/6e5d67ea56c66dd13531c8b70c180bf35c823d32/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go#L51-L67
The helpers proceed to call the internal api of vcenter and there I can't seem to find an option to pass this setting.
For example the vmtx deploy helper has a similarly named option called `VMHomeStorage` but that expects a `DiskStorage` object which has no property to pass a path on the datastore, only the id of the datastore and a policy object. Do you have any idea how to proceed here?

spacegospod:
You're right @aklakina it turns out that there's no option for datastore path when deploying a library item

Reference:
https://developer.broadcom.com/xapis/vsphere-automation-api/latest/data-structures/Vcenter%20VmTemplate%20LibraryItems%20DeploySpec/

https://developer.broadcom.com/xapis/vsphere-automation-api/latest/api/vcenter/vm-template/library-items/templateLibraryItem__action=deploy/post/

Please add a note in the docs that the new property can't be used in this case.
Tag me when you are ready and I will run the tests one last time

aklakina:
@spacegospod

I have updated the documentation to reflect this as requested. I think acc tests should run green now.

spacegospod:
[gotest.log](https://github.com/user-attachments/files/28947897/gotest.log)
re-ran all VM tests

spacegospod (approved):

spacegospod:
@tenthirtyam do you have any comments and/or concerns with the latest diff?

tenthirtyam (approved):
Looks good!

PR #2743, opened 2026-07-10
fix: mark "services" on `r/vnic` as computed

### LINKED ISSUE #2044: vsphere_vnic does not work correctly with vmotion netstack
### Community Guidelines

- [X] I have read and agree to the [HashiCorp Community Guidelines ](https://www.hashicorp.com/community-guidelines).
- [X] Vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue initial description to help the maintainers prioritize.
- [X] Do not leave "+1" or other comments that do not add relevant information or questions.
- [X] If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Terraform

v1.5.7

### Terraform Provider

v2.5.1

### VMware vSphere

v7.0.3

### Description

vsphere_vnic with a netstack of vmotion do not allow updates in place because of the netstack and services conflict

### Affected Resources or Data Sources

resource/vsphere_vnic

### Terraform Configuration

``` hcl
resource "vsphere_vnic" "vmotion_vnic" {
  for_each = vsphere_host.hosts

  host                    = each.value.id
  distributed_switch_port = vsphere_distributed_virtual_switch.vds.id
  distributed_port_group  = vsphere_distributed_port_group.port_groups["vmotion"].id

  ipv4 {
    ip      = cidrhost(lookup(var.vmotion_subnet_prefix, local.workspace.region), tonumber(each.key) + lookup(var.vmotion_subnet_offset, local.workspace.environment))
    netmask = cidrnetmask(lookup(var.vmotion_subnet_prefix, local.workspace.region))
  }

  netstack = "vmotion"

  mtu = 9000

  lifecycle {
    ignore_changes = [
      services,
    ]
  }
}
```

### Debug Output

│ Error: services can only be configured when netstack is set to defaultTcpipStack 

### Panic Output

_No response_

### Expected Behavior

The vmotion vnic should allow for omitting the services and allow updates in places of the vnic.

### Actual Behavior

vnic cannot be updated in place due to the error
`│ Error: services can only be configured when netstack is set to defaultTcpipStack`

### Steps to Reproduce

Creating a vmotion vnic can be done by specifying the netstack. When doing this the first time omitting services is required as both service and netstack can't be provided at the same time.

When applying a second time, it will want to remove the vmotion service vcenter adds automatically. This can then be ignore with the lifecycle as shown above. 

This allows users to create/destroy the vnic but I think it's still broken. When you try to amend the vnic, e.g. change ip address or subnet the error regarding the services appears again, this essentially means you can't change a vmotion netstack vnic after it's been created.

### Environment Details

_No response_

### Screenshots

image

### References

https://github.com/hashicorp/terraform-provider-vsphere/pull/2017#issuecomment-1753832562

image

image

spacegospod:
### Summary

The `services` attribute on `r/vnic` is only allowed and applicable when using the default TCP/IP stack.
All other stacks, including the ones not yet supported by the provider, serve as presets and enable a specific service while simultaneously disallowing further server configration.

The problem with this design is that the read handler for `r/vnic` sets the `services` attribute regardless of the stack. This results in unwanted detection of non existent changes to the resource state.

To solve this I am marking `services` as both optional and computed effectively allowing Terraform to set a value to this attribute without triggering unwanted change detection.

### Type

- [X] `fix`: Bug Fix
- [ ] `feat`: Feature or Enhancement
- [ ] `docs`: Documentation
- [ ] `refactor`: Refactoring
- [ ] `chore`: Build, Dependencies, Workflows, etc.
- [ ] `other`: Other (Please describe.)

### Breaking Changes?

- [ ] Yes, there are breaking changes.
- [X] No, there are no breaking changes.

### Tests

- [X] Tests have been added or updated.
- [X] Tests have been completed.

Output:

[gotest.log](https://github.com/user-attachments/files/29885487/gotest.log)

### Documentation

- [ ] Documentation has been added or updated.

### Issue References

Fixes #2044

### DIFF
--- a/vsphere/resource_vsphere_vnic.go
+++ b/vsphere/resource_vsphere_vnic.go
@@ -25,6 +25,8 @@ const (
 	vnicServiceTypeVsan       = "vsan"
 	vnicServiceTypeVmotion    = "vmotion"
 	vnicServiceTypeManagement = "management"
+
+	vnicNetStackDefault = "defaultTcpipStack"
 )
 
 var vnicServiceTypeAllowedValues = []string{
@@ -312,12 +314,13 @@ func BaseVMKernelSchema() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Optional:    true,
 			Description: "TCP/IP stack setting for this interface. Possible values are 'defaultTcpipStack', 'vmotion', 'provisioning'",
-			Default:     "defaultTcpipStack",
+			Default:     vnicNetStackDefault,
 			ForceNew:    true,
 		},
 		"services": {
 			Type:        schema.TypeSet,
 			Optional:    true,
+			Computed:    true,
 			Description: "Enabled services setting for this interface. Current possible values are 'vmotion', 'management' and 'vsan'",
 			Elem: &schema.Schema{
 				Type:         schema.TypeString,
@@ -395,7 +398,7 @@ func updateVnicService(d *schema.ResourceData, hostID string, nicID string, meta
 }
 
 func precheckEnableServices(d *schema.ResourceData) error {
-	if d.Get("netstack").(string) != "defaultTcpipStack" && len(d.Get("services").(*schema.Set).List()) != 0 {
+	if d.Get("netstack").(string) != vnicNetStackDefault && len(d.Get("services").(*schema.Set).List()) != 0 {
 		return fmt.Errorf("services can only be configured when netstack is set to defaultTcpipStack")
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


spacegospod:
Considered another alternative solution - to skip setting `services` in the read handler when a non-default netstack is used. The downside to this approach is that the value for `services` would not have matched what the user would have been able to retrieve from the API or would have observed in the vCenter UI

tenthirtyam (approved):
LGTM

